Issue #44, opened 2025-03-06
FunctionNode getting tool information from docstrings


Amir-R25:
Currently `FunctionNode` does not pass any information about what the tool expects regarding inputs and outputs unless we define `tool_info` function in a class that is inheriting `FunctionNode` which defeats the purpose of using it.

We should implement `tool_info` inside `FunctionNode` so that it extracts the `Tool` class from the docstrings of the passed in function.

PR #53, opened 2025-03-15
44 improve function node

Amir-R25:
Implemented support for incorporating a user-defined function as a tool using only its docstrings and parameter type-hints. This addresses the feature request in #44.

Modifications in this PR:
* Docstrings are now parsed using `tools/docstring_parser.py`(tests available at `tests/rc_tests/nodes/library/test_function.py`
* Improved readability of `Tool` class by having different handlers for different parameter input types for functions that are meant to be used as tools.
* Organized functionalities related/needed to tools in `src/requestcompletion/llm/tools` directory
* `Tool` class now has a `from_function` method that directly creates a `DynamicFunctionNode` which can be used as below in a `ToolCallLLM` Node:
```python
        @classmethod
        def connected_nodes(cls) -> Set[Type[Node]]:
            return {from_function(test_function)}
```

### Current supported input types:
1. **Primitives** (int, float, str, bool)
2. **Sequences** (list/List, tuple/Tuple)
3. **Pydantic** models and sequences of pydantic models

## Checklist for Author

### DIFF
--- a/src/requestcompletion/nodes/library/function.py
+++ b/src/requestcompletion/nodes/library/function.py
@@ -45,19 +45,20 @@ def from_function(func: Callable) -> Type[Node]:
             try:
                 # Convert kwargs to appropriate types based on function signature
                 converted_kwargs = self._convert_kwargs_to_appropriate_types()
-                
+
                 # Call the function with converted arguments
                 result = func(*self.args, **converted_kwargs)
                 return result
             except Exception as e:
-                warnings.warn(f"Error invoking function: {str(e)}\nProvidedArgs: {self.args}\nProvided kwargs: {self.kwargs}.\n")
-                return f"Error invoking function {func.__name__}"
-        
+                warnings.warn(
+                    f"Error invoking function {func.__name__}\nProvidedArgs: {self.args}\nProvided kwargs:\n{self.kwargs}.\n: {str(e)}")
+                raise e
+
         def _convert_kwargs_to_appropriate_types(self) -> Dict[str, Any]:
             """Convert kwargs to appropriate types based on function signature."""
             converted_kwargs = {}
             sig = inspect.signature(func)
-            
+
             # Process all parameters from the function signature
             for param_name, param in sig.parameters.items():
                 # If the parameter is in kwargs, convert it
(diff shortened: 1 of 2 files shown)


src/requestcompletion/llm/models/_llama_index_wrapper.py
```diff
@@ -45,25 +86,48 @@ def _to_llama_chat(message: Message, tool_call_fn: Callable[[ToolCall], Dict]) -
...
+                schema["additionalProperties"] = False
+        elif isinstance(tool.parameters, dict):
+            # If it's already a dict, use it directly
+            schema = tool.parameters
+            if "additionalProperties" not in schema:
+                schema["additionalProperties"] = False
+            if "type" not in schema:
+                schema["type"] = "object"
+    
+    # If we couldn't get a schema, create a minimal valid one
```
soulFood5632, line 118: We may want to raise a warning here to let them know something went wrong. Like `warnings.warn("tool {name} failed to compile")`

src/requestcompletion/llm/models/openai.py
```diff
@@ -1,10 +1,9 @@
 from ._llama_index_wrapper import LlamaWrapper
 from llama_index.llms.openai import OpenAI
-import openai
 import json
 
 from ..content import ToolCall
-
+from ..content import ToolCall
```
soulFood5632: maybe this is just an artifact, but it says you import the same thing twice

src/requestcompletion/llm/tools/tool.py
```diff
@@ -0,0 +1,153 @@
...
+            SequenceParameterHandler(),
+            DictParameterHandler(),
+            DefaultParameterHandler()
+        ]
+        
+        parameters: Set[Parameter] = set()
+        
+        for param in signature.parameters.values():
+            # Skip 'self' parameter for class methods
+            if in_class and param.name == "self":
```
soulFood5632: We may want to skip `cls` as well if they decide to use `@classmethod`

src/requestcompletion/nodes/library/function.py
```diff
@@ -1,21 +1,185 @@
...
 
+from ..nodes import Node, Tool, Parameter
+import inspect
+from pydantic import BaseModel
 
-from ..nodes import (
-    Node,
-)
 
+TOutput = TypeVar("TOutput")
```
soulFood5632: Small Nitpick, but can we make this generic private so it is `_TOutput`.

soulFood5632: Python's generic syntax is really odd 🙄

src/requestcompletion/nodes/library/function.py
```diff
@@ -1,21 +1,185 @@
...
+        def invoke(self) -> Any:
+            """Invoke the function with converted arguments."""
+            try:
+                # Convert kwargs to appropriate types based on function signature
+                converted_kwargs = self._convert_kwargs_to_appropriate_types()
+                
+                # Call the function with converted arguments
+                result = func(*self.args, **converted_kwargs)
+                return result
+            except Exception as e:
```
soulFood5632, line 52: We should probably still raise this exception after it is thrown.

On another note we should probably think through the whole logging piece. I can put together another ticket for that. Ideally we do not have loose print statements and they are all configured using the python `logging` library

src/requestcompletion/llm/tools/tool.py
Amir-R25: great catch I'll add it in with a testcase

src/requestcompletion/nodes/library/function.py
Amir-R25: I was thinking for now we handle it like similar frameworks where we don't stop executing the entire framework for one faulty function call? What do you think? I agree instead of a print I'll add a warning and then return a `Error invoking tool` instead of the action function's return

Amir-R25: https://github.com/RailtownAI/rc/blob/2dec5c4b30d163e08fbfabd25fd41f0578992242/src/requestcompletion/nodes/library/function.py#L51-L54

What do you think about this?

soulFood5632: Yea I think that makes sense. I think it may make more sense to bubble it up. That way it allows the user to handle the error themselves (kinda related to guan's comment last Tuesday). I can see the argument either way

In our internal modules (like ToolCallLLM) I think this should be the default behaviour tho.

soulFood5632: I can imagine we have some funny typing issues if the expected return is a int or float and it returns a string on an error. Could cause some rather annoying issues.

Issue #49, opened 2025-03-11
Easy-Usage Wrappers


soulFood5632:
Right now creating a new node requires you to write a lot of boiler plate code. See below for a bit of an example:

```python 
class NodeA(TerminalLLM):

  def pretty_name(self):
    return "NodeA"

  def __init__(self, system_message, model, message_history):
     super().__init__(system_message + message_history, model)

```

Things are a bit icky but it would be nice if we could do:

```python 
NodeA = terminal_llm_node(system_message, model)
````

There are some lingering questions how we could implement such a function which generates a node class dynamically. For example could we make it configurable to allow for system_message dependency injection instead of it being passed in at the class level. 

This issues should cover the creation of a wide variety of helper wrappers. Each one should be well documented and include both examples and testing.

PR #55, opened 2025-03-19
43 improve static injection

soulFood5632:
This folks is the big one. We have officially reached a point of simplicity working with the framework that rivals other tooling. Lets just take a peak at what we will be working with. (note some of the examples are dependent on #49 being finished which it is sadly not yet)

```python 
import rc
    
def geolocate():
    # ...

def current_weather(location: str):
    # ...

def long_term_weather(location: str):
    # ...


GeoLocate = rc.from_function(geolocate)

CurrentWeather = rc.from_function(current_weather)
LongTermWeather = rc.from_function(long_term_weather)

WeatherBot = rc.ToolCallLLM([CurrentWeather, LongTermWeather])

async def top_level():
    location = await rc.call(GeoLocate)
    
    prompt = "Tell me about the weather in {location}. Use the tools available to you."
    prompt.format(location)
    
    weather_info = await rc.call(WeatherBot, prompt, model))
    
    return weather_info

TopLevel = rc.from_function(top_level)

with rc.Executor() as executor:
    response = executor.run_sync(TopLevel)
```

In roughly 20 lines of a code you can create an agentic application that is able to complete a wide variety of actions.

In this PR we accomplish the following pieces:
1. The migration to async/await structure for concurrency. This makes a ton more sense for our specific application and its ease of use will allow new users to onboard with the new system with relative ease. [note also that for subservient nodes they can be sequential, they will be automatically mapped into our system]. 
2. Error handling is much different now. The errors thrown by nodes are logged by the system but are bubbled up to the node that called them. This will allow for programmatic handling of errors in subservient nodes. 
3. New simplified mechanism for running a flow. Using the `with` syntax you can spin up an executor which will run your full suite of nodes. 
4. `rc.stream`, `rc.call`, `rc.cancel` are the new tools for interacting with `rc` allowing you to call other nodes etc. Note that `rc.call` will return a coroutine which can be run using the tooling available in async/await. 
5. Continued support for previous library modules like ToolCallLLM. 
6. New test cases for the new system
7. Refactored the code base to allow for simplified import logic

### Future work 
Here is a list of unfinished items. They will be documented in the next couple days for further discussion (they will either be collected as tickets or will be completed while the PR is being reviewed:

1. Documentation of new tooling
2. Testing for `rc.cancel`
3. Wrappers for ToolCallLLM #49 
4. Simplified Stack traces for debugging
5. More user guides for the new changes
6. Change log for the updates
7. A review of design/approach

### DIFF
--- a/tests/rc_tests/nodes/library/test_function.py
+++ b/tests/rc_tests/nodes/library/test_function.py
@@ -206,7 +206,6 @@ class TestPrimitiveInputTypes:
             assert isinstance(children.output, Failure)
 
 
-
 class TestSequenceInputTypes:
     @pytest.mark.parametrize("model_provider", MODEL_PROVIDERS)
     def test_single_list_input(self, model_provider):


src/requestcompletion/info.py
```diff
@@ -28,50 +20,32 @@ def __init__(
...
     ):
         self.request_heap = request_heap
         self.node_heap = node_heap
         self.stamper = stamper
-        self.context = context
-        self.subscriber = subscriber
-        self.exception_history = (
-            exception_history if exception_history is not None else []
-        )
+        self.exception_history = exception_history if exception_history is not None else []
```
Amir-R25: nitpick but this also works:
```python
self.exception_history = exception_history or []
```

src/requestcompletion/nodes/library/function.py
```diff
@@ -192,16 +205,18 @@ class FunctionNode(Node[_TOutput]):
...
+        self.args = args
         self.kwargs = kwargs
 
-    def invoke(self) -> _TOutput:
+    async def invoke(self) -> _TOutput:
         try:
-            result = self.func(**self.kwargs)
-            if result and isinstance(result, str):
-                self.data_streamer(result)
+            if asyncio.iscoroutinefunction(self.func):
```
Amir-R25, line 217: correct me if I'm wrong but are we trying to support tool functions provided by `async def` implementation? And regardless run them concurrently? Wouldn't it be easier if to just ask users to avoid using `async def` method like other frameworks if the end result is the same? Just thinking off the top of my head, feels like the reverse logic might be more valuable that a user can mark a tool as "not async" and we support that?

Regardless, `asyncio.iscoroutinefunction` seems to [be deprecated and to be removed in `python 3.16`](https://docs.python.org/3/whatsnew/3.12.html#:~:text=Unicode%20characters%20instead.-,asyncio%3A,Contributed%20by%20Jiahao%20Li%20and%20Kumar%20Aditya%20in%20gh%2D122875.) so best avoid using it from the get go

soulFood5632: Thanks for that catch. I will use `iscoorotinue` instead

Can you walk me through your argument? My idea with this is it would allow the user to provide an async defintion or a non-async defintion and this logic automatically map the non-async stuff into async without the user being aware. I think that matches what you were saying but lmk

Amir-R25: I guess the main logic I suggest we handle is what if there is a function that the user definitely does not want handled async? 

It will be better from a user perspective to say all functions will be run async unless stated otherwise. Probably fine to support that async -> async as that doesn't really change anything.

soulFood5632: yea I think that makes sense. Let's put up a separate Issue that addresses that. We can have this discussion around what specific use cases would require sequential code. Ultimately async await may allow the users accomplish much of the requirements of sequential code anyway but I am sure situations will come up.

src/requestcompletion/interaction/call.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # we create a new context to prevent bleeding memory from the previous context.
+
+    created_node = node(*args, **kwargs)
+
+    # the reference to current node running must be collected and passed into the state object
+    p_n_id = parent_id.get()
+    # but we also must update the variable so if the child node makes its own calls it is operating on the correct ID.
+    prev_token = parent_id.set(created_node.uuid)
+    try:
```
Amir-R25: correct me if I'm wrong but are we trying to call a node here and if it doesn't finish we cancel it? Will this cancellation cancel the entire framework? Or will the rest continue? Also it might be worth it to also warn the user of what node is cancelled if it the cancellation were successful

Amir-R25: come to think of it, it might also be worth add a boolean parameter that is defaulted to `False` where the user can end the framework on failure of a node or not. Let me know if this sounds confusing

soulFood5632: I haven't yet tested this idea of cancellation. For the sake of this PR I think I am going to disable it. 

The idea is cancelling a node it is kinda an optional tool that the user can use to make sure it is not doing something pointless. (lIke If it makes a pair API calls and just wants to use the first one that comes back). The idea is that it just cancels that individual request and the rest of graph continues as normal. 

And yes I need to make sure that the logic of the cancelation is very documented and defined.

soulFood5632: As per your second comment, I think I deleted that feature and I'm not sure why I did. I'll add it as a flag in the exector config object.

src/requestcompletion/nodes/library/terminal_llm.py
```diff
@@ -17,26 +17,17 @@ def __init__(self, message_history: MessageHistory, model: ModelBase):
...
         """Makes a call containing the inputted message and system prompt to the model and returns the response
 
         Returns:
             (TerminalLLM.Output): The response message from the model
         """
-        try:
-            returned_mess = self.model.chat(self.message_hist)
-        except Exception as e:
-            raise ResetException(node=self, detail=f"{e}")
+
```
Amir-R25, line 69: removed the try/catch because it will be caught in `rc.call`?

src/requestcompletion/tools/__init__.py
Amir-R25, line 1: We will now have two modules called `tools` one is in `requestcompletion/tools/` and the other `requestcompletion/llm/tools`. This might result in some confusions while debugging in the future? What are your thoughts?

src/requestcompletion/state/node.py
```diff
@@ -56,13 +56,13 @@ def __init__(self):
...
+        self.registration_details = None
 
     def __getitem__(self, item):
-        # note the node in item is mutable so we have to preform a deep copy for the sake of a safety.
+
         node = self._heap[item]
-        node.node
         try:
-            return deepcopy(node)
+            return node
```
Amir-R25: is the `NodeCopyException` being raised below this line still relevant if we are removing the `deepcopy(node)`?

playground.py
Amir-R25, line 1: is this file needed? Seems like some testing file that was used?

src/requestcompletion/nodes/library/function.py
Amir-R25, line 1: Another general comment, seems like we are now using `from_function` as a general creating a Node from a workflow like the example provided in the PR. 
```python
async def top_level():
    location = await rc.call(GeoLocate)
    
    prompt = "Tell me about the weather in {location}. Use the tools available to you."
    prompt.format(location)
    
    weather_info = await rc.call(WeatherBot, prompt, model))
    
    return weather_info

TopLevel = rc.from_function(top_level)
```

I am wondering if this might end up being confusing for the users down the road and if we need to rename what giving an LLM a ToolNode that is made from a function is called. As this way a lot of the error handling surrounding around the `DynamicFunctionNode` creation will be kind of strange since the RC framework defining functions may not have complete docstrings/type-hints and they don't need to either but we want to force the users to provide complete docstrings/type-hints for functions that are meant to be used as tools for an LLM.

This sort of goes back to our discussion of what you had in mind for FunctionNode and what I ended up implementing so I think it might be worth it to make that distinction clear here before proceeding.

src/requestcompletion/tools/__init__.py
soulFood5632: Yea, I'll just rename it to utils. Was a name without meaning anyway

src/requestcompletion/nodes/library/terminal_llm.py
soulFood5632: Yea, I am gonna add a issue to create some sort of generic logic here exception handling in both the LLM module and our library nodes.

examples/terminal-llm-example.ipynb
Amir-R25, line 1: seems like we have both `terminal-llm-example.ipynb` and `terminal-llm.ipynb`. I think it's best we combine these too.

examples/tool-call-llm.ipynb
Amir-R25, line 1: I guess this one is not updated with the given LLM Node creation logic?

src/requestcompletion/nodes/library/structured_llm.py
```diff
@@ -9,6 +10,52 @@
 _TOutput = TypeVar("_TOutput", bound=BaseModel)
 
 
+def structured_llm(
```
Amir-R25, line 13: I think it's better to use `Optional[ModelBase] = None`  ; similar for `str` and `SystemMessage`

src/requestcompletion/nodes/library/terminal_llm.py
```diff
@@ -1,7 +1,50 @@
+import warnings
+
 from ...llm import MessageHistory, ModelBase, SystemMessage
 from ..nodes import Node, ResetException
 
-from abc import ABC, abstractmethod
+from abc import ABC
+
+
+def terminal_llm(
```
Amir-R25, line 9: again using `Optional[ParamName]=None` is more "pythonic"

src/requestcompletion/nodes/library/function.py
soulFood5632: Yea I get that. I think we should probably have a bit of a complete thought out plan for the different use cases. 

The way I see it we should have 2 major pieces:
1. Tools to be used in LLMs
2. General Node creation (like in the above example)

Let's chat about it in the meeting. I lean towards allowing all the nodes to created through one interface and only surfacing errors like (no doctoring provided) if you attempt to use them in the context of a LLM. This would keep the interface clean while still signalling errors as they come up.

soulFood5632: In general tho, we should do a full analysis of the logic you had in #53 to make sure all that logic works within the paradigm we decide.

examples/terminal-llm-example.ipynb
soulFood5632: Yeah I can just get rid of it.

src/requestcompletion/nodes/library/structured_llm.py
soulFood5632: I think I kept it as required parameter becuase you can make a structured LLM object without a pydantic type

src/requestcompletion/nodes/library/terminal_llm.py
soulFood5632: Are you saying we should use Optional[str] instead of str | None?

PR #70, opened 2025-03-31
Bug Fix -> message_history mutation issue in library nodes

Aryan-Railtown:
### Issue
Previously, when a message history was passed to any rc.call (eg.- `await rc.call(Node, message_history=message_history)`), the function would mutate the original message history by adding system messages directly to it. This behavior was undesirable as functions should not modify input parameters.
### Fix
The issue has been fixed by deep copying the message history before adding system messages. This ensures the original message history passed by the caller remains unchanged.
### Tests
Comprehensive tests have been added that verify message histories are not mutated after calling various node types:

- Terminal LLM nodes
- Structured LLM nodes
- Tool calling LLM nodes

Tests cover both class-based and wrapper-based node construction methods to ensure consistent behavior across the library.
## Checklist for Author

---
See [Issue 61](https://github.com/RailtownAI/rc/issues/61)

### DIFF
--- a/tests/rc_tests/interaction/conftest.py
+++ b/tests/rc_tests/interaction/conftest.py
@@ -241,7 +241,7 @@ def tool_calling_nodes(request, model, tool_call_llm_system_messages, curreny_co
         
         return currency_converter_node, travel_planner_node
     elif fixture_name == "class_based":
-        def make_tool_call_llm_class_version(pretty_name: str, system_message: rc.llm.SystemMessage, connected_nodes: List[rc.nodes.Node]):
+        def make_tool_call_llm_class_version(pretty_name: str, system_message: rc.llm.SystemMessage, connected_nodes: List[rc.Node]):
             class ToolCallLLMNode(rc.library.ToolCallLLM):
                 def __init__(
                     self,
--- a/tests/rc_tests/interaction/test_call.py
+++ b/tests/rc_tests/interaction/test_call.py
@@ -132,5 +132,5 @@ async def test_message_history_not_mutated_tool_call_llm(model, tool_calling_nod
     with rc.Runner() as runner:
         message_history = rc.llm.MessageHistory([rc.llm.UserMessage("I want to plan a trip to from Delhi to New York for a week. Please provide me with a budget summary for the trip.")])
         original_message_history = deepcopy(message_history)
-        _ = await runner.run(TrravelSummarizerNode, message_history=message_history)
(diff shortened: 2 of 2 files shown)


tests/rc_tests/interaction/conftest.py
```diff
@@ -0,0 +1,271 @@
...
+    ConvertCurrency = rc.library.from_function(convert_currency)
+    AvailableLocations = rc.library.from_function(available_locations)
+    CurrencyUsed = rc.library.from_function(currency_used)
+    AverageLocationCost = rc.library.from_function(average_location_cost)
+
+    if fixture_name == "easy_wrapper":
+        currrency_converter_node = rc.library.tool_call_llm(connected_nodes=[AvailableCurrencies, ConvertCurrency], pretty_name="Currency Converter Node", system_message=system_currency_converter, model=model)
+        travel_planner_node = rc.library.tool_call_llm(connected_nodes=[AvailableLocations, CurrencyUsed, AverageLocationCost], pretty_name="Travel Planner Node", system_message=system_travel_planner, model=model)
+        
+        return currrency_converter_node, travel_planner_node
```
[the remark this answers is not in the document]
soulFood5632: Maybe we should create a RC agent that automatically does PR reviews on just like the copilot suggestions. might be a relevant things demo.

src/requestcompletion/nodes/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .nodes import Node, NodeException, ResetException, FatalException, CompletionException
+
+__all__ = [
```
soulFood5632: I don't think we want to have this behaviour for imports. We expose the `Node` type at the root of the library in `src/requestcompletion/__init__.py` so the only time we are going to access these types is intenerannly and I would rather not have maintain these types of imports in the `__init__.py`

tests/rc_tests/fixtures/nodes.py
```diff
@@ -7,7 +7,7 @@
 from enum import Enum
 from typing import List, Callable
 import requestcompletion as rc
-
+from requestcompletion.nodes import FatalException
```
soulFood5632: See the above comment, but this import path will need to change.

src/requestcompletion/nodes/__init__.py
Aryan-Railtown: Okay, I can remove these.

PR #81, opened 2025-04-08
Terminal LLM easy_usage wrappers tool support

Aryan-Railtown:
# Add Terminal LLM Tool Functionality to easy_wrapper

This PR adds support for using Terminal LLMs as tools in the tool_call_llm node. Terminal LLMs can now be configured to act as tools by providing `tool_details` and `tool_params` during initialization.

## Changes

- Extended `terminal_llm` function to support tool functionality
- Added validation for tool configuration parameters
- Implemented `tool_info()` and `prepare_tool()` methods when configured as a tool
- Added comprehensive test cases for the new functionality
- Created example to demonstrate usage

## Usage

Terminal LLMs can now be configured as tools by specifying required parameters:

```python
encoder = terminal_llm(
    pretty_name="Encoder",  # Required when used as a tool
    system_message=encoder_system_message,
    model=model,
    tool_details="A tool used to encode text into bytes.",   # Required when used as a tool
    tool_params={Parameter("text_input", "string", "The string to encode.")}  # Required when used as a tool
)

# Can be used with tool_call_llm
tool_llm = tool_call_llm(connected_nodes={encoder}, ...)
```

The PR includes validation to prevent misconfiguration and improved error messages for better developer experience.

# Validation and Error Message Improvements

- Added validation to ensure `tool_details` and `tool_params` are provided together
- Added early validation requiring `pretty_name` when configuring as a tool
- Implemented validation to prevent empty parameter sets for tools
- Added check for duplicate parameter names in tool configurations
- Added clarifying warning when `terminal_llm` is being used as a tool

These validations and error messages help developers identify configuration issues early and understand how to properly use the tool functionality.

--- 
## Checklist for Author

---

### DIFF
--- a/tests/rc_tests/interaction/test_call.py
+++ b/tests/rc_tests/interaction/test_call.py
@@ -3,10 +3,10 @@ from copy import deepcopy
 import requestcompletion as rc
 from conftest import structured_nodes, tool_calling_nodes, terminal_nodes, model
 
-NODE_INIT_METTHODS = ["class_based", "easy_wrapper"]
+NODE_INIT_METHODS = ["class_based", "easy_wrapper"]
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("terminal_nodes", NODE_INIT_METTHODS, indirect=True)
+@pytest.mark.parametrize("terminal_nodes", NODE_INIT_METHODS, indirect=True)
 async def test_message_history_not_mutated_terminal_llm(model, terminal_nodes):
     """
     Verify that message history is not modified after rc.call when passed to nodes constructed using different methods.
@@ -54,7 +54,7 @@ async def test_message_history_not_mutated_terminal_llm(model, terminal_nodes):
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("structured_nodes", NODE_INIT_METTHODS, indirect=True)
+@pytest.mark.parametrize("structured_nodes", NODE_INIT_METHODS, indirect=True)
 async def test_message_history_not_mutated_structured_llm(model, structured_nodes):
     """
     Verify that message history is not modified after rc.call when passed to nodes constructed using different methods.
@@ -99,7 +99,7 @@ async def test_message_history_not_mutated_structured_llm(model, structured_node
         assert all(orig.content == new.content for orig, new in zip(original_message_history, message_history)), "Message history modified after runner run"
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("tool_calling_nodes", NODE_INIT_METTHODS, indirect=True)
+@pytest.mark.parametrize("tool_calling_nodes", NODE_INIT_METHODS, indirect=True)
 async def test_message_history_not_mutated_tool_call_llm(model, tool_calling_nodes):
     """
     Verify that message history is not modified after rc.call when passed to nodes constructed using different methods.


tests/rc_tests/interaction/test_call.py
```diff
@@ -3,8 +3,10 @@
 import requestcompletion as rc
 from conftest import structured_nodes, tool_calling_nodes, terminal_nodes, model
 
+NODE_INIT_METTHODS = ["class_based", "easy_wrapper"]
```
Aryan-Railtown: snuck in a quick refactor to the tests

src/requestcompletion/nodes/library/terminal_llm.py
```diff
@@ -1,8 +1,8 @@
 import warnings
-from typing import Type
-from ...llm import MessageHistory, ModelBase, SystemMessage
+from typing import Type, Dict, Any
+from ...llm import MessageHistory, ModelBase, SystemMessage, UserMessage
```
Amir-R25, line 3: what are your thoughts on these relative imports? I think at triple dot level might become harder to understand/debug in the future?

Amir-R25: totally just a suggestion btw

src/requestcompletion/nodes/library/terminal_llm.py
```diff
@@ -69,9 +71,34 @@ def __init__(
         @classmethod
         def pretty_name(cls) -> str:
             if pretty_name is None:
+                if tool_details and tool_params:
```
Amir-R25: What is the expected behaviour if tool_params is None, we still return `TerminalLLM`? This makes me think this should `or` instead of `and`? What do you think?

src/requestcompletion/nodes/library/terminal_llm.py
```diff
@@ -69,9 +71,34 @@ def __init__(
...
+                return cls(message_hist)
+                
+    if tool_details and not tool_params:
+        raise RuntimeError("Tool details provided but no tool parameters provided.")
+    elif tool_params and not tool_details:
+        raise RuntimeError("Tool parameters provided but no tool details provided.")
+    elif tool_details and tool_params:
+        if len([x.name for x in tool_params]) != len(set([x.name for x in tool_params])):
+            raise ValueError("Duplicate parameter names are not allowed")
+        warnings.warn("Tool details and tool parameters provided. TerminalLLM will be used as a tool in a tool_call_llm node.")
```
Amir-R25: Please correct me if I'm wrong but isn't this the correct case? 

If so, I don't think using a warning is the most appropriate, perhaps it could be replaced with logging

src/requestcompletion/nodes/library/terminal_llm.py
Aryan-Railtown: i agree with the suggestion, but I am not sure if github checks will complete their run with absolute paths, I think they fail. We can try to run it, if it works, i believe absolute paths might be better

src/requestcompletion/nodes/library/terminal_llm.py
Aryan-Railtown: Good catch let me make some modifications to incorporate that and change some test cases. Thank you

Aryan-Railtown: Not sure if adding logging to this lvl goes with the current implementation of the logger in RC. We can just skip the warning completely. What do you think @soulFood5632 ?

src/requestcompletion/nodes/library/terminal_llm.py
Aryan-Railtown: The checks couldn't run with absolute paths :/

src/requestcompletion/nodes/library/terminal_llm.py
Aryan-Railtown: Corrected this.

src/requestcompletion/nodes/library/terminal_llm.py
soulFood5632: Looking through the implementation I think it makes the most sense for it to not trigger any warnings or logging.

Warnings should be kept for if anything is gonna go wrong (which doesn't seem to be case here). 

Logging could make sense here but it definitely should be at the DEBUG level. However, I'd be inclined to move logging logic into the tool call LLM. My idea was that whenever a ToolCallLLM was invoked it would log the tools calls that were attatched which would serve to replace the above logging logs.

soulFood5632: I think for now skipping the warning is ok. I would just like to make sure at some point they can have visibility into the what the tool that is being injected into the LLM looks like. (although this should be a future work item)

Aryan-Railtown: That makes sense. I'll skip logging/warnings for now.

src/requestcompletion/nodes/library/_tool_call_llm_base.py
```diff
@@ -74,7 +74,7 @@ async def invoke(
 
                     tool_responses = await asyncio.gather(*contracts, return_exceptions=True)
                     tool_responses = [
-                        x if not isinstance(x, Exception) else "There was an error running the tool"
+                        x if not isinstance(x, Exception) else f"There was an error running the tool: \n Exception message: {x} "
```
soulFood5632, line 77: There probably should be some mention that this is what is happening under the hood in terms of our documentation. I also would like to see how this performs in a more general system, but I think the approach you provided should be better. 

(this is more of a general comment than something to change)

src/requestcompletion/nodes/library/terminal_llm.py
soulFood5632: Yea there are some issues around circular imports, that is why I switched to relative ones. This could be split into a work item to improve long term readability.

tests/rc_tests/interaction/test_call.py
soulFood5632: Nit pick but small spelling error in METTHODS

tests/rc_tests/interaction/test_call.py
```diff
@@ -52,7 +54,7 @@ async def make_math_game_node(message_history: rc.llm.MessageHistory):
```
soulFood5632, line 54: In general, great work on the testing front. I will be using a lot of the work here in #1.

PR #103, opened 2025-05-09
Add colorful logging

Levivus:
The logging was previously all the same color and was different across IDEs, and was difficult to read. 
This PR makes the log colors more logical, with white logs for info logs, red for errors, etc.
Additionally, it appears to be consistent for both pycharm and VSCode.

PyCharm:
image
image

VSCode:
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,7 @@ license = { file = "LICENSE" }
 classifiers = ["License :: OSI Approved :: MIT License"]
 dynamic = ["version", "description"]
 dependencies = [
+    "colorama >= 0.4.6",
     "requests >=2.31,<3",
     "pydantic >= 2.5.3,<3",
     "requests-mock >= 1.12.1",


src/requestcompletion/utils/logging/config.py
```diff
@@ -1,18 +1,57 @@
 import logging
 import warnings
 from typing import Literal, Optional
+import re
+from colorama import Fore, Style, init
```
jbueza-railtownai, line 5: does colorama need to be added to project.toml?

Levivus: yes thanks for the reminder, added now

src/requestcompletion/utils/logging/create.py
```diff
@@ -10,7 +10,7 @@ def get_rc_logger(name: str | None):
     The method is essentially a wrapper of the `logging` method to collect the logger, but it will add a reference to
     the RC root logger.
 
-    If the name is not provided it returns the root RC logger
+    If the name is not provided it returns the root RC logger.
```
soulFood5632, line 13: lol

tests/rc_tests/utils/logging/test_logging.py
```diff
@@ -3,11 +3,20 @@
 
 import pytest
 import requestcompletion as rc
+import re
 
 
 RNGNode = rc.library.from_function(random.random)
 
 
+def strip_ansi_colors(message: str) -> str:
```
soulFood5632, line 12: I don't really care right now, but we may want to have a better mechanism here. For instance if for some reason someone named a node something that matched that regex we would have a bit of an issue. This would a be a super low priority bug to deal with.

soulFood5632: nvm I lie, didn't see that this was a test

Issue #104, opened 2025-05-10
Revert to concurrent.futures Style Approach


soulFood5632:
A month or so ago we made a major change in the way that nodes call other nodes. Now we use the function called `rc.call` to interact with the system. This new tooling used the async await syntax to leverage parallelism. This approach worked well but there were a couple of awkward things that popped up from time to time that made life kinda annoying to deal with. 
- It is really tricky to enforce the requirements of async functions. 
- Sometimes you want to ensure that a function is being executed synchronously (or have any sense of locking)
- It also requires us to use an external API in `asyncio`. 

In light of all these concerns I propose we move the logic to another familiar interface seen in concurrent.futures (although note we will be abstracting the connection from the user). To see what I mean concretely check out the below example. 

Previous approach:
```python 
def node_a():
    # sequentially call node_b
    result = await rc.call(NodeB, "hello world")
    
    # now using the result above make a sequence of parallel calls to other nodes
    contracts = [rc.call(NodeC, result), rc.call(NodeD, result)]
    
    results = await asyncio.gather(*contracts)
    
    return results
 
```

New approach:
```python
def node_a():
    # sequentially call node_b
    future = rc.submit(NodeB, "hello world")
    result = future.result()

    # now using the result above make a sequence of parallel calls to other nodes
    futures = [rc.submit(NodeC, result), rc.submit(NodeD, result)]
    results = rc.gather(futures)

    return results
```

The approaches do not differ a terrible amount but there are gonna be some key features that we can now implement once we have made the switch from asyncio to concurrent.futures. I will outline the specific things that user a should be able to do within the RC framework. All of these should completed as part of this issue. 
1. As a user I should be able to launch sequential node invocations
2. As a user I should be able to run node invocations in parallel. 
3. As a user I should be able to cancel a future (see #93) [implement assuming this has been already completed]
4. As a user there should not be a large learning curve to get a handle for how these systems work 
5. As a user there should be one unified interface between calls from the interior and entrance efforts (see #102)

Issue #19, opened 2025-02-04
Come up with a better way to handle Context.


soulFood5632:
Our current mechanism to inject context into each node does not make a ton of sense. Although the strangeness of this approach is completely hidden from the user of the framework, the code in `Node` is spooky scary. 

```python 
    def __init__(
        self,
    ):
        # TODO add type checking here.
        # if no streamer is provided it will default to the null function. n
        self.data_streamer = self.__default_data_streamer
        self.context = self.__default_context()
        self._invoke_node = self.__default_invoke_node
        self.is_filled = False
        self.uuid = str(uuid.uuid4())
```

Relying on the usage of the default context object etc makes me really uneasy from a software design perspective. This needs to be rethought of completly. Note that the default streamer object is actually a similarly awkward problem, but because it isn't interacted with as much it is less of a pain. 

### General Goal
In essence we need to decide whether the inclusion of this `Context` object makes sense. 
And after deciding on that. finalize the decision so I don't have come back every 2 months to say "wtf was I thinking"

PR #106, opened 2025-05-12
102 create a unified call interface

soulFood5632:
In this PR I have added some badly needed freedom to generalize the functionality of `rc.call`. 

Previously to start a RC flow you needed to do:
```python 
with rc.Runner(...) as runner:
  runner.run(NodeA)
```
and to make a call to another node 
```python 
def node_A():
  rc.call(NodeB)
```

However, many would note that `call` and `run` are doing very similar things. So in service of that I have made it so that you can either use `rc.call` or use the runner provided and call `runner.run` (either option will work). 
So now you don't even need to use the context manager to run things.
```python 
# so instead of this:
with rc.Runner(...) as runner:
  runner.run(NodeA)

# it is just this:
rc.call(NodeA)

# or if you still want the context manager 
with rc.Runner(...) as runner:
  rc.call(NodeA)
```

All of the above equivalent except that the one with `runner.run` returns an `ExecutionInfo` objects instead of just the output. 

This new simplification lead to some complexity and thought which I will outline below:
### What do we about configurations
Previously you set the config variables when defining the runner:
```python
with rc.Runner(ExecutorConfig={...})
```
In the case when you don't use the context managers there should be some way to adjust the global config values in order to change them for that run. 
So I introduce a hierarchal approach to setting both the config values and streamer configurations: 
The default values when creating a runner will be read from the global values, if none are provided it will use the framework defaults. However, if you ever need specific config values that should not propagate globals, then you will be forced to use the context managers. The cost in writing one more line of code is definitely worth it so the user is aware that the config values they pass in are temporary. So in essence you can set global config via `rc.set_config` and if you would like to override the values for a local event you will be forced to pass them into the context manager. 

### What do we do about the Return types
One natural consequence of the above work is the issue around the return types  of `run` vs. `call`. `run` as a function returns an ExecutionInfo object which contains many reproducible details about the system allowing for all manner of future tasks. Call is designed to be modular component with no access to the above info so it just returns the output of the node that was called. In this PR I did not stray from that original approach. 
I added a piece of functionality to collect the same info object as a `@property` of the `runner`. However, if you have called the `call` object without the context manager, the runner will be killed before you can review its contents. This is not a super ideal behavior but the alternative would require a bit of a rethink of the state object since one of the salient invariants of a state object is that there is only one entry point and allowing access to many entry points would be challenging to work with. In addition there are a couple of important modules like logging etc. that need to have correct shutdown procedures to operate properly, so it is best to keep them under lock and key of the context manager. 

### Nested Runners
This is a point that @Granine brought up and after some further thought on the manner I am much less concerned about it now. I will provide an example of such a situation. Lets say you are developing a 2-part system. One is a top level system which is able to call a subserviant system. However, during development of the subservient system you write the following function:
```python 

def suberviant_flow():
  # some logic here
  pass

SubserviantTopLevel = rc.library.from_function(subserviant_flow)

def subserviant_full_task()
  with rc.Runner() as runner:
    result = runner.run(SubserviantTopLevel))

  return result.answer
```
This is fairly common becuase you will often want to your systems to modular enough so you will be testing your sub systems in RC and then stitching them together at the end. However, the way to fix this and make it more RC, is actually pretty simple. Instead of using the context manager in the `subserviant_full_task` function, move that runner logic to your testing script and do make it look the following:
```python 
def suberviant_flow():
  # some logic here
  pass

SubserviantTopLevel = rc.library.from_function(subserviant_flow)

def subserviant_full_task()

  return rc.call(SubserviantTopLevel))

# then in your testing script you can still run 
with rc.Runner() as runner:
  runner.run(SubserviantTopLevel)
```

I have designed that if you make a nested call it will throw 

```
requestcompletion.run.RunnerCreationError: A runner already exists, you cannot create nested Runners. Replace this runner with the simpler `rc.call` to call the node.
```

All in all these changes are not exactly large but they set the stage for the later changes in #104 and define a mock idea for how #19.

### DIFF
--- a/src/requestcompletion/state/execute.py
+++ b/src/requestcompletion/state/execute.py
@@ -212,7 +212,7 @@ class RCState:
         assert all(
             [n in self._node_heap for n in children]
         ), "You cannot add a request for a node which has not yet been added"
-        parent_node_name = self._node_heap.id_type_mapping[parent_node] if parent_node else "START"
+        parent_node_name = self._node_heap.id_type_mapping[parent_node] if parent_node is not None else "START"
         # to simplify we are going to create a new request for each child node with the parent as its source.
         request_ids = list(
             map(


src/requestcompletion/state/execute.py
```diff
@@ -247,7 +203,7 @@ def _create_new_request_set(
         Note that all the identifiers for the
 
         Args:
-            parent_node: The identifier of the parent node which is calling the children.
+            parent_node: The identifier of the parent node which is calling the children. If none is provided
```
Levivus: unfinished comment - "If none is provided..."

src/requestcompletion/state/execute.py
```diff
@@ -256,20 +212,18 @@ def _create_new_request_set(
         assert all(
             [n in self._node_heap for n in children]
         ), "You cannot add a request for a node which has not yet been added"
-
+        parent_node_name = self._node_heap.id_type_mapping[parent_node] if parent_node else "START"
```
Levivus: be more explicit on what this means - None

Issue #90, opened 2025-05-01
Exception handling support removed in v0.1 — need replacement mechanism


Aryan-Railtown:
## **Description:**

Prior to the v0.1 release, the `rc` framework provided several specific exceptions that allowed for fine-grained error handling, including:

```python
except rc.exceptions.ResetException
except rc.exceptions.CompletionException
except rc.exceptions.FatalException
except rc.exceptions.MalformedFunctionException
except rc.exceptions.NodeException
```

Each of these exceptions enabled developers to handle specific failure modes appropriately (e.g., distinguishing between client errors, retryable failures, and fatal system errors). For example, we could map these to relevant HTTP status codes and return actionable error messages to users.

Since the v0.1 release, it appears these exceptions have been deprecated or removed, and we're now only able to catch general exceptions. This limits the ability to respond to different failure modes in a meaningful way.

---
## **Request:**

We'd like to see some form of structured exception support reintroduced. It doesn't have to mirror the old exception classes exactly, but there should be a mechanism to distinguish between common categories of errors (e.g., user input issues, retryable failures, internal errors, etc.).

This would significantly improve debuggability, error reporting, and resilience in applications built on `rc`.

---

PR #111, opened 2025-05-20
enable tool call llms as tools

Levivus:
This PR adds functionality for LLM tools to be usable as tool calls for other LLM tools. This allows for complex agentic workflows by allowing specialized LLM tools that can be accessed as needed. 

ToolCallLLM and StructuredLLM now accept tool_details and tool_params arguments, which enable them to be treated as tools by parent agents. 

Added validation when using easy wrappers that ensure tool_details and tool_params were both included as expected if the Node is to be configured as a tool. 
Recursive tools are not validated, as it cannot be done with normal usage. The user would have to manually update the connected nodes, so this situation is not a large concern.

Added tests to verify that both class-based and functional implementations work as expected, testing all 4 permutations of easy wrapper/class implementations for the parent and child node. 

Example Usage
```
# Define a nested Notion agent
NotionAgenticTool = rc.library.tool_call_llm(
    connected_nodes={find_page_node, create_page_node, add_block_node},
    pretty_name="Notion Agentic Tool",
    system_message=rc.llm.SystemMessage("You are a Notion specialist..."),
    model=rc.llm.OpenAILLM("gpt-4o"),
    tool_details="Handles page creation, content addition, and page lookup in Notion.",
    tool_params={
        rc.llm.Parameter(
            name="notion_task_request", 
            param_type="string", 
            description="The detailed Notion task to perform."
        )
    }
)

# Compose into a higher-level Travel Agent
TravelAgent = rc.library.tool_call_llm(
    connected_nodes={convert_currency_node, available_locations_node, NotionAgenticTool},
    pretty_name="Travel Agent",
    system_message=rc.llm.SystemMessage("You are a travel planner..."),
    model=rc.llm.OpenAILLM("gpt-4o")
)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "python-dotenv >= 1.0.0",
     "pyvis >= 0.3.1",
     "llama-index >= 0.12.0",
-    "llama-index-llms-openai >= 0.3.0",
+    "llama-index-llms-openai >= 0.3.0,<=0.3.38",
     "llama-index-llms-anthropic >= 0.6.0",
     "nltk >= 3.9.1"
 ]


src/requestcompletion/nodes/library/easy_usage_wrappers/structured_llm.py
```diff
@@ -50,4 +55,31 @@ def pretty_name(cls) -> str:
             else:
                 return pretty_name
 
+        @classmethod
+        def tool_info(cls):
+            if not tool_details:
+                raise ValueError("Tool details are not provided.")
+            return Tool(
```
soulFood5632, line 62: We should have some way to present this to the user in #90

PR #119, opened 2025-05-21
Fix issues with Dynamic Function Node by trying inspect signature

Tristan-railtown:
Add try block and exception to deal with builtin functions being passed.

List any dependencies that are required for this change.

## Checklist for Author

---

## Screenshots or videos

Screenshot 2025-05-21 112243


src/requestcompletion/nodes/library/function.py
```diff
@@ -70,7 +71,12 @@ async def invoke(self) -> _TOutput:
         def _convert_kwargs_to_appropriate_types(self) -> Dict[str, Any]:
             """Convert kwargs to appropriate types based on function signature."""
             converted_kwargs = {}
-            sig = inspect.signature(func)
+
+            try:
+                sig = inspect.signature(func)
```
soulFood5632, line 75: You should add handling for all the possible exceptions. See the docs for the other issues that could pop up 

https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object

src/requestcompletion/nodes/library/function.py
```diff
@@ -70,7 +71,12 @@ async def invoke(self) -> _TOutput:
...
+            except ValueError:
+                raise RuntimeError("Cannot convert kwargs for builtin functions. " \
+                "Please use a custom function.")
```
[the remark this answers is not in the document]
soulFood5632: 🙄 , don't listen to it, I think the your exception handling is a fine approach

src/requestcompletion/nodes/library/function.py
```diff
@@ -70,7 +70,20 @@ async def invoke(self) -> _TOutput:
...
+
+            except ValueError:
+                raise RuntimeError("Cannot convert kwargs for builtin functions. " \
+                "Please use a custom function.")
+            
+            except TypeError:
+                raise RuntimeError("Argument passed is not a function and cannot be called")
+            
+            except Exception as e:
+                raise RuntimeError(f"caught unexpected exception while getting function signature: {e}")
```
jbueza-railtownai: probably want to add tests to exercise this?

src/requestcompletion/nodes/library/function.py
```diff
@@ -70,7 +70,20 @@ async def invoke(self) -> _TOutput:
...
+
+            try:
+                sig = inspect.signature(func)
+
+            except ValueError:
+                raise RuntimeError("Cannot convert kwargs for builtin functions. " \
+                "Please use a custom function.")
+            
+            except TypeError:
+                raise RuntimeError("Argument passed is not a function and cannot be called")
```
jbueza-railtownai: same same

src/requestcompletion/nodes/library/function.py
```diff
@@ -70,7 +70,20 @@ async def invoke(self) -> _TOutput:
         def _convert_kwargs_to_appropriate_types(self) -> Dict[str, Any]:
             """Convert kwargs to appropriate types based on function signature."""
             converted_kwargs = {}
-            sig = inspect.signature(func)
+
+            try:
+                sig = inspect.signature(func)
+
+            except ValueError:
+                raise RuntimeError("Cannot convert kwargs for builtin functions. " \
```
jbueza-railtownai: add unit test please

jbueza-railtownai: also, is this formatting right using black or ruff?

Tristan-railtown: Neither but I will be formatting it with black when I add the unit test.

src/requestcompletion/nodes/library/function.py
soulFood5632: Agreed

src/requestcompletion/nodes/library/function.py
Tristan-railtown: @soulFood5632 I'm thinking about reverting back to only handling the ValueError exceptions because tool.py already checks to make sure that it is callable when it checks for __qualname__. I've tried but it seems impossible to get it to throw a TypeError exception for this reason. Looking at the documentation again it is also impossible to throw any other exception because of the way we are calling inspect.signature(). I misread that the first time....

Screenshot 2025-05-21 161919

soulFood5632: Perfect, thanks for the investigation.

PR #124, opened 2025-05-27
Use mcp tools in rc

Levivus:
This PR adds the ability to use MCP tools in RC. Currently only stdio is supported, though a lot of the backbone required for http-stream is enabled, to be looked at in a later PR.

To use this functionality, use the function `from_mcp_server`, which returns a list of Tools (in RC format) for the provided server. Those tools can then be used like any tool, such as passing it into the connected_nodes of a Node. Each time one of those tools is invoked, it opens a connection to the server, and gets the response. It can later be looked at if the client should stay open for some amount of time instead of being created at invocation, but this is for now a sufficient solution.

There were problems with Windows and the async functionality that is implemented, resulting in the inability to use Streamlit, however it seems to work fine in normal use.

To see a simple demo, look at examples/test_mcp.py. The mcp servers used require either python or npx as a command. The python command needs to be modified to your own executable. npx required Node.js installed.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -27,6 +27,12 @@ jobs:
           pip install flit
           flit install --deps production --extras "test"
 
+      - name: Check pyproject.toml dependency order
+        run: |
+          python -m pip install --upgrade pip
+          pip install tomlkit==0.11.8
+          python scripts/check_dependencies_sorted.py
+
       - name: Run tests
         run: |
           export OPENAI_API_KEY=${{ secrets.INTEGRATION_TESTS_OPENAI_API_KEY }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,17 +16,17 @@ classifiers = ["License :: OSI Approved :: MIT License"]
 dynamic = ["version", "description"]
 dependencies = [
     "colorama >= 0.4.6",
-    "requests >=2.31,<3",
+    "llama-index >= 0.12.0",
+    "llama-index-llms-anthropic >= 0.6.0",
+    "llama-index-llms-openai >= 0.3.0,<=0.3.38",
+    "mcp >= 1.9.0",
+    "nltk >= 3.9.1",
+    "openai <= 1.81.0",
     "pydantic >= 2.5.3,<3",
-    "requests-mock >= 1.12.1",
     "python-dotenv >= 1.0.0",
     "pyvis >= 0.3.1",
-    "llama-index >= 0.12.0",
-    "llama-index-llms-openai >= 0.3.0,<=0.3.38",
-    "llama-index-llms-anthropic >= 0.6.0",
-    "mcp >= 1.9.0",
-    "openai <= 1.81.0",
-    "nltk >= 3.9.1"
+    "requests >=2.31,<3",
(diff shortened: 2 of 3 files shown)


examples/test_mcp.py
```diff
@@ -0,0 +1,40 @@
+import asyncio
+import requestcompletion as rc
+from requestcompletion.nodes.library.mcp_tool import from_mcp_server
+
+#%%
+TIME_MCP_COMMAND = r"C://Users//Levi//anaconda3//python.exe"
```
soulFood5632: I don't really like you having your path in the examples, maybe better to use environment variables

src/requestcompletion/nodes/library/mcp_tool.py
```diff
@@ -0,0 +1,40 @@
+from typing import Type, Optional
+from requestcompletion.utils.mcp_utils import MCPAsyncClient, from_mcp
+from requestcompletion.nodes.nodes import Node
+
+
+async def from_mcp_server(
+    command: str,
+    args: list,
+    transport_type: str = "stdio",
```
soulFood5632: Use `Literal["stdio", "http-stream"] to enforce the typing there

src/requestcompletion/nodes/library/mcp_tool.py
```diff
@@ -0,0 +1,40 @@
+from typing import Type, Optional
+from requestcompletion.utils.mcp_utils import MCPAsyncClient, from_mcp
+from requestcompletion.nodes.nodes import Node
+
+
+async def from_mcp_server(
```
soulFood5632: Having this be a async method is kinda weird. This means that to create a node type we would have `await rc.library.mcp_tool()`. This deosn't really match the previous interface and probably should.

src/requestcompletion/utils/mcp_utils.py
```diff
@@ -0,0 +1,309 @@
...
+                    "DELETE",
+                    self.base_url,
+                    headers={"Mcp-Session-Id": self.session_id, **self.http_headers}
+                )
+            except Exception:
+                pass
+            self.session_id = None
+
+
+def from_mcp(
```
soulFood5632, line 253: As per previous comment, this should probably be the one exposed to the user

src/requestcompletion/utils/mcp_utils.py
```diff
@@ -0,0 +1,309 @@
...
+    tool,
+    command: str,
+    args: list,
+    transport_type: str = "stdio",
```
soulFood5632: Same here, Use literal

src/requestcompletion/nodes/library/mcp_tool.py
```diff
@@ -0,0 +1,33 @@
...
+    Args:
+        command: Command to launch the MCP server.
+        args: Arguments for the command.
+        transport_type: 'stdio' or 'http-stream'.
+        transport_options: Optional dict for HTTP Stream configuration.
+
+    Returns:
+        List of Nodes, one for each discovered tool.
+    """
+    return asyncio.run(
```
soulFood5632, line 26: I don't think this is worth fixing now, but this will break if they ever run it a jupyter notebook or a async situation.

PR #135, opened 2025-05-29
125 mcp http stream

Levivus:
This PR adds http support to mcp, both with mcp and sse. Also added a streamlit demo, where mcp servers can be added by putting their url into the list. from_mcp_server and async_from_mcp_server are used to get the tools from a server, which now takes a config input.
image


examples/mcp_demo.py
```diff
@@ -0,0 +1,63 @@
...
+# User input
+if prompt := st.chat_input("Type your message..."):
+    st.session_state.history.append({"role": "user", "content": prompt})
+    st.chat_message("user").write(prompt)
+
+    async def get_response(history):
+        message_history = rc.llm.MessageHistory(
+            [rc.llm.UserMessage(m["content"]) if m["role"] == "user" else rc.llm.SystemMessage(m["content"]) for m in history if m["role"] != "assistant"]
+        )
+        with rc.Runner(executor_config=rc.ExecutorConfig(logging_setting="QUIET", timeout=1000)) as runner:
```
soulFood5632, line 56: 😆 Timeout isn't even operational rn.

src/requestcompletion/nodes/library/mcp_tool.py
```diff
@@ -1,33 +1,50 @@
 from typing import Type, Optional, Literal
+
+from mcp import StdioServerParameters
 from requestcompletion.nodes.nodes import Node
```
soulFood5632, line 4: I've been working with a convention to use relative imports instead of the direct package throughout. Unless you have some great reason to use the package imports I think we should keep the relative approach.

src/requestcompletion/nodes/library/mcp_tool.py
```diff
@@ -1,33 +1,50 @@
...
 import asyncio
 
-from requestcompletion.utils.mcp_utils import from_mcp_server_async
+from requestcompletion.utils.mcp_utils import MCPAsyncClient, from_mcp
```
soulFood5632, line 7: ditto

src/requestcompletion/nodes/library/mcp_tool.py
```diff
@@ -1,33 +1,50 @@
...
+
+from requestcompletion.utils.mcp_utils import MCPHttpParams
 
 
 def from_mcp_server(
-    command: str,
-    args: list,
-    transport_type: Literal["stdio", "http-stream"] = "stdio",
-    transport_options: Optional[dict] = None
+    config: Union[StdioServerParameters, MCPHttpParams],
```
soulFood5632, line 14: Total nitpick but I have been preferring we use `Type1 | Type2` instead `Union[Type1, Type2]`

src/requestcompletion/nodes/library/mcp_tool.py
```diff
@@ -1,33 +1,50 @@
...
         List of Nodes, one for each discovered tool.
     """
     return asyncio.run(
-        from_mcp_server_async(
-            command,
-            args,
-            transport_type=transport_type,
-            transport_options=transport_options
-        )
+        async_from_mcp_server(config)
```
soulFood5632, line 26: Looks much better

src/requestcompletion/utils/mcp_utils.py
```diff
@@ -1,271 +1,115 @@
...
-            self.http_headers = self.transport_options.get("headers", {})
-            self.endpoint = self.transport_options.get("endpoint", "/mcp")
-            self.base_url = self.transport_options.get("base_url", "http://localhost:8080") + self.endpoint
-            self.response_mode = self.transport_options.get("responseMode", "batch")
-            self.session_id = None
+        elif isinstance(self.config, MCPHttpParams):
+            if not self.sse_client:
+                try:
+                    await self._init_session(streamablehttp_client)
+                except Exception:  # McpError
```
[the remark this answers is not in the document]
soulFood5632: I agree with copilot friend unless you don't know how to know what the error is.

PR #147, opened 2025-06-02
pinned llama index core at 12.38

soulFood5632:
There was a issue with the most recent release of `llama-index-core`

image

Specifcally the "Add `require_tool` param broke things. I have pinned our requirements to not download this newest version.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install pytest pytest-asyncio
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -17,6 +17,7 @@ dynamic = ["version", "description"]
 dependencies = [
     "colorama >= 0.4.6",
     "llama-index >= 0.12.0",
+    "llama-index-core >= 0.12.0, <= 0.12.38", # Pinned because of an issue in 0.12.39
```
jbueza-railtownai, line 20: why not just do == 0.12.38? does it impact ci or something if it isn't a range?

PR #149, opened 2025-06-03
134 add oauth support to mcp

Levivus:
125: 
This PR adds http support to mcp, both with mcp and sse. Also added a streamlit demo, where mcp servers can be added by putting their url into the list. from_mcp_server and async_from_mcp_server are used to get the tools from a server, which now takes a config input.
134:
This add OAuth support to MCP, allowing for servers such as paypal which require you to log in in the browser. 
Currently client are still only active when invoking a tool, which means that it needs to reauthorize each tool call, but for now we will leave it stateless and think about managing that later.

image

### DIFF
--- a/tests/rc_tests/nodes/library/test_mcp.py
+++ b/tests/rc_tests/nodes/library/test_mcp.py
@@ -48,7 +48,7 @@ def test_from_mcp_server_with_llm():
         response = asyncio.run(runner.run(parent_tool, message_history=message_history))
 
     assert response.answer is not None
-    assert response.answer.content is not "It didn't work!"
+    assert response.answer.content != "It didn't work!"
 
 
 def test_from_mcp_server_with_http():


tests/rc_tests/nodes/library/test_mcp.py
```diff
@@ -1,49 +1,78 @@
 import asyncio
 import pytest
+from unittest.mock import patch
```
soulFood5632: Unused import I think

tests/rc_tests/nodes/library/test_mcp.py
```diff
@@ -1,49 +1,78 @@
...
 import requestcompletion as rc
+from mcp import StdioServerParameters
 from requestcompletion.nodes.library.mcp_tool import from_mcp_server
 from requestcompletion.nodes.nodes import Node
 
 import pytest
```
soulFood5632, line 7: Double import of pytest. Really not sure why the linter is not catching these things.

Issue #131, opened 2025-05-28
open-source routing - cloud providers


Amir-R25:
As the framework is maturing and we are migrating to LiteLLM for model routing, we should start with a few integrations for popular open-source models hosted on cloud providers.

Possible Models: Llama, Deepseek, etc

Possible Providers: VertexAI, Bedrock, Azure

Additionally, not every open-source model is also fine-tuned for tool calling. 

It's important to test to see and find which models also support the feature offerings and handle through exceptions/warnings for ones that do not.

Related: #126

Issue #153, opened 2025-06-04
Investigate Local Testing Issues


soulFood5632:
Inside the RC package we use relative imports and inside the testing module we use the package itself to test things. 

However, we have run into issues where sometimes the assertions are failing because we are comparing types from the local files in `src` to those in `site-packages`. The reason for the behavior is known to be affected by the setup of the environment you are working in. However, we do not know the root cause. 

This issue should be a mainly researched focused issue where we answer the question, "How does one properly download run the tests referencing the correct package"

PR #156, opened 2025-06-04
Ollama Models

Amir-R25:
Added support for users being able to use local models using Ollama.

Since different models have different capabilities, checks were added for tool calling support. Additionally, the user needs to make sure the model they intend to use has already been pulled in their Ollama server.

This was done as a first step towards addressing #131 , next step would be models hosted on different cloud environments.

Example usage:

```python

MODEL_NAME = "ollama_chat/llama3.2:latest"

def get_todays_date(tell_time: bool):
    """
    Returns the correct date once called. Time can also be provided

    Args:
        tell_time (bool): if set to True will also return time
    """
    import datetime

    if tell_time:
        return str(datetime.datetime.now())
    else:
        return str(datetime.date.today())

DateAgent = rc.library.tool_call_llm(
    pretty_name="Date Agent",
    system_message=rc.llm.SystemMessage("You are an agent capable of telling date and time"),
    model = rc.llm.OllamaLLM(MODEL_NAME),
    connected_nodes={
        rc.nodes.library.from_function(get_todays_date)
    }
)

with rc.Runner() as runner:
    result = runner.run_sync(DateAgent, message_history=rc.llm.MessageHistory([rc.llm.UserMessage("What is today's date and current time?")]))

print(result.answer)
```

### DIFF
--- a/src/requestcompletion/llm/models/local/ollama.py
+++ b/src/requestcompletion/llm/models/local/ollama.py
@@ -1,5 +1,6 @@
 import os
 import requests
+from typing import Literal
 
 import litellm
 
@@ -7,6 +8,7 @@ from .._litellm_wrapper import LiteLLMWrapper
 from ....utils.logging.create import get_rc_logger
 
 LOGGER_NAME = "OLLAMA"
+DEFAULT_DOMAIN = "http://localhost:11434"
 
 
 class OllamaError(Exception):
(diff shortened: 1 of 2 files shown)


src/requestcompletion/llm/models/local/ollama.py
```diff
@@ -0,0 +1,54 @@
+import os
+import requests
+
+import litellm
+
+from .._litellm_wrapper import LiteLLMWrapper
+from ....utils.logging.create import get_rc_logger
```
soulFood5632, line 8: Yea this a bit ridiculous. We definetly need to think about a better way in #153

src/requestcompletion/llm/models/local/ollama.py
```diff
@@ -0,0 +1,54 @@
...
+
+LOGGER_NAME = "OLLAMA"
+
+
+class OllamaError(Exception):
+    pass
+
+
+class OllamaLLM(LiteLLMWrapper):
```
soulFood5632, line 18: Would you mind adding some basic docs here saying the requirements of ollama's `__init__`

src/requestcompletion/llm/models/local/ollama.py
```diff
@@ -0,0 +1,54 @@
...
+    @classmethod
+    def model_type(cls):
+        return "Ollama"
+
+    def __init__(self, model_name: str, **kwargs):
+        super().__init__(model_name, **kwargs)
+        self.domain = os.getenv("OLLAMA_HOST", "http://localhost:11434")
```
soulFood5632: I'd prefer if this was passed in as an optional argument in the __init__, better to be explicit about these types of inputs

Amir-R25: did a small redesign on how to cover this thoroughly, let me know what you think

soulFood5632: Yea looks good to me

Issue #162, opened 2025-06-06
Add Configurability of Max Tool calls


soulFood5632:
## Tell me what you would like to see
You should be able to configure the number of max tool calls you want to have in a `ToolCallLLM`. Right now it is set to a default value and is not difficult to configure as a user. 

```python 
def __init__(
        self,
        message_history: MessageHistory,
        model: ModelBase,
        max_tool_calls: int = 10,
    ):
```

## Additional context
Anything else you think is relevant.

Issue #91, opened 2025-05-01
We need better timeout handling logic.


CoronRing:
# intro 

It would be ideal to make timeout handling easier, currently, the timeout is 25 seconds, but if a user is debuting or using a long generation, this can fail. It can also fail as awkward locations like when a lock is held, breaking the concurrency guarantee, causing a deadlock. 

# potential solution

We can make the time infinite in the short term,

Create some "work mode" like production vs debug mode in the medium term,

or implement the complex management with state in the long run.

Issue #69, opened 2025-03-31
Move streaming function into ExecutorConfig


soulFood5632:
I have clarified the line between the `ExectutorInfo` and `ExecutorConfig` object. 

Simplified the line is: Any objects which could be classified as configurable parameters which would not effect the state of the run should be put in the config object. Basically anything passive should be placed in the config object. Otherwise the objects should be placed in the `info` object. 

This new definition largely fits what we were doing previously, however, our streamer is getting passed as in an additional parameter in Runner object in addition to the config object. 

```python 
class Runner:

    def __init__(
        self,
        subscriber: Callable[[str], None] = None,
        executor_config: ExecutorConfig = ExecutorConfig(),
    ):
```

Ideally we would like to move the passed in subscriber logic into the config object. Specifically that means adding a parameter in the ExecutorConfig object which is a method. 

If there are issues surrounding this related to the saving of the config object please flag them in this issue and we will make a larger design decision around how we want to deal with that.

PR #171, opened 2025-06-09
Updated Configuration

soulFood5632:
In the following PR I made various updates to the functionality of the configuration details. 

1. Max Number of Calls Guardrail in `ToolCallLLM` #162 
When creating a tool call llm you can say the max number of tool calls you want. You can also adjust the default max level which is currently defaulted to 30. It can be configured in both the easy usage and class based initializations. 
2. Moved Streaming setting into the config object. #69 
To set a streaming handler you now must use the `ExecutorConfig` object. 
```python 
with rc.Runner(rc.ExecutorConfig(subscriber=handler.handle)) as run:
        result = run.run_sync(StreamingNode)
        assert result.answer == None
```
3. Reimplented Timeout Functionality #91 
You can now provide a top level timeout variable to prevent long running loops and failures.

### DIFF
--- a/src/requestcompletion/nodes/library/_tool_call_llm_base.py
+++ b/src/requestcompletion/nodes/library/_tool_call_llm_base.py
@@ -27,26 +27,18 @@ class OutputLessToolCallLLM(Node[_T], ABC, Generic[_T]):
         self,
         message_history: MessageHistory,
         model: ModelBase,
-        max_tool_calls: int | None = None,
+        max_tool_calls: int | None = 30,
     ):
         super().__init__()
         self.model = model
         self.message_hist = deepcopy(message_history)
         self.structured_resp_node = None  # The structured LLM node
 
-        if max_tool_calls is None:
-            max_tool_calls = self.default_max_tool_calls()
         self.max_tool_calls = max_tool_calls
 
     @abstractmethod
     def connected_nodes(self) -> Set[Type[Node]]: ...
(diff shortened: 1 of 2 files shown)


src/requestcompletion/config.py
```diff
@@ -3,44 +3,34 @@
 
 from pydantic import BaseModel, Field
 
+from typing import Callable, Coroutine
+
 from .utils.logging.config import allowable_log_levels
 
 
 # There may be moments when you are confused what should be in this config object vs. what should
```
soulFood5632: Delete this comment

src/requestcompletion/nodes/library/_tool_call_llm_base.py
```diff
@@ -27,17 +27,27 @@ def __init__(
...
         message_history: MessageHistory,
         model: ModelBase,
-        max_tool_calls: int = 30,
+        max_tool_calls: int | None = None,
     ):
         super().__init__()
         self.model = model
         self.message_hist = deepcopy(message_history)
         self.structured_resp_node = None  # The structured LLM node
+
```
Amir-R25, line 36: do we really need `default_max_tool_calls` to be a separate function here?

soulFood5632: Yea I can understand your reasoning. It has been updated.

PR #176, opened 2025-06-10
Sync Functions -> to_thread

soulFood5632:
In the following PR, we introduce better functionality to introduce blocking synchronized functions. Now they are wrapped in `asyncio.to_thread` which allows the IO blocking tasks to be operated parallelly. 

I also fixed some bad logic in the dynamic function node which converted bad types in the invoke method instead of handling it in the `prepare_node` function (that is the one designed for that type of behavior). 

```python 
import time


def sync_blocking(timeout: float):
    time.sleep(timeout)


async def async_blocking(timeout: float):
    """
    An async function that blocks for a given timeout.
    """
    await rc.library.from_function(sync_blocking)(timeout)


Blocking = rc.library.from_function(sync_blocking)

lengths = [1, 2, 2, 2, 1]


async def top_level_sync_blocking():
    """
    A top-level function that blocks for a given timeout.
    """
    await rc.batch(Blocking, lengths)
    return None


TopLevel = rc.library.from_function(top_level_sync_blocking)

with rc.Runner() as runner:
    start_time = time.time()
    runner.run_sync(TopLevel)
    end_time = time.time()
    print(f"Sync blocking task {lengths} took {end_time - start_time:.2f} seconds")
````

In the following script the blocking `time.sleep` did not cause issues and the script ran in ~2 seconds. This is the behavior we are looking for. ✅ 

## Future Work
There is some future work tied to this item. Currently we do not support sync calls to `rc.call` which could make sense if users want to just use sequential code. We may want to open up that functionality so that they can work with it in that context.
```python 
def some_func():
  # sequential call to NodeB
  result = rc.call(NodeB)
```

### DIFF
--- a/tests/rc_tests/run/test_streaming.py
+++ b/tests/rc_tests/run/test_streaming.py
@@ -54,9 +54,7 @@ def test_slow_streamer():
             self.finished_message = item
 
     sub = Sub()
-    with rc.Runner(
-        executor_config=ExecutorConfig(subscriber=sub.handle)
-    ) as runner:
+    with rc.Runner(executor_config=ExecutorConfig(subscriber=sub.handle)) as runner:
         finished_result = runner.run_sync(StreamingRNGNode)
 
     assert isinstance(finished_result.answer, float)


scripts/evaluations/sync_async_benchmark.py
```diff
@@ -0,0 +1,35 @@
+import requestcompletion as rc
+import time
+
+
+def sync_blocking(timeout: float):
+    time.sleep(timeout)
+
+
+async def async_blocking(timeout: float):
```
Amir-R25: making note for removal: cc: @soulFood5632

soulFood5632: fixed 😄

PR #187, opened 2025-06-12
Immutable Context

soulFood5632:
In this simple phase I have implemented basic functionality for immutable context for RC to allow for key value pair storage of variables. 

For instance if you need some credentials passed down for some reason you can use the following logic. 

```python 
def some_node():
  ...
  p_id = rc.context.get("project_id")
  ...

context_vars = {"project_id": "some value"}

with rc.Runner(context=context_vars) as run:
  run.run(...)
```

This simple context object supports one simple behavior `get()` which can use a default parameter if you would not like to throw an error when the key doesn't exist. 

Below is the new `get` function specs.
```python 
def get(
    key: str,
    default: Any | None = None,
):
    """
    Get a value from the context object
    
    Args:
        key (str): The key to retrieve.
        default (Any | None): The default value to return if the key does not exist.
    Returns:
        Any: The value associated with the key, or the default value if the key does not exist.
        
    Raises:
        KeyError: If the key does not exist and no default value is provided.
    """
    return protected_context.get(key, default=default)
```

### DIFF
--- a/src/requestcompletion/context/central.py
+++ b/src/requestcompletion/context/central.py
@@ -63,6 +63,7 @@ def delete_globals():
 
 def get(
     key: str,
+    /,
     default: Any | None = None,
 ):
     """
@@ -70,7 +71,7 @@ def get(
 
     Args:
         key (str): The key to retrieve.
-        default (Any | None): The default value to return if the key does not exist.
+        default (Any | None): The default value to return if the key does not exist. If set to None and the key does not exist, a KeyError will be raised.
     Returns:
         Any: The value associated with the key, or the default value if the key does not exist.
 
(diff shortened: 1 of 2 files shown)


src/requestcompletion/context/central.py
```diff
@@ -0,0 +1,95 @@
...
+        key (str): The key to retrieve.
+        default (Any | None): The default value to return if the key does not exist.
+    Returns:
+        Any: The value associated with the key, or the default value if the key does not exist.
+
+    Raises:
+        KeyError: If the key does not exist and no default value is provided.
+    """
+    context = external_context.get()
+    return context.get(key, default=default)
```
Levivus, line 82: It looks like this won't ever raise KeyError since you're using .get().

PR #194, opened 2025-06-16
 Allow String System Messages for Easy Usage

### LINKED ISSUE #181: Allow str for System Messages of Library types
This is a good catch for now. We will want to allow string type objects in a issue I need to create.

_Originally posted by @soulFood5632 in https://github.com/RailtownAI/rc/pull/178#discussion_r2138925443_

We should be flexible and allow string inputs for system messages for our easy usage and class based internal node types.

Aryan-Railtown:
## Summary
Quick fix to improve developer experience by accepting both string and SystemMessage objects for system messages in easy usage wrappers.

### Flexibility Enhancement
- **Updated system message validation** to accept `str` or `SystemMessage` objects
- **Modified exception handling** to support dual input types

### Testing Updates
- **Updated 2 test cases** to verify string input behavior
- **Ensured backward compatibility** with existing SystemMessage objects
- **Validated error handling** for invalid input types

## Usage Examples
```python
# Now both work:
_ = rc.library.terminal_llm(
        system_message="You are a helpful assistant that can structure the response into a structured output.",
        model=rc.llm.OpenAILLM("gpt-4o"),
        pretty_name="Terminal",
    )  # ✅ New
_ = rc.library.terminal_llm(
        system_message=rc.llm.SystemMessage("You are a helpful assistant that can structure the response into a structured output."),
        model=rc.llm.OpenAILLM("gpt-4o"),
        pretty_name="Terminal",
    ) # ✅ Still works
```

This small change removes friction for developers while maintaining the structured approach for advanced use cases.
## Checklist for Author

This should close #181 .

### DIFF
--- a/src/requestcompletion/exceptions/node_creation/validation.py
+++ b/src/requestcompletion/exceptions/node_creation/validation.py
@@ -118,6 +118,7 @@ def _check_pretty_name(pretty_name: str | None, tool_details: Any) -> None:
             get_message(ExceptionMessageKey.MISSING_PRETTY_NAME_MSG)
         )
 
+
 def _check_system_message(system_message: SystemMessage | str | None) -> None:
     """
     Validate that system_message is an instance of SystemMessageType if provided.
@@ -127,10 +128,13 @@ def _check_system_message(system_message: SystemMessage | str | None) -> None:
     Raises:
         NodeCreationError: If system_message is not of the correct type.
     """
-    if system_message is not None and not isinstance(system_message, (SystemMessage, str)):
+    if system_message is not None and not isinstance(
+        system_message, (SystemMessage, str)
+    ):
         raise NodeCreationError(
             get_message(ExceptionMessageKey.INVALID_SYSTEM_MESSAGE_MSG),
         )
+
 
 def _check_tool_params_and_details(tool_params: Any, tool_details: Any) -> None:
     """


src/requestcompletion/exceptions/node_invocation/validation.py
```diff
@@ -14,17 +16,19 @@ def check_message_history(message_history: MessageHistory) -> None:
...
         )
-    elif len(message_history) == 0:
+    elif len(message_history) == 0 and system_message is None:
         raise NodeInvocationError(
             message="Message history must contain at least one message",
             notes=["Please provide an initial message"],
             fatal=True,
         )
-    elif message_history[0].role != "system":
+    elif len(message_history) > 0 and message_history[0].role != "system":
```
soulFood5632, line 25: I think I missed this in #90, but I think you are not supposed to input a system message as part of the message history becuase the toolCallLLM automatically inserts it. Lets raise this as a separate issue an modify the logic.

Aryan-Railtown: I believe toolCallLLM  automatically inserts the system message in case of easy_usage wrapper, but this function is also being used to check the message_history for class based inits, where the user might insert the system message in the `__init__`. Something like this:

```
class Encoder(rc.library.TerminalLLM): 
        def __init__(
                self,
                message_history: rc.llm.MessageHistory,
                model: rc.llm.ModelBase = model,
            ):
                message_history = [x for x in message_history if x.role != "system"]
                message_history.insert(0, rc.llm.SystemMessage("You are a helpful assistant that can encode text into bytes."))
                super().__init__(
                    message_history=message_history,
                    model=model,
                )
        @classmethod
        def pretty_name(cls) -> str:
            return "Simple Node"
```

PR #196, opened 2025-06-17
192 make global context mutable

### LINKED ISSUE #192: Make global context mutable
After much consideration, it has been decided that stage 1 of global context implementation is simply making the context mutable so that it can be added to by any node.

One of the main reasons behind it being mutable initially was that mutability would make state reversion much more difficult, however that is not a current priority. Additionally, mutability is more pythonic, allowing developers to use it as they want to - though we will encourage best practices.

In the future, the mutability can be upgraded if needed, such as for allowing a lock to prevent race conditions.

Levivus:
Small PR that adjusts the global context to be mutable. Really not much else to it, adds implementation for `put`, and replaces `define` with `update` to better reflect functionality.


soulFood5632 (changes requested):
Looks good. I don't see how you exposed the `put` method to the user.

soulFood5632 (changes requested):
Sorry I should reviewed more completely last time.  We should defintely have tests for this to make sure that the put method is working as expected.  Specifically in the integration tests (not unit tests).

soulFood5632 (commented):
LGTM

PR #198, opened 2025-06-17
Add @to_node decorator for simplified node creation

### LINKED ISSUE #190: Make a Wrapper which maps functions to nodes
To make a node we currently have to do the following:
```python
def some_func(...):
   pass

Node = rc.library.from_function(some_func)
```

It would be nice if we could make a wrapper which accomplished the same thing:

```python 
@rc.to_node
def some_func(...):
  pass
```

As before we should allow for the same customization but this should reduce some boiler plate.

Aryan-Railtown:
## Summary
This PR introduces a new `@to_node` decorator that provides a more convenient way to create nodes from functions, reducing boilerplate code.
### Core Implementation
- **`function.py`**: Added `to_node` decorator function that wraps `from_function`
- **`__init__.py`**: Added `to_node` to the `__all__` list for public API exposure

### Testing
- Added unit tests to verify the decorator behavior
- Ensures decorator produces equivalent results to `from_function`

## Implementation Details
The `to_node` decorator internally calls `rc.library.from_function`, preserving all existing functionality while providing a more Pythonic interface for node creation.

## Usage Example
```python
# Before
def my_function(x, y):
    return x + y
Node = rc.library.from_function(my_function)

# After  
@rc.to_node
def my_function(x, y):
    return x + y
```
## Checklist for Author

---

Should close #190

### DIFF
--- a/tests/rc_tests/nodes/library/test_function.py
+++ b/tests/rc_tests/nodes/library/test_function.py
@@ -21,6 +21,23 @@ import requestcompletion as rc
 MODEL_PROVIDERS = ["openai"]
 
 
+# ===== Unit Tests =====
+def test_to_node():
+    """Test that to_node decorator works correctly."""
+    @rc.to_node
+    def secret_phrase() -> str:
+        """
+        Function that returns a secret phrase.
+
+        Returns:
+            str: The secret phrase.
+        """
+        return "Constantinople"
+
+    assert issubclass(secret_phrase, rc.Node)
+    assert secret_phrase.pretty_name() == "secret_phrase Node"
+
+
 # ===== Test Classes =====
 class TestPrimitiveInputTypes:
     @pytest.mark.parametrize("model_provider", MODEL_PROVIDERS)


src/requestcompletion/exceptions/node_creation/exception_messages.yaml
```diff
@@ -7,7 +7,7 @@ CONNECTED_NODES_EMPTY_MSG: "connected_nodes must not return an empty set."
 CONNECTED_NODES_EMPTY_NOTES:
   - "Please provide a set nodes that can be used as tools by the ToolCallLLM node."
 
-CONNECTED_NODES_TYPE_MSG: "connected_nodes must return a set of Nodes or functions."
+CONNECTED_NODES_TYPE_MSG: "connected_nodes must return a set of Nodes."
```
Aryan-Railtown, line 10: Just noticed this while testing for `to_node`. Slipped my min in the prev PR.

PR #207, opened 2025-06-18
133 integrate logging

soulFood5632:
In phase 1 of this challenge I have improved our logging system to make the system a little bit less cluttered and added file logging support. 

```python 
with rc.Runner(ExecutorConfig(logging_setting="REGULAR", log_file="path")
```

To interact with our logger system you should now use `get_logger(name: str)`. 

**Notes:**
- Currently, we we will only put logs of level INFO into the file but if people are interested in more configurablity we can add that as needed. 
- The connecting to railtowns logging suite should be pretty easy from here. I want figure out how we specifically want to make that work. It will tackled in a future item. 
- I have reduced the amount of info in the logs for a very specific reason. Keeping the command line clean and driving users to use our debugging tools will probably be for the best.

**Other Items**
- I have also added some debug statements that should help us to get to the bottom of issues quicker from a dev standpoint. 

## Examples
**Verbose Mode**
image

**Regular Mode**
image

**Quiet Mode**
image

**File Log**
image

### DIFF
--- a/src/requestcompletion/info.py
+++ b/src/requestcompletion/info.py
@@ -94,7 +94,7 @@ class ExecutionInfo:
 
             # we need to quickly check to make sure these ids are valid
             for identifier in ids:
-                if identifier not in self.node_heap:
+                if identifier not in self.request_heap:
                     raise ValueError(
                         f"Identifier '{identifier}' not found in the current state."
                     )


tests/integration_tests/interaction/test_batch.py
```diff
@@ -12,7 +12,7 @@
         ([1, 2, 3, 2, 1], 3.5, 0.75),
         ([1, 5, 1], 5.25, 0.5),
         ([1] * 35, 1.25, 0.5),
-        ([2] * 100 + [3] * 50, 3.5, 0.75),
+        ([2] * 100 + [3] * 50, 3.5, 1),
```
Aryan-Railtown: ran into this issue in my local haha. Should be good now.

PR #208, opened 2025-06-19
Tristan/201/create-sample-Notion-Agent

Tristan-railtown:
Created Small Notion-Agent that utilizes Notion's MCP features to help update and edit Notion pages specified by a user.

List any dependencies that are required for this change.

## Checklist for Author

---

## Screenshots or videos

Drag and drop your images (.png, .jpg) or videos (.mp4) here to show what was implemented.

### DIFF
--- a/examples/sample_agents/notion_agent.py
+++ b/examples/sample_agents/notion_agent.py
@@ -38,6 +38,7 @@ def notion_agent(  # noqa: C901
 
     Args:
         notion_api_token: your notion api token, if not passed this will be taken from you .env
+        notion_version: For you to specify the NotionMCP server notion if it is not up to date
         model:
             The type of LLM you would like to use
             output_type: Specifies what portion of the conversation to return. You can choose between


examples/sample_agents/notion_agent.py
```diff
@@ -0,0 +1,183 @@
...
+            else:
+                return self.message_hist[-1]
+
+        def __init__(
+            self,
+            message_history: MessageHistory,
+            llm_model: ModelBase | None = None,
+            max_tool_calls: int | None = 30,
+        ):
+            check_message_history(
```
soulFood5632: I think this is a lot of repeated code here. This `__init__` logic is already a part of the parent class so I don't think we want to repeat it here (DRY [Don't repeat yourself]).

examples/sample_agents/Notion_Demo.py
soulFood5632, line 1: file names should not have capitols, https://peps.python.org/pep-0008/#package-and-module-names

soulFood5632, line 1: I also think we should throw these demo scripts in a seperate file for now. Don't worry too much about it being presentable but we should have them separated for now.

examples/sample_agents/notion_agent.py
```diff
@@ -0,0 +1,183 @@
...
+from typing_extensions import Self
+from requestcompletion.exceptions import NodeCreationError
+from requestcompletion.exceptions.node_creation.validation import validate_tool_metadata
+from requestcompletion.exceptions.node_invocation.validation import (
+    check_model,
+    check_message_history,
+)
+import requestcompletion as rc
+
+load_dotenv()
```
soulFood5632: We shouldn't be loading environment variables here. It is already done in the root `__init__` file so it is not needed.

examples/sample_agents/notion_agent.py
soulFood5632, line 1: This is a general comment but, you should probably leverage parent classes to handle some of the mcp integration nodes. They are all gonna have the same basic structure where they do this.    
```python
tools = rc.nodes.library.from_mcp_server(
        StdioServerParameters(
            command=MCP_COMMAND,
            args=MCP_ARGS,
            env={
                "OPENAPI_MCP_HEADERS": f'{{"Authorization": "Bearer {notion_api_token}", "Notion-Version": "2022-06-28" }}'
            },
        )
    )
    connected_nodes = {*tools}
```

The only difference I see as of right now would be the system prompt and any specific configurations related to this system. Basically everything you have as a global variable should the be input parameters for that new base class. 

For an example of what I am talking about check out the ` _tool_call_llm_base`.

examples/sample_agents/notion_agent.py
soulFood5632: Upon further review, this isn't your bad at all. I will create an issue to create a base class that is inherited by (the easy usage wrappers) `NotionAgent`, `ToolCallLLM` etc.

examples/sample_agents/notion_agent.py
```diff
@@ -0,0 +1,183 @@
...
+                name=cls.pretty_name().replace(" ", "_"),
+                detail=tool_details,
+                parameters=tool_params,
+            )
+
+        @classmethod
+        def prepare_tool(cls, tool_parameters: Dict[str, Any]) -> Self:
+            message_hist = MessageHistory(
+                [
+                    UserMessage(f"{param.name}: '{tool_parameters[param.name]}'")
```
soulFood5632: I don't really love this logic but I know you used it from the other easy usage wrappers. I will create a ticket to be smarter about this.

examples/sample_agents/notion_agent.py
soulFood5632: See my top level comment for this review, but this is just more food for thought.

There may be the ability to create one single base class which creates a `ToolCallLLM` with a provided MCP server. This may make it so your implementations of each sample `Notion`, `Github` is only as simple as passing in the correct configurations and using the base method. If you think this is doable then we should do that becuase otherwise we are gonna have a ton of copied code.

examples/sample_agents/notion_agent.py
```diff
@@ -0,0 +1,183 @@
...
+                    raise self.structured_output
+                return self.structured_output
+            elif output_type == "MessageHistory":
+                return self.message_hist
+            else:
+                return self.message_hist[-1]
+
+        def __init__(
+            self,
+            message_history: MessageHistory,
```
soulFood5632: Another minor comment for discussion. What would you think about making the input a `str` "instruction" instead of a message history. I am thinking that could make more sense for something like a notion agent that should be accomplishing a specific task. 

Let me know your thoughts

examples/sample_agents/notion_agent.py
```diff
@@ -0,0 +1,183 @@
...
+    ):  # TODO: add support for MessageHistory output type with output_model. Maybe resp.answer = message_hist and resp.structured = model response
+        raise NotImplementedError(
+            "MessageHistory output type is not supported with output_model at the moment."
+        )
+    tools = rc.nodes.library.from_mcp_server(
+        StdioServerParameters(
+            command=MCP_COMMAND,
+            args=MCP_ARGS,
+            env={
+                "OPENAPI_MCP_HEADERS": f'{{"Authorization": "Bearer {notion_api_token}", "Notion-Version": "2022-06-28" }}'
```
Amir-R25: let's make the date an input parameter with this default value so when Notion updates their MCP version it'll be easier

examples/sample_agents/notion_agent.py
```diff
@@ -0,0 +1,183 @@
...
+)
+import requestcompletion as rc
+
+load_dotenv()
+MCP_COMMAND = "npx"
+MCP_ARGS = ["-y", "@notionhq/notion-mcp-server"]
+SYSTEM_MESSAGE = """You are a master notion page designer. You love creating beautiful and well-structured Notion pages and make sure that everything is correctly formatted."""
+
+
+def notion_agent(  # noqa: C901
```
Amir-R25, line 28: can you please add docstrings

examples/sample_agents/notion_agent.py
Tristan-railtown: So don't add any changes to this then?

examples/sample_agents/Notion_Demo.py
Tristan-railtown: Like another folder called demos in examples?

examples/sample_agents/Notion_Demo.py
```diff
@@ -0,0 +1,27 @@
+import os
+from dotenv import load_dotenv
+import requestcompletion as rc
+from requestcompletion.llm import MessageHistory, UserMessage
+from notion_agent import notion_agent
+import asyncio
+
+load_dotenv()
```
Amir-R25: I guess load_dotenv isn't needed here either

examples/sample_agents/Notion_Demo.py
soulFood5632: yea that would work

examples/sample_agents/notion_agent.py
soulFood5632: Yea I have created a separate issue to deal with it.

examples/sample_agents/notion_agent.py
Tristan-railtown: I'll look into this

examples/sample_agents/Notion_Demo.py
Tristan-railtown: This has been taken care of

Tristan-railtown: Done

examples/sample_agents/notion_agent.py
Tristan-railtown: No longer loading in dotenv

examples/sample_agents/notion_agent.py
Tristan-railtown: I've implemented an _mcp_agent_base in library to accomplish this

examples/sample_agents/notion_agent.py
Tristan-railtown: That sounds good to me. I've changed the message_history to instructions and it's now a string that _mcp_agent_base deals with on the back end

examples/sample_agents/notion_agent.py
Tristan-railtown: Done

examples/sample_agents/Notion_Demo.py
Tristan-railtown: Done

examples/sample_agents/notion_agent.py
Tristan-railtown: All done

examples/demos/notion_demo.py
```diff
@@ -0,0 +1,23 @@
+import requestcompletion as rc
+from requestcompletion.llm import MessageHistory, UserMessage
+from ..sample_agents.notion_agent import notion_agent
```
soulFood5632, line 3: hmmm, these sample agents exist outside the package structure. This isn't an issue for this PR, but I will put one up there to deal with it ASAP.

examples/demos/notion_demo.py
```diff
@@ -0,0 +1,23 @@
...
+"""To see better results initialize NOTION_ROOT_PAGE_ID variable in your .env file with the ID of the page you want to use as the parent page for your operations.
+Otherwise you can specify the title of the parent page in the system message"""
+
+USER_PROMPT = """Would you be able to create a new page for me in Notion labeled "Jokes"? I would like to have a small joke at the top of the page. Anything but the scientist joke please."""
+
+
+notion_agent = notion_agent(model=rc.llm.OpenAILLM("gpt-4o"))
+
+with rc.Runner() as run:
+    result = asyncio.run(
```
soulFood5632, line 18: Instead of asyncio.run you should just use `run.run_sync`

examples/sample_agents/notion_agent.py
soulFood5632: Awesome work

examples/demos/notion_demo.py
soulFood5632: Issue has been posted in #232

examples/demos/notion_demo.py
Tristan-railtown: This leads to an error saying that coroutine Runner.run was never awaited

examples/demos/notion_demo.py
Tristan-railtown: Would it not be ok to create a sample_agents folder in library the same way we have an easy_usage folder?

soulFood5632: That is a probably what we will land on but lets not worry about that for now.

Issue #200, opened 2025-06-17
Add Prompt functionality


Levivus:
Add prompt functionality from local config and mcp.

PR #222, opened 2025-06-24
218 feature request token usage

soulFood5632:
We can now track token usage during the run of a node. This data will be saved as part of the node in its debug information. 

Each node will contain a new field called `self._details` which is a dictionary which can be interacted with as needed to store data. In a future PR, we will enforce serialization constraints (see #__). 

In another section of this PR we add the functionality of pre and post llm hooks. You can now add a function which modifies `MessageHistory`, which should be quite helpful in #200. We can also add hooks to run after collecting the response. This should prove very useful in future endeavors. In this PR's case we attach a post llm hook which intercepts the message and stores the data inside the node. 

As of right now I'm storing the least possible information just message history and model name and history. This information allows us to completely figure out token and cost usage for each node for each call in its history. In essence it is the most low level information we could use to track token usage.

### DIFF
--- a/src/requestcompletion/llm/models/_litellm_wrapper.py
+++ b/src/requestcompletion/llm/models/_litellm_wrapper.py
@@ -224,7 +224,6 @@ class LiteLLMWrapper(ModelBase, ABC):
 
         litellm_tools = [_to_litellm_tool(t) for t in tools]
 
-        kwargs.setdefault("tool_choice", "auto")
         kwargs["tools"] = litellm_tools
 
         resp = self._invoke(messages, **kwargs)
--- a/examples/mcp_example.py
+++ b/examples/mcp_example.py
@@ -1,18 +1,16 @@
+#%%
 import asyncio
-#%%
 import requestcompletion as rc
-#%%
 from mcp import StdioServerParameters
-
 from requestcompletion.nodes.library.mcp_tool import async_from_mcp_server, from_mcp_server
 
 #%%
 # Install mcp_server_time for time tools:
-MCP_COMMAND = "uv"
-MCP_ARGS = ["--directory", r"C:\Users\Levi\Documents\MCP\node\node_server.py", "run", "node_server.py"]
+MCP_COMMAND = "uvx"
+MCP_ARGS = ["mcp-server-time"]
 # Airbnb MCP server requires Node.js and the `npx` command to run.
 
-
+# 
 #%%
 # Discover all tools
 fetch_tools = from_mcp_server(StdioServerParameters(command=MCP_COMMAND, args=MCP_ARGS))
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,13 @@ lint.mccabe.max-complexity = 10
 
 
 [project.optional-dependencies]
-test = ["pytest >=7.1.2"]
+test = [
+    "pytest >=7.1.2",
+    "pytest-asyncio",
+    "flit",
+    "tomlkit==0.11.8",
+    "pytest-timeout"
+]
 
 [tool.pytest.ini_options]
 pythonpath = ["src"]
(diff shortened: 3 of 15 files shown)


src/requestcompletion/llm/models/_litellm_wrapper.py
```diff
@@ -181,12 +177,12 @@ def _invoke(
             model=self._model_name, messages=litellm_messages, stream=stream, **merged
         )
 
-    def chat(self, messages: MessageHistory, **kwargs) -> Response:
+    def _chat(self, messages: MessageHistory, **kwargs) -> Response:
```
Amir-R25, line 180: What's the reason for making `_chat` and `_chat_with_tools` private?

src/requestcompletion/llm/model.py
```diff
@@ -16,22 +15,160 @@
...
+        """Adds a post-hook to modify the response after receiving it from the model."""
+        self._post_hook.append(hook)
+
+    def add_exception_hook(
+        self, hook: Callable[[MessageHistory, Exception], None]
+    ) -> None:
+        """Adds an exception hook to handle exceptions during model interactions."""
+        self._exception_hook.append(hook)
+
+    def remove_pre_hooks(self) -> None:
```
Amir-R25, line 62: Not sure when this will come in to play (Removing hooks)?

src/requestcompletion/llm/models/_litellm_wrapper.py
```diff
@@ -252,3 +248,9 @@ def __str__(self) -> str:
         if len(parts) == 2:
             return f"LiteLLMWrapper(provider={parts[0]}, name={parts[1]})"
         return f"LiteLLMWrapper(name={self._model_name})"
+
```
Amir-R25, line 251: should probably be a property method

src/requestcompletion/llm/models/local/ollama.py
```diff
@@ -83,8 +78,8 @@ def _run_check(self, endpoint: str):
 
             model_names = {model["name"] for model in models["models"]}
 
-            if self.model_name not in model_names:
-                error_msg = f"{self.model_name} not available on server {self.domain}. Avaiable models are: {model_names}"
+            if self.model_name() not in model_names:
```
Amir-R25, line 81: once model_name is a property method these can be removed

src/requestcompletion/llm/model.py
soulFood5632: It comes into to play in a really weird bug that pops up to do some deep copying shenanigans.

src/requestcompletion/llm/models/_litellm_wrapper.py
soulFood5632: So I can properly attach the hooks inside the base class instead of delegating that responsibility the subclasses.

src/requestcompletion/llm/models/_litellm_wrapper.py
soulFood5632: I didn't love the consistency of making this property and then `model_type` a `@classmethod`. Making one a `property` and the other having to accessed via method made me feel queasy.

src/requestcompletion/llm/models/local/ollama.py
soulFood5632: See above

src/requestcompletion/llm/model.py
```diff
@@ -16,22 +15,160 @@
...
+
     def chat(self, messages: MessageHistory, **kwargs) -> Response:
+        """Chat with the model using the provided messages."""
+        for hook in self._pre_hook:
+            hook(messages)
+        try:
+            result = self._chat(messages, **kwargs)
+        except Exception as e:
+            for hook in self._exception_hook:
+                hook(messages, e)
```
Levivus, line 97: Will the result of exception hooks also need to be assigned to messages? If so I can add it to my PR along with pre_hook.

soulFood5632: You handle this in your PR :)

PR #223, opened 2025-06-24
200 add prompt functionality

Levivus:
This PR allows for prompt injection from context. 
When a message is added with fstring-like formatting, such as `"View the following text: {text}"`, it will automatically be populated from the global context from that key. If you want to have {} normally, use double brackets such as "{{text}}".
Alternatively, this behavior can be turned off by setting prompt_injection to False in the ExecutorConfig.

### DIFF
--- a/tests/integration_tests/library/test_function.py
+++ b/tests/integration_tests/library/test_function.py
@@ -65,7 +65,7 @@ class TestPrimitiveInputTypes:
             Returns:
                 str: The result of the function.
             """
-            return f"str(input_num)" * input_num
+            return str(input_num) * input_num
 
         agent = create_top_level_node(
             magic_number,


src/requestcompletion/config.py
```diff
@@ -35,3 +36,4 @@ def __init__(
         self.subscriber = subscriber
         self.run_identifier = run_identifier if run_identifier else str(uuid.uuid4())
         self.log_file = log_file
+        self.prompt_injection = prompt_injection
```
soulFood5632, line 40: We should have proper docs showing exactly what prompt injection means/does

src/requestcompletion/llm/message.py
```diff
@@ -27,7 +27,10 @@ class Message(Generic[_T]):
     """
 
     def __init__(
-        self, content: _T, role: Literal["assistant", "user", "system", "tool"]
+        self,
+        content: _T,
+        role: Literal["assistant", "user", "system", "tool"],
+        inject_prompt: bool = True,
```
soulFood5632, line 33: same here we need some docs

src/requestcompletion/llm/message.py
```diff
@@ -58,6 +62,20 @@ def role(self) -> Role:
...
 
+    @property
+    def inject_prompt(self) -> bool:
+        """
+        A boolean that indicates whether this message should be injected into the prompt.
+        This is used to control whether the message is included in the prompt or not.
+        """
+        return self._inject_prompt
+
+    def set_inject_prompt(self, value: bool):
```
soulFood5632: I'd use `@inject_prompt.setter` here instead

src/requestcompletion/prompts/prompt.py
```diff
@@ -0,0 +1,33 @@
+import string
+from ..context import get
+from ..context.central import get_config
+from ..llm import MessageHistory, Message
+
+
+def fill_prompt(prompt: str) -> str:
+    class ContextDict(dict):
+        def __getitem__(self, key):
+            return get(key)
```
soulFood5632: For readability sake, can you use `rc.context.get`

src/requestcompletion/prompts/prompt.py
```diff
@@ -0,0 +1,33 @@
+import string
+from ..context import get
+from ..context.central import get_config
+from ..llm import MessageHistory, Message
+
+
+def fill_prompt(prompt: str) -> str:
+    class ContextDict(dict):
```
soulFood5632: Lets move this outside the function defintion and make it a private class. No need to use run-time defined classes if we don't need to.

src/requestcompletion/pubsub/publisher.py
```diff
@@ -87,6 +87,7 @@ async def publish(self, message: _T):
 
         """
         # logger.debug(f"Publishing message: {message}")
+        print(f"Publishing message: {message}, {str(message)}")
```
soulFood5632: loose print statement

tests/rc_tests/prompt/test_prompt.py
```diff
@@ -0,0 +1,60 @@
...
+        model=MockLLM(chat=return_message)
+    )
+
+    with rc.Runner(context={"secret_value": "tomato"}) as runner:
+        response = runner.run_sync(node, message_history=MessageHistory())
+
+    assert response.answer == "{secret_value}"
+
+
+def test_prompt_injection_global_config_bypass():
```
soulFood5632, line 80: Make sure to add some of those test cases around the edge cases we worked through with `{0}` etc.

soulFood5632: This needs to be rock solid for an edge case perspective

PR #233, opened 2025-06-25
Unified Tool Calling LLM with Configurable Call Limits

### LINKED ISSUE #80: Single Tool Call Agent
There are potential use cases where you want to limit an LLM to a single tool call and then force it to give a response. In reality their are infinite different ways you could configure an LLM in such a manner however, this ticket will simplify the work to just allowing 1 (or n if that is easy) set of tool calls before turning off the ability to call tools. 

This is similar to what Langchain has already implemented. 

# Completion Requirements
- [ ] Implement a new library node which is limited to a single tool call before returning
- [ ] If you think it would be easy to create a bit more configurability to allow n tool calls feel free implement that node as well. 
- [ ] As always  a bit of testing

Aryan-Railtown:
## Summary
Implements unified tool calling behavior with configurable limits and graceful degradation, replacing inconsistent exception-throwing with forced final answers when limits are reached.

## Changes
- **Unified Interface**: Merged `ToolCallLLM` and `LimitedToolCallLLM` into single `ToolCallLLM` class
- **Graceful Degradation**: Replaced exception throwing with forced final answer when `max_tool_calls` reached
- **Smart Defaults**: `max_tool_calls=30` by default to prevent runaway billing, `max_tool_calls=None` for unlimited (with warning)
- **Code Organization**: Refactored monolithic `tool_call_llm` into modular folder structure:
  - `base.py` - Base functionality
  - `structured_tool_call_llm.py` - Structured tool calling
  - `message_hist_tool_call_llm.py` - Message history tool calling
- **Test Improvements**: 
  - Updated tests for unified behavior
  - Reorganized tests to match new structure
  - Moved integration tests from `rc_tests` to proper integration folder
  - Added comprehensive unit tests for all tool call LLM variants
  - Improved test performance through better separation

## Testing
- ✅ All existing tests pass
- ✅ New unit tests for limited tool calling functionality  
- ✅ Integration tests properly categorized
- ✅ Faster unit test execution

### Easy Usage Wrapper
```python
rc.library.tool_call_llm(
    connected_nodes=tool_nodes,
    pretty_name="Tool Call Node",
    system_message=sys_msg,
    model=model,
    max_tool_calls=1,  # Defaults to 30, set to None for unlimited (with warning)
)
```

### Class-based Approach
```python
class CustomToolCallNode(ToolCallLLM):
    def __init__(self, message_history, model=model):
        message_history.insert(0, sys_msg)
        super().__init__(message_history, model, max_tool_calls=1)
    
    @classmethod
    def connected_nodes(cls):
        return {tool1, tool2}
    
    @classmethod
    def pretty_name(cls):
        return "Custom Tool Call Node"
```

## Completion Requirements Met
- [x] Unified tool calling interface with configurable limits
- [x] Graceful degradation instead of exceptions
- [x] Smart defaults (30) with unlimited option (None)
- [x] Comprehensive testing coverage

## Checklist for Author

This closes #80

### DIFF
--- a/src/requestcompletion/exceptions/node_creation/validation.py
+++ b/src/requestcompletion/exceptions/node_creation/validation.py
@@ -6,6 +6,7 @@ from pydantic import BaseModel
 from ...llm import SystemMessage
 import warnings
 
+
 def validate_function(func: Callable) -> None:
     """
     Validate that the function is safe to use in a node.


src/requestcompletion/exceptions/node_invocation/validation.py
```diff
@@ -48,3 +48,16 @@ def check_model(model: ModelBase):
             ],
             fatal=True,
         )
+
+
+def check_max_tool_calls(max_tool_calls: int):
+    if max_tool_calls < 0:
+        raise NodeInvocationError(
+            message="max_tool_calls must be a non-negative integer.",
```
soulFood5632: This is a bit of a violation of DRY when you have already defined the text in the yaml file

src/requestcompletion/nodes/library/easy_usage_wrappers/tool_call_llm.py
```diff
@@ -50,7 +48,7 @@ def tool_call_llm(  # noqa: C901
         )
 
     # If a function is passed in, we will convert it to a node
-    for elem in connected_nodes:
+    for elem in list(connected_nodes):
```
soulFood5632, line 50: What was the reason for this change?

src/requestcompletion/exceptions/node_creation/exception_messages.yaml
```diff
@@ -31,6 +32,12 @@ MISSING_TOOL_DETAILS_MSG: "Tool parameters are provided, but tool details are mi
 MISSING_TOOL_DETAILS_NOTES:
   - "If you want to use this node as a tool, you must provide tool details."
 
+MAX_TOOL_CALLS_NEGATIVE_MSG: "max_tool_calls must be a non-negative integer."
+MAX_TOOL_CALLS_NEGATIVE_NOTES:
+  - "The default value for any general RC tool_call_llm node is 30."
```
soulFood5632: This interface is a bit confusing. To summarize we have a `ToolCallLLM` which defaults to 30 which will throw an exception if it ever reaches that number. Additionally, we have `LimitedToolCallLLM` which defaults to 1 which will force a regular chat after 1 tool call. 

I am half thinking we should just make it all one type since that makes a bit more sense with the default behavior being unlimited number of tool calls. 

Let me know what you are thinking in regards to this @Aryan-Railtown

src/requestcompletion/nodes/library/easy_usage_wrappers/tool_call_llm.py
```diff
@@ -29,18 +28,17 @@ def tool_call_llm(  # noqa: C901
...
+    max_tool_calls: int | None = None,
     system_message: SystemMessage | str | None = None,
     output_type: Literal["MessageHistory", "LastMessage"] = "LastMessage",
     output_model: BaseModel | None = None,
     tool_details: str | None = None,
     tool_params: dict | None = None,
 ) -> Type[OutputLessToolCallLLM[Union[MessageHistory, AssistantMessage, BaseModel]]]:
     if output_model:
-        OutputType = output_model  # noqa: N806
+        output = output_model
```
soulFood5632, line 38: I think I am gonna write to the PEP masters asking for this to be changed as a best pratice😄. I hate type variables being non-capitalized so damn much.

soulFood5632: I think this is fine btw, just my personal pet peeve.

src/requestcompletion/nodes/library/tool_calling_llms/_base.py
```diff
@@ -1,19 +1,23 @@
 import asyncio
 from typing import TypeVar, ParamSpec, Generic, Set, Type, Dict, Any, Union, Callable
 from copy import deepcopy
-from ..nodes import Node
-from ...llm import (
+from ...nodes import Node
+from ....llm import (
```
soulFood5632: Lets use the package import so its `rc.llm` here

src/requestcompletion/nodes/library/tool_calling_llms/_base.py
```diff
@@ -1,19 +1,23 @@
...
     MessageHistory,
     ModelBase,
     ToolCall,
     ToolResponse,
     ToolMessage,
     UserMessage,
+    AssistantMessage,
 )
-from ...interaction.call import call
+from ....interaction.call import call
```
soulFood5632: same here

src/requestcompletion/nodes/library/tool_calling_llms/_base.py
```diff
@@ -1,19 +1,23 @@
...
 from abc import ABC, abstractmethod
-from ...exceptions import NodeCreationError, LLMError
-from ...exceptions.node_invocation.validation import check_message_history
+from ....exceptions import NodeCreationError, LLMError
+from ....exceptions.node_invocation.validation import (
```
soulFood5632: same

src/requestcompletion/nodes/library/tool_calling_llms/_base.py
```diff
@@ -30,6 +36,12 @@ def __init__(
         max_tool_calls: int | None = 30,
     ):
         super().__init__(model=model, message_history=message_history)
+        self.model = model
+        check_message_history(
```
Aryan-Railtown: little oversight while merging, fixing this. llm_base checks for message hist. Tool_calling base checks for max_tool_calls

src/requestcompletion/exceptions/node_creation/validation.py
```diff
@@ -206,11 +206,27 @@ def _check_tool_params_and_details(tool_params: Any, tool_details: Any) -> None:
...
 
+def _check_max_tool_calls(max_tool_calls: int | None) -> None:
+    """
+    Ensure max_tool_calls is a non-negative integer.
+    Args:
+        max_tool_calls: The maximum number of tool calls allowed.
+    Raises:
+        NodeCreationError: If max_tool_calls is negative.
+    """
+    if max_tool_calls is not None and max_tool_calls < 0:
```
Amir-R25: catching those troll users :D

src/requestcompletion/exceptions/node_creation/exception_messages.yaml
soulFood5632: I think we could stick with the current interface I just worry about it being confusing at the moment.

src/requestcompletion/nodes/library/tool_calling_llms/limited_tool_call_llm.py
```diff
@@ -0,0 +1,47 @@
...
+            String representing the final answer.
+        """
+        super().__init__(message_history, model, max_tool_calls=max_tool_calls)
+        self.max_tool_calls = max_tool_calls
+        assert self.message_hist[0].role == "system"
+        # append a message to the system message, so that the LLM can plan out the tool calls
+        self.message_hist[0] = SystemMessage(
+            self.message_hist[0].content
+            + "\n"
+            + f"You are only allowed to make a maximum of {self.max_tool_calls} tool call(s) in total. "
```
soulFood5632: I don't super love slapping in extra details into a system message. Users could be a bit bothered if we are modifying there system message. Did you run into issues before adding this prompt?

src/requestcompletion/nodes/library/tool_calling_llms/mess_hist_tool_call_llm.py
```diff
@@ -1,5 +1,5 @@
-from ._tool_call_llm_base import OutputLessToolCallLLM
-from ...llm import MessageHistory
+from ._base import OutputLessToolCallLLM
+from ....llm import MessageHistory
```
soulFood5632: I think same goes here, lets use the package instead of relative imports

src/requestcompletion/nodes/library/tool_calling_llms/structured_tool_call_llm.py
```diff
@@ -1,9 +1,8 @@
-from typing import Type
-
-from ._tool_call_llm_base import OutputLessToolCallLLM
-from ...llm import MessageHistory, ModelBase, SystemMessage
-from .easy_usage_wrappers.structured_llm import structured_llm
+from ._base import OutputLessToolCallLLM
+from ....llm import MessageHistory, ModelBase, SystemMessage
```
soulFood5632: same here

src/requestcompletion/exceptions/node_creation/validation.py
Aryan-Railtown: 😆

src/requestcompletion/nodes/library/easy_usage_wrappers/tool_call_llm.py
Aryan-Railtown: Yup, ruff wouldn't let me keep it capital haha

src/requestcompletion/nodes/library/tool_calling_llms/limited_tool_call_llm.py
Aryan-Railtown: The prompt was more for the llm to plan head, because in most cases where the llm needs to make more than `x` tool_calls and the limit is less than `x`, the llm kind of has to improvise because of the tool_calls list getting chopped off. Removing this prompt has a negative effect on the quality of the response, but not by alot. Do you think it is better for us to just not do this?

src/requestcompletion/exceptions/node_creation/exception_messages.yaml
Aryan-Railtown: I agree, it is a bit confusing. Honestly, we can unify this into a single `ToolCallLLM` class:

• remove excception throwing and replace with forced final answer (same as LimitedToolCallLLM), so when `max_tool_calls` is reached, force a chat response instead of erroring
• if `max_tool_calls` is not provided we can keep `max_tool_calls=30` as default to prevent runaway billing (learned from our OpenAI incident)
• `max_tool_calls=None` enables unlimited calls with a warning to user

that would be a lot cleaner, we can remove the limitedToolcallLLM class. What do you think @soulFood5632

src/requestcompletion/exceptions/node_invocation/validation.py
Aryan-Railtown: good catch. I will move the exception messages 1 level up so that both validation files can make use of it.

src/requestcompletion/nodes/library/easy_usage_wrappers/tool_call_llm.py
Aryan-Railtown: oh i just realized `connected_nodes` is a set. Just good practice to traverse a list ig.

src/requestcompletion/exceptions/node_creation/exception_messages.yaml
soulFood5632: Yea that sounds good to me. Honestly lets up the max tool calls to 60 for the time being. But I agree with the that redesign.

src/requestcompletion/nodes/library/tool_calling_llms/limited_tool_call_llm.py
soulFood5632: I'm a bit torn because I can understand how that would come up. Lets leave it in there for now. We can always remove it later if we determine to do that then.

src/requestcompletion/exceptions/node_creation/exception_messages.yaml
Aryan-Railtown: Okay I'll will get to working on that then.

soulFood5632: Sorry for making this ticket a bit roundabout. I didn't really think about how those 2 additions would work together.

Aryan-Railtown: All good, we have the logic down now, it's just a little refact at this point.

PR #234, opened 2025-06-25
Context and Global Config scoping to Runner

soulFood5632:
In this PR, I have completed the following items:
1. Restructured the context object so we have one object which contains `runner_id`, `InternalContext`, and `ExternalContext`. 
2. Rethought the logic around the connection between the runner object and global Execucutor config + local Executor config object. (Noting that local is in reference to the one that belongs to this runner).
3. Applied the new updates to the context object as well.

### DIFF
--- a/src/requestcompletion/run.py
+++ b/src/requestcompletion/run.py
@@ -10,7 +10,7 @@ from .config import ExecutorConfig
 from .context.central import (
     register_globals,
     delete_globals,
-    set_global_config, get_global_config, set_local_config,
+    get_global_config,
 )
 from .execution.coordinator import Coordinator
 from .execution.execution_strategy import AsyncioExecutionStrategy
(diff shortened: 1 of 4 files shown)


tests/rc_tests/context/test_central.py
Levivus, line 1: Is this just an empty file?

soulFood5632: Yea I am slowly adding files for unit testing but I realized that file elicits some of the stupidest unit tests so I didn't even bother.

src/requestcompletion/run.py
```diff
@@ -75,18 +73,12 @@ def __init__(
     ):
         # first lets read from defaults if nessecary for the provided input config
```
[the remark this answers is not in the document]
soulFood5632: This had already been solved in a different PR

PR #239, opened 2025-06-30
LiteLLM supports async

soulFood5632:
Many of our modules that rely on LLM interaction were hilariously non-optimized. They were essentially operating as large blocking tasks and were not using async await "parallelism". I have added new functionality in the llm modules so that we can now run `achat` which is an async method with identical functionality to that of `chat`. I have created an async pair for all 4 methods `chat`, `chat_with_tools`, `stream_chat`, and `structured_chat`. 

Now when running flows with LLM's we will have proper interleaving which should have a major increase in performance in multi LLM flows.

### DIFF
--- a/src/requestcompletion/llm/models/_litellm_wrapper.py
+++ b/src/requestcompletion/llm/models/_litellm_wrapper.py
@@ -284,7 +284,6 @@ class LiteLLMWrapper(ModelBase, ABC):
     def _stream_chat(self, messages: MessageHistory, **kwargs) -> Response:
         stream_iter, info = self._invoke(messages, stream=True, **kwargs)
 
-
         return self._stream_handler_base(stream_iter)
 
     async def _astream_chat(self, messages: MessageHistory, **kwargs) -> Response:


src/requestcompletion/llm/models/_litellm_wrapper.py
```diff
@@ -177,36 +177,117 @@ def _invoke(
             model=self._model_name, messages=litellm_messages, stream=stream, **merged
         )
 
-    def _chat(self, messages: MessageHistory, **kwargs) -> Response:
-        raw = self._invoke(messages, **kwargs)
+    async def _ainvoke(
```
Aryan-Railtown, line 196: Theese 2 functions `_invoke` and `_ainvoke_` share about 90% of the code, do you think it'd be a good idea to try to make a `_prepare_invoke` or `_prepare_litellm_call` to extract that code, or would that make it a little too DRY?

Aryan-Railtown: I get why we don't want to make this just one `_invoke` function, always a good idea to keep async and sync stuff separate. Ultimately, it is a trade-off, so it is your call. Other that this the PR looks great :)

soulFood5632: Yea its kinda annoying, the entire file is a bit of a DRY nightmare. but its largely because async and sync functions really do not play nicely.

PR #245, opened 2025-07-02
RAG system and test

CoronRing:
# intro
Added RAg system that embed text or file in memory

## Feature
- chunking
- in memory
- factor
- schema
- lots of base class
- type checks
- work with file and string

## Issues
- core_node contains integration with RC, however, it is currently in raw state and not tested due to discussion to migratethe  easy-usage wrapper to __init_sub__

## Performance test
- system is functional for RAG corpus of size 1280 document

### DIFF
--- a/tests/rc_tests/rag/test_embedding_service_unit.py
+++ b/tests/rc_tests/rag/test_embedding_service_unit.py
@@ -17,13 +17,6 @@ def patch_litellm(monkeypatch):
 
 from requestcompletion.RAG.embedding_service import BaseEmbeddingService, EmbeddingService
 
-def test_base_service_not_implemented():
-    base = BaseEmbeddingService()
-    with pytest.raises(NotImplementedError):
-        base.embed("abc")
-    with pytest.raises(NotImplementedError):
-        base.embed_many(["a", "b"])
-
 def test_repr_includes_class_and_model():
     es = EmbeddingService(model="test-model")
     assert "EmbeddingService" in repr(es)


src/requestcompletion/RAG/rag_node.py
```diff
@@ -0,0 +1,112 @@
...
+            def pretty_name(cls) -> str:
+                return "Easy RAG Node"
+
+            def invoke(self) -> list:
+                return self.result
+
+        return RAGNodeMini
+
+
+if __name__ == "__main__":
```
CoronRing: This file is am example use ported over from easy-usage-wrapper, in next version we will change easy usage wrapper logic, hence this file is left for reference only. that is why there is duplicated function and main here

src/requestcompletion/RAG/embedding_service.py
```diff
@@ -0,0 +1,93 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class BaseEmbeddingService:
+    """
+    Base class for embedding services.
+    """
+
+    def embed(self, text: str, **kwargs) -> List[float]:
```
soulFood5632: pedantic but lets keep this consisent across the code base and use `abstractmethod` from `abc`

src/requestcompletion/RAG/rag_node.py
soulFood5632, line 1: I think for this PR lets modify this file to be the following:

```python 
def rag_node(documents: list, model, config):
    rag_core = RAG(
        docs=documents,
        embed_config={
            "model": model,
        },
        store_config={},
        chunk_config={"chunk_size": 100, "chunk_overlap": 20, "model": "gpt-4o"},
    )
    rag_core.embed_all()

    def query(query: str) -> list:
        result = rag_core.search(query, top_k=2)
        return result

    return rc.library.from_function(query)
```

src/requestcompletion/RAG/rag_node.py
soulFood5632: Took a look here I have some comments below.

Issue #215, opened 2025-06-20
Shared Code in easy usage wrappers


soulFood5632:
In the easy-usage-wrappers we have a lot of copied code where we check message histories while creating the class. The code is fairly hardcoded in nature but we should have one central location that prepares classes of this type to avoid repeating ourselves (DRY principle). 

This lack of centralized approach means that in ever new agent we deliver we have to implement manually copying over vast swaths of code. 

For instance:
```python 
    class ToolCallLLM(OutputLessToolCallLLM[OutputType]):
        def return_output(self):
            if output_model:
                if isinstance(self.structured_output, Exception):
                    raise self.structured_output
                return self.structured_output
            elif output_type == "MessageHistory":
                return self.message_hist
            else:
                return self.message_hist[-1]

        def __init__(
            self,
            message_history: MessageHistory,
            llm_model: ModelBase | None = None,
            max_tool_calls: int | None = 30,
        ):
            check_message_history(
                message_history, system_message
            )  # raises NodeInvocationError if any of the checks fail
            message_history_copy = deepcopy(message_history)
            if system_message is not None:
                if len([x for x in message_history_copy if x.role == Role.system]) > 0:
                    warnings.warn(
                        "System message already exists in message history. We will replace it."
                    )
                    message_history_copy = [
                        x for x in message_history_copy if x.role != Role.system
                    ]
                    message_history_copy.insert(0, system_message)
                else:
                    message_history_copy.insert(0, system_message)

            if llm_model is not None:
                if model is not None:
                    warnings.warn(
                        "You have provided a model as a parameter and as a class variable. We will use the parameter."
                    )
            else:
                check_model(
                    model
                )  # raises NodeInvocationError if any of the checks fail
                llm_model = model

            super().__init__(
                message_history_copy, llm_model, max_tool_calls=max_tool_calls
            )

            if output_model:
                system_structured = SystemMessage(
                    "You are a structured LLM that can convert the response into a structured output."
                )
                self.structured_resp_node = structured_llm(
                    output_model, system_message=system_structured, model=llm_model
                )

        def connected_nodes(self) -> Set[Union[Type[Node], Callable]]:
            return connected_nodes

        @classmethod
        def pretty_name(cls) -> str:
            if pretty_name is None:
                return (
                    "ToolCallLLM("
                    + ", ".join([x.pretty_name() for x in connected_nodes])
                    + ")"
                )
            else:
                return pretty_name

        @classmethod
        def tool_info(cls):
[...]

PR #279, opened 2025-07-10
Local Chat Interface

### LINKED ISSUE #231: [Feature Request] Chat easy usage wrapper
## Tell me what you would like to see
Currently our methods are "one-shot" meaning that you interact with the agent/workflow once and you get an output.

What we want is to have the option of interacting with the agent/flow in a chat conversation.

This entails the following considerations:
1. UI choice for inputting user messages -> ie command line, localhost
    Additional tabs in the UI -> Tool Use/ Reasoning/ Debugging graph
3. Internal design for optimal efficiency
4. Potential concerns for portability## Additional context

Amir-R25:
This PR addresses the need for when the desired Agent should be capable of having a conversation with the user that persists it's state.

It uses localhost to run a chat server so the user can talk to their agent. 

Note: Unit tests were only written for `chat_ui.py` since the easy usage wrapper methodology is undergoing significant refactoring.

Example:
```python
import random
import datetime
from requestcompletion.nodes.library.easy_usage_wrappers.chat_tool_call_llm import (
    chat_tool_call_llm,
)


def get_todays_date(tell_time: bool, tell_location: bool):
    """
    Returns the correct date once called. Time can also be provided.

    Args:
        tell_time (bool): if set to True will also return time
    """

    if tell_location:
        raise ValueError("Location is not supported in this example.")

    if tell_time:
        if random.random() < 0.8:
            raise RuntimeError("Random exception occurred!")
        return str(datetime.datetime.now())
    else:
        return str(datetime.date.today())


ChatBot = chat_tool_call_llm(
    port=5000,
    auto_open=True,
    llm_model=rc.llm.OpenAILLM("gpt-4o"),
    pretty_name="ChatBot",
    system_message=rc.llm.SystemMessage(
        "You are an agent capable of answering user questions."
    ),
    connected_nodes={get_todays_date},
)

with rc.Runner(rc.ExecutorConfig(timeout=600)) as runner:
    resp = runner.run_sync(
        ChatBot,
        rc.llm.MessageHistory(),
    )
```

This sample code also makes one of the tools throw an error randomly so that part of the interface can be visualized as well.

Closes #231 
## Checklist for Author

---

## Screenshots or videos
image
image

### DIFF
--- a/src/requestcompletion/nodes/_node_builder.py
+++ b/src/requestcompletion/nodes/_node_builder.py
@@ -230,6 +230,15 @@ class NodeBuilder(Generic[_TNode]):
         self,
         chat_ui: ChatUI,
     ):
+        """
+        Configure a chat UI for the node.
+
+        Starts the chat UI server asynchronously and sets it as an override
+        for the node being built.
+
+        Args:
+            chat_ui (ChatUI): The chat UI instance to configure for this node.
+        """
         chat_ui.start_server_async()
         self._with_override("chat_ui", chat_ui)
 


pyproject.toml
```diff
@@ -16,6 +16,7 @@ classifiers = ["License :: OSI Approved :: MIT License"]
 dynamic = ["version", "description"]
 dependencies = [
     "colorama >= 0.4.6",
+    "fastapi >= 0.104.0",
```
soulFood5632, line 19: Do we want to attach the chat dependencies to a different list

src/requestcompletion/nodes/library/easy_usage_wrappers/chat_tool_call_llm.py
```diff
@@ -0,0 +1,240 @@
...
+                    ],
+                )
+        elif isfunction(elem):
+            connected_nodes.remove(elem)
+            connected_nodes.add(from_function(elem))
+
+    chat_ui = ChatUI(port=8000)
+    server_address = chat_ui.start_server_async()
+
+    class ChatToolCallLLM(OutputLessToolCallLLM[output]):
```
soulFood5632: See my comment but we are definitely going to want to move this class logic out of the easy usage wrapper. Waiting on #215 for those changes to occur.

src/requestcompletion/visuals/browser/chat_ui.py
```diff
@@ -0,0 +1,214 @@
...
+            else:
+                user_msg = await self.user_input_queue.get()
+                
+            return user_msg.get("message") if user_msg else None
+            
+        except asyncio.TimeoutError:
+            return None
+
+    def run_server(self):
+        uvicorn.run(self.app, host="0.0.0.0", port=self.port, log_level="warning")
```
soulFood5632: I don't see this used anywhere but I think hardcoding the host ip is a bit scary

Amir-R25: it's used in the threading on line 211. I'll make it a default value

Amir-R25: using 127.0.0.1 for more security for default value

pyproject.toml
Amir-R25: It's just fastapi, not sure. I'm inclined to think it's ok for now

jbueza-railtownai: rc shouldn't know about fastapi if devs don't need it, same with streamlit etc

i'd move all this into a separate repo or into the examples folder with a separate project toml

https://kellysutton.com/2017/05/29/deletability.html

src/requestcompletion/nodes/library/easy_usage_wrappers/chat_tool_call_llm.py
```diff
@@ -0,0 +1,240 @@
...
+                    message=f"Tools must be of type Node or FunctionType but got {type(elem)}",
+                    notes=[
+                        "Please make sure you are passing in a function or a Node object to connected_nodes"
+                    ],
+                )
+        elif isfunction(elem):
+            connected_nodes.remove(elem)
+            connected_nodes.add(from_function(elem))
+
+    chat_ui = ChatUI(port=8000)
```
jbueza-railtownai: this should be an environment varaible

devs might have another server running on 8000 and it just won't run due to conflicting ports

src/requestcompletion/visuals/browser/chat_ui.py
```diff
@@ -0,0 +1,216 @@
...
+                        # Send heartbeat
+                        yield f"data: {json.dumps({'type': 'heartbeat', 'timestamp': datetime.now().isoformat()})}\n\n"
+            
+            return StreamingResponse(
+                event_generator(),
+                media_type="text/event-stream",
+                headers={
+                    "Cache-Control": "no-cache",
+                    "Connection": "keep-alive",
+                    "Access-Control-Allow-Origin": "*",
```
jbueza-railtownai: probably want to set safer CORS headers, like just allow localhost

src/requestcompletion/visuals/browser/chat_ui.py
```diff
@@ -0,0 +1,216 @@
...
+            content = self._get_static_file_content("chat.html")
+            return HTMLResponse(content)
+        
+        @app.get("/chat.css")
+        async def get_chat_css():
+            """Serve the chat CSS file"""
+            content = self._get_static_file_content("chat.css")
+            return Response(content, media_type="text/css")
+        
+        @app.get("/chat.js")
```
jbueza-railtownai, line 141: hollonnow

do we really need a python server to serve a static asset? i thought fastapi already serves static assets automatically if you put them in the same directory as the html file

pyproject.toml
Amir-R25: thanks Jaime, but fastapi is used by the local chat interface, so idea is devs will be using to for their PoC chatbots?

src/requestcompletion/nodes/library/easy_usage_wrappers/chat_tool_call_llm.py
Amir-R25: Great catch, thanks, I'll make the addition

pyproject.toml
soulFood5632: Yea Its almost like the chat functionality should be a seperate package just like our LLM module should also be one. Same goes for all of additional plug ins like RAG, etc.

src/requestcompletion/nodes/_node_builder.py
```diff
@@ -225,6 +226,13 @@ def mcp_llm(self, mcp_command, mcp_args, mcp_env, max_tool_calls):
         self._with_override("connected_nodes", classmethod(lambda cls: connected_nodes))
         self._with_override("max_tool_calls", max_tool_calls)
 
+    def chat_ui(
+        self,
+        chat_ui: ChatUI,
+    ):
+        chat_ui.start_server_async()
```
soulFood5632, line 242: Can you add some docs for this function

PR #289, opened 2025-07-14
281 make mcp clients persist throughout the run

Levivus:
This PR is somewhat of an overhaul of how from_mcp_server works. from_mcp_server no longer returns a list of tools/nodes, but instead a new class, MCPServer. This is the new user facing MCP interaction object. It allows for clients to be persistent, and gives the user control to be able to close the server. To get the tools like we did before, MCPServer.tools returns the same list. 

So, usage will be as follows:
```python
mcp_server = from_mcp_server(config)`  # At this point, the server has started in a background thread.
mcp_tools = mcp_server.tools

# ... Code that uses the tools or whatever ...
mcp_server.close() # This really seems to be unnecessary, only if they need to close it for whatever reason
```

### DIFF
--- a/tests/integration_tests/rc_mcp/test_node_to_mcp.py
+++ b/tests/integration_tests/rc_mcp/test_node_to_mcp.py
@@ -90,11 +90,11 @@ def test_add_nums_tool(mcp_server):
     with rc.Runner(
         executor_config=rc.ExecutorConfig(logging_setting="QUIET", timeout=1000)
     ) as runner:
-        response = asyncio.run(runner.run(server.tools[0], num1=1, num2=3, print_s="Hello"))
+        response = rc.call_sync(server.tools[0], num1=1, num2=3, print_s="Hello")
 
     assert (
-        response.answer[0].text == "14"
-    ), f"Expected 14, got {response.answer[0].text}"
+        response.content[0].text == "14"
+    ), f"Expected 14, got {response.content[0].text}"
 
 
 # --------------------------------------------------------------------------- #


src/requestcompletion/rc_mcp/main.py
```diff
@@ -191,14 +198,17 @@ class MCPToolNode(Node):
...
             self.kwargs = kwargs
-            self.client = MCPAsyncClient(config)
 
         async def invoke(self):
-            async with self.client:
-                result = await self.client.call_tool(tool.name, self.kwargs)
-            if hasattr(result, "content"):
-                return result.content
-            return result
+            print("Invoke!")
```
Aryan-Railtown: Might have left this print haha

tests/unit_tests/rc_mcp/test_main.py
```diff
@@ -99,64 +101,69 @@ async def test_async_client_init_http_uses_correct_transport(
...
-    mock_client.call_tool.return_value = "valonly"
-    result = await node.invoke()
-    assert result == "valonly"
+    mock_result = MagicMock(content="abc")
+    mock_client.call_tool.return_value = mock_result
+
+    # Patch run_coroutine_threadsafe to return a Future with result
+    future = MagicMock()
+    future.result.return_value = mock_result
+    with patch("asyncio.run_coroutine_threadsafe", return_value=future):
```
Aryan-Railtown, line 156: So basically in all of the tests we need to mock the asyncio thread, since we now have a background running mpc server, right?

src/requestcompletion/rc_mcp/main.py
Levivus: yep realized and fixed haha

tests/unit_tests/rc_mcp/test_main.py
Levivus: either that or create a thread, but yes

examples/rc_basics/mcp_example.py
```diff
@@ -3,6 +3,32 @@
 import requestcompletion as rc
 from mcp import StdioServerParameters
 from requestcompletion.nodes.library.mcp_tool import async_from_mcp_server, from_mcp_server
+from requestcompletion.rc_mcp import MCPHttpParams
+
+#%%
+
+# async def main():
```
soulFood5632: supposed to be commented out?

src/requestcompletion/rc_mcp/main.py
```diff
@@ -172,16 +111,80 @@ async def _default_redirect_handler(authorization_url: str) -> None:
...
+    def close(self):
+        """
+        Close the MCP server connection.
+        """
+        self._loop.call_soon_threadsafe(self._shutdown_event.set)
+        self._thread.join()
+
+    @property
+    def tools(self):
+        return self._tools
```
soulFood5632, line 191: quick clean docstring is required here bc it is user facing

src/requestcompletion/rc_mcp/main.py
```diff
@@ -191,14 +194,18 @@ class MCPToolNode(Node):
...
-            self.client = MCPAsyncClient(config)
 
         async def invoke(self):
-            async with self.client:
-                result = await self.client.call_tool(tool.name, self.kwargs)
-            if hasattr(result, "content"):
-                return result.content
-            return result
+            try:
+                future = asyncio.run_coroutine_threadsafe(
```
soulFood5632, line 218: I assume you changed this previous logic becuase it broke things, but I'm worried this is gonna operate as a blocking task. Did you do some checking to see if that was the case?

src/requestcompletion/rc_mcp/main.py
```diff
@@ -172,16 +111,80 @@ async def _default_redirect_handler(authorization_url: str) -> None:
...
+        self._loop.run_until_complete(self.setup())
+        self._ready_event.set()
+
+        loop.run_until_complete(self._shutdown_event.wait())
+
+        self._loop.close()
+
+    async def setup(self):
+        """
+        Set up the MCP server and fetch tools.
```
soulFood5632: Good to specify in this docstring that the function should be run once

src/requestcompletion/rc_mcp/main.py
```diff
@@ -172,16 +111,80 @@ async def _default_redirect_handler(authorization_url: str) -> None:
...
+
+    @property
+    def tools(self):
+        return self._tools
+
+
 def from_mcp(
     tool,
-    config: StdioServerParameters | MCPHttpParams,
+    client,
```
soulFood5632: If we can throw the type here.

src/requestcompletion/rc_mcp/main.py
```diff
@@ -172,16 +111,80 @@ async def _default_redirect_handler(authorization_url: str) -> None:
...
+        self._loop.call_soon_threadsafe(self._shutdown_event.set)
+        self._thread.join()
+
+    @property
+    def tools(self):
+        return self._tools
+
+
 def from_mcp(
     tool,
```
soulFood5632: Same thing here. A type would be nice

src/requestcompletion/rc_mcp/main.py
```diff
@@ -191,14 +194,18 @@ class MCPToolNode(Node):
...
-            async with self.client:
-                result = await self.client.call_tool(tool.name, self.kwargs)
-            if hasattr(result, "content"):
-                return result.content
-            return result
+            try:
+                future = asyncio.run_coroutine_threadsafe(
+                    client.call_tool(tool.name, self.kwargs), loop
+                )
+                result = future.result(timeout=30)
```
soulFood5632: Hardcoded timeout values should probably be avoided. I can see this being a bit of an issue. It would probably be better to allow the MCP client object to pass this is a param. There are other options as well, up to you

src/requestcompletion/rc_mcp/main.py
```diff
@@ -191,14 +194,18 @@ class MCPToolNode(Node):
...
+                return result.content if hasattr(result, "content") else result
```
soulFood5632: when does this come up?

soulFood5632: I know it was in the previous code but I don't really remember why it was there.

examples/rc_basics/mcp_example.py
Levivus: removed now

src/requestcompletion/rc_mcp/main.py
Levivus: When do you think it will be blocking? I have tested it under normal conditions and it works

soulFood5632: Blocking is probably the wrong word. Have you seen what happens when you call 2 MCP tools at the same time? Does it operate them in parallel or do the operate sequentially.?

Levivus: Oh right I get you, I'll look into it

PR #299, opened 2025-07-15
move ruff checks into build instead of push

Levivus:
This is a simple PR that moves the ruff lint checking from push to build that that if the ruff check fails it won't run tests.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,6 +34,7 @@ jobs:
   build:
     name: Build and Test
     runs-on: windows-latest
+    needs: ruff-lint
     steps:
       - name: Print System Architecture
         run: uname -m


.github/workflows/build.yaml
```diff
@@ -9,9 +9,32 @@ on:
       - main
 
 jobs:
+  ruff-lint:
+    name: Check Code with Ruff
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
```
soulFood5632, line 17: These steps are duplicated below

Levivus: They need to be, as each job in a GitHub Actions workflow runs in a separate, clean environment, so steps and files are not shared between jobs unless you explicitly pass artifacts.

soulFood5632: Fantastic check :)

PR #302, opened 2025-07-15
Automatic State Saving to .covailence directory 

soulFood5632:
With this new PR we have the following new behavior:

It will save the data to the following path `.covailence/{run_id}.json`. By default it will create the directory and save it to the current working directory. You can turn off the saving by setting a flag in `ExecutorConfig`. 

One question that still remains is whether we would like this path to be configurable. Right now I do not allow it to be configurable but we could change that. (@jbueza-railtownai, @Amir-R25 what are your thoughts)


.gitignore
```diff
@@ -10,6 +10,7 @@ package-lock.json
 package.json
 test-results.xml
 .coverage
+.covailence
```
jbueza-railtownai, line 13: `covailence` or `covalence`? 😆

soulFood5632: Looks like that is what we are doing for now 😆

src/requestcompletion/run.py
```diff
@@ -106,6 +107,28 @@ def __enter__(self):
...
+                covailence_dir = Path(".covailence")
+                covailence_dir.mkdir(
+                    exist_ok=True
+                )  # Creates if doesn't exist, skips otherwise.
+
+                file_path = (
+                    covailence_dir / f"{self.executor_config.run_identifier}.json"
+                )
+                if file_path.exists():
+                    logger.warning("File %s already exists, overwriting..." % file_path)
```
Amir-R25, line 121: would this ever happen given run_identifiers are unique sha's?

soulFood5632: User can provide them in as a param to `ExecutorConfig` if they want so it could happen.

PR #291, opened 2025-07-14
Tristan/215/shared code in easy usage wrappers

soulFood5632:
-Create and implement a Node Building class that can be used to encourage DRY in our code base. 
-Get rid of abstract class in nodes and replace with __init_subclass__
-Move all validation and checking as low down the inheritance chain as possible
-Restructure structured_tool_call_llm so that the structured_llm class is made on class creation rather than object instantiation
-Changed all system message parameters to be strings. No more acceptance of System Messages as those will be handled internally
-Provided default Pretty name methods to all user facing llm sub classes so pretty name validation has become somewhat obsolete now

PR #307, opened 2025-07-16
Refactor of DynamicFunctionNode and from_function

soulFood5632:
In the following PR I have refactored the `DynamicFunctionNode` used to create `from_function`. And finally I integrated the `from_function` logic into the work completed in #291. This change drastically reducing the amount of boilerplate code in the function and makes it much more workable.

### DIFF
--- a/src/requestcompletion/nodes/library/easy_usage_wrappers/node_builder.py
+++ b/src/requestcompletion/nodes/library/easy_usage_wrappers/node_builder.py
@@ -16,7 +16,6 @@ from typing import (
     overload,
 )
 
-from mcp import StdioServerParameters
 from pydantic import BaseModel
 
 from requestcompletion.exceptions.node_creation.validation import (
@@ -32,6 +31,7 @@ from requestcompletion.nodes.library.mcp_tool import from_mcp_server
 
 from ....llm import MessageHistory, ModelBase, SystemMessage, Tool, UserMessage
 from ....nodes.nodes import Node
+from ....rc_mcp import MCPStdioParams
 from ...library._llm_base import LLMBase
 from ...library.tool_calling_llms._base import OutputLessToolCallLLM
 from ...library.tool_calling_llms.tool_call_llm import ToolCallLLM
@@ -72,6 +72,9 @@ class NodeBuilder(Generic[_TNode]):
         *,
         pretty_name: str | None = None,
         class_name: str | None = None,
+        return_into: str | None = None,
+        format_for_return: Callable[[Any], Any] | None = None,
+        format_for_context: Callable[[Any], Any] | None = None,
     ):
         self._node_class = node_class
         self._name = class_name or f"Dynamic{node_class.__qualname__}"
@@ -79,6 +82,18 @@ class NodeBuilder(Generic[_TNode]):
         if pretty_name is not None:
             self._with_override(
                 "pretty_name", classmethod(lambda cls: pretty_name or cls.__name__)
+            )
+        if return_into is not None:
+            self._with_override("return_into", classmethod(lambda cls: return_into))
+        if format_for_context is not None:
+            self._with_override(
+                "format_for_context",
+                classmethod(lambda cls, x: format_for_context(x)),
+            )
+        if format_for_return is not None:
+            self._with_override(
+                "format_for_return",
+                classmethod(lambda cls, x: format_for_return(x)),
             )
 
     def llm_base(
@@ -167,13 +182,12 @@ class NodeBuilder(Generic[_TNode]):
             f"To perform this operation the node class we are building must be of type LLMBase but got {self._node_class}"
         )
 
-        for elem in connected_nodes:
-            if isfunction(elem):
-                # we have to use lazy imports here to avoid circular imports
-                from ..function import from_function
+        from ..function import from_function
 
-                connected_nodes.remove(elem)
-                connected_nodes.add(from_function(elem))
+        connected_nodes = {
+            from_function(elem) if isfunction(elem) else elem
+            for elem in connected_nodes
+        }
 
         if not isinstance(connected_nodes, set):
             connected_nodes = set(connected_nodes)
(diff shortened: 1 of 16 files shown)


src/requestcompletion/nodes/library/easy_usage_wrappers/node_builder.py
```diff
@@ -108,6 +106,7 @@ def llm_base(
             f"To perform this operation the node class we are building must be of type LLMBase but got {self._node_class}"
         )
         if llm_model is not None:
+            # TODO fix whatever this is.
```
Tristan-railtown, line 124: In testing a method was being passed for model which was causing this to fail. Might have been missing something bigger but that's why I added it in

soulFood5632: Weird, I'll take a look at it.

src/requestcompletion/nodes/library/easy_usage_wrappers/structured_tool_call_llm.py
```diff
@@ -61,12 +61,10 @@ def structured_tool_call_llm(  # noqa: C901
...
         class_name="EasyStructuredToolCallLLM",
-        tool_details=tool_details,
-        tool_params=tool_params,
     )
     builder.llm_base(llm_model, system_message)
     builder.tool_calling_llm(connected_nodes, max_tool_calls)
-    if tool_details is not None:
+    if tool_details is not None or tool_params is not None:
         builder.tool_callable_llm(tool_details, tool_params)
     builder.structured(schema)
```
Tristan-railtown, line 83: I believe we throw an exception if tool_details is None but tool_params is not so potentially redundant.

src/requestcompletion/nodes/library/easy_usage_wrappers/terminal_llm.py
```diff
@@ -45,11 +45,9 @@ def terminal_llm(  # noqa: C901
         TerminalLLM,
         pretty_name=pretty_name,
         class_name="EasyTerminalLLM",
-        tool_details=tool_details,
-        tool_params=tool_params,
     )
     builder.llm_base(llm_model, system_message)
-    if tool_details is not None:
+    if tool_details is not None or tool_params is not None:
         builder.tool_callable_llm(tool_details, tool_params)
```
Tristan-railtown, line 66: Same as above

src/requestcompletion/llm/type_mapping.py
```diff
@@ -0,0 +1,155 @@
...
+from requestcompletion.llm.tools.parameter_handlers import UnsupportedParameterError
+
+
+class TypeMapper:
+    """
+    A simple type that allow you to attach a function which will signature will be stored.
+
+    By calling convert_kwargs_to_appropriate_types with a dictionary of kwargs, it will convert the values into the
+    expected value types based on the function signature.
+    """
```
Tristan-railtown: I found this to be a bit confusing. If user facing for sure fix first line

src/requestcompletion/nodes/library/easy_usage_wrappers/structured_tool_call_llm.py
soulFood5632: The logic that threw that exception was never getting triggered after I moved it out of the `__init__` in `NodeBuilder` and into this method. Basically without this addition the exception was never thrown because we never entered the validation function.

soulFood5632: And btw I moved it out the `__init__` method because not all nodes would want to have that logic.

src/requestcompletion/nodes/library/easy_usage_wrappers/terminal_llm.py
soulFood5632: See above

src/requestcompletion/nodes/library/easy_usage_wrappers/node_builder.py
```diff
@@ -256,30 +304,83 @@ def tool_callable_llm(
...
         """
 
-        def tool_info(cls: Type[_TNode]) -> Tool:
-            return Tool(
-                name=cls.pretty_name().replace(" ", "_"),
-                detail=tool_details,
-                parameters=tool_params,
+        if tool is not None:
+            assert name is None and tool_details == "" and tool_params is None, (
+                "If you pass a Tool object to override_tool_info, you cannot pass name, tool_details or tool_params."
```
Tristan-railtown, line 368: Potentially we could give a warning and ignore the arguments they passed the same way we do with llm_model or system_message?

src/requestcompletion/nodes/library/easy_usage_wrappers/structured_llm.py
```diff
@@ -53,12 +53,10 @@ def structured_llm(  # noqa: C901
...
         pretty_name=pretty_name,
         class_name="EasyStructuredLLM",
-        tool_details=tool_details,
-        tool_params=tool_params,
     )
     builder.llm_base(llm_model, system_message)
     builder.structured(schema)
-    if tool_details is not None:
+    if tool_details is not None or tool_params is not None:
         builder.tool_callable_llm(tool_details, tool_params)
```
Tristan-railtown, line 74: Comment below about redundancy

src/requestcompletion/nodes/library/easy_usage_wrappers/tool_call_llm.py
```diff
@@ -56,8 +56,6 @@ def tool_call_llm(  # noqa: C901
         ToolCallLLM,
         pretty_name=pretty_name,
         class_name="EasyToolCallLLM",
-        tool_details=tool_details,
-        tool_params=tool_params,
     )
     builder.llm_base(llm_model, system_message)
     builder.tool_calling_llm(connected_nodes, max_tool_calls)
```
Tristan-railtown: Looks like for consistency you would want to add the "or if tool_params not none" statement if you choose to leave that in?

src/requestcompletion/nodes/library/easy_usage_wrappers/node_builder.py
soulFood5632: This logic was specifically designed to uphold the invariants of the typing schema seen in the above `@overload` methods. It will never be reached as long as the developers follows typing rules so it makes sense to be an assertion

src/requestcompletion/nodes/library/easy_usage_wrappers/tool_call_llm.py
soulFood5632: Good catch.

PR #334, opened 2025-07-21
203 bug easy usage wrapper typing issue

soulFood5632:
There was a lot of typing shenanigans gone wrong in our Node logic. This PR implements a tighter use of generics to allow the typing to actually work properly so you are aware when something has gone wrong. 

See in the below image this new typing in work. The biggest update is not the returns from `call_sync` are the correct inferred type.  

image

### DIFF
--- a/tests/integration_tests/library/test_function.py
+++ b/tests/integration_tests/library/test_function.py
@@ -41,7 +41,7 @@ class TestPrimitiveInputTypes:
             secret_phrase,
             model_provider=model_provider,
         )
-        with rc.Runner(rc.ExecutorConfig(logging_setting="VERBOSE")) as run:
+        with rc.Runner(rc.ExecutorConfig(logging_setting="NONE")) as run:
             response = run.run_sync(
                 agent,
                 rc.llm.MessageHistory(


src/requestcompletion/nodes/_node_builder.py
```diff
@@ -149,9 +153,6 @@ def structured(
 
         self._with_override("schema", classmethod(lambda cls: schema))
 
-    def struct_mess_hist(self):
-        self._with_override("struct_mess_hist", True)
-
```
Tristan-railtown, line 154: Is this rewritten somewhere else? This method is important for structured_message_history_tool_call_llm easy_usage_wrapper. Maybe could use a better name though.

soulFood5632: Yep, this detail gets not needed since we handle this override with class inheritance instead

tests/integration_tests/library/test_function.py
```diff
@@ -41,7 +41,7 @@ def secret_phrase() -> str:
             secret_phrase,
             model_provider=model_provider,
         )
-        with rc.Runner(rc.ExecutorConfig(logging_setting="NONE")) as run:
+        with rc.Runner(rc.ExecutorConfig(logging_setting="VERBOSE")) as run:
```
Tristan-railtown: Just highlighting in case you made Verbose for debugging and forgot to change back.

soulFood5632: Yep forgot that

PR #337, opened 2025-07-21
Create add_attribute method for _node_builder.py

Tristan-railtown:
Added method that can add methods or fields to node classes being made by NodeBuilder.

This can take methods or values and then make them class methods or fields.

## Checklist for Author

### DIFF
--- a/src/requestcompletion/nodes/_node_builder.py
+++ b/src/requestcompletion/nodes/_node_builder.py
@@ -383,7 +383,7 @@ class NodeBuilder(Generic[_TNode]):
 
     def add_attribute(self, name: str, attribute, make_function: bool, *args, **kwargs):
         """
-        Add or override an attribute or method on the dynamically built node class. 
+        Add or override an attribute or method on the dynamically built node class.
         This takes functions or values and can make them class methods or class fields
 
         Args:


src/requestcompletion/nodes/_node_builder.py
```diff
@@ -381,6 +381,34 @@ def prepare_tool(cls, tool_parameters: Dict[str, Any]):
 
         self._with_override("prepare_tool", classmethod(prepare_tool))
 
+    def add_attribute(self, name: str, attribute, make_function: bool, *args, **kwargs):
```
soulFood5632, line 384: Sorry I lied, I am actually a bit confused by why you handle the attribute differently if its a callable?

Tristan-railtown: So that's why it might be a bit over-engineered but basically it gives the option to either use the function immediately and store the returned value as a field or to store the function itself and then the class can call it at some later time as a function with parameters rather than just a value/field.

Tristan-railtown: So to further clarify, you have 4 cases
pass a value, it can then become a class value or class method that returns that value,
pass a function, it can then be called and the return is a class value, or it is stored as a class method

soulFood5632: Yes I follow. This makes sense.

PR #367, opened 2025-07-23
Prompt docs

Levivus:
Add prompt documentation page

### DIFF
--- a/docs/llm_support/prompts.md
+++ b/docs/llm_support/prompts.md
@@ -109,3 +109,6 @@ system_msg = Message(role="system", content="You are a {role}.", inject_prompt=T
 # This message will not have context injection applied
 user_msg = Message(role="user", content="Tell me about {topic}.", inject_prompt=False)
 ```
+
+This can be useful when you want to control which messages should have context injected and which should not. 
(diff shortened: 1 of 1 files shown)


docs/llm_support/prompts.md
```diff
@@ -0,0 +1,111 @@
...
+
+1. **Check context values**: Ensure the context contains the expected values for your placeholders
+2. **Verify prompt injection is enabled**: Check that `prompt_injection=True` in your `ExecutorConfig`
+3. **Look for syntax errors**: Ensure your placeholders use the correct format `{variable_name}`
+
+## Advanced Usage
+
+### Message-Level Control
+
+Context injection can be controlled at the message level using the `inject_prompt` parameter:
```
soulFood5632, line 101: You should mention as a note that this largely uneeded becuase it is abstracted away for most uses

Levivus: I guess I was thinking they could have it generally enabled, but one prompt might have like math so needs to not get messed up by { or something. So for some uses it might be needed
Though I could write something along those lines, but I figured the fact it's in advanced usage means most people wouldn't think it's necessary

soulFood5632: Yea that makes sense. I would just say provide that as a specific example so they understand when it could be useful.

PR #414, opened 2025-08-05
399 feature context interaction added features

Levivus:
Making context a dict/importing it and using the object directly instead of using imported functions seemed like it would require more changes, so this is a quick pr that just adds update and delete methods to context.

### DIFF
--- a/packages/railtracks/src/railtracks/context/central.py
+++ b/packages/railtracks/src/railtracks/context/central.py
@@ -296,7 +296,7 @@ def put(
 
 def update(data: dict[str, Any]):
     """
-    Update the context with the provided data.
+    Sets the values in the context. If the context already has values, this will overwrite them, but it will not delete any existing keys.
 
     Args:
         data (dict[str, Any]): The data to update the context with.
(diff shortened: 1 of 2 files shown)


packages/railtracks/src/railtracks/context/__init__.py
```diff
@@ -1,9 +1,8 @@
...
     get,
     put,
+    update,
 )
 
-__all__ = [
-    "put",
-    "get",
-]
+__all__ = ["put", "get", "update", "delete"]
```
soulFood5632, line 8: [pedantic] but can me make this vertical list

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -284,15 +284,41 @@ def put(
...
+        key (str): The key to set.
+        value (Any): The value to set.
     """
     context = safe_get_runner_context()
     context.external_context.put(key, value)
 
 
+def update(data: dict[str, Any]):
+    """
+    Update the context with the provided data.
```
soulFood5632: better docs please

packages/railtracks/src/railtracks/context/__init__.py
Levivus: ruff wants this

PR #424, opened 2025-08-06
Send session id as part of Payload + allow multiple runs. 

soulFood5632:
Send `session_id` as part of the output payload. 

Critically: 
- Added a new method to session object `payload()` to create a payload of intrest
- Modified the logic in the info object so the outputted json includes both the session_id and properly traversed the tree for all potential insertion requests

### DIFF
--- a/packages/railtracks/src/railtracks/state/info.py
+++ b/packages/railtracks/src/railtracks/state/info.py
@@ -35,7 +35,7 @@ class ExecutionInfo:
         self.stamper = stamper
 
     @classmethod
-    def default(cls, session_id: str) -> ExecutionInfo:
+    def default(cls) -> ExecutionInfo:
         """Creates a new "empty" instance of the ExecutionInfo class with the default values."""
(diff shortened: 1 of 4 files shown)


packages/railtracks/src/railtracks/state/info.py
```diff
@@ -133,58 +156,40 @@ def graph_serialization(self) -> str:
...
-                "steps": self.all_stamps,
-            },
+                "session_id": session_id,
+                "name": info.name,
+                "run_id": str(uuid.uuid4()),
+                "nodes": info.node_forest.to_vertices(),
+                "edges": info.request_forest.to_edges(),
+                "steps": _get_stamps_from_forests(
+                    info.node_forest, info.request_forest
+                ),
```
jbueza-railtownai, line 175: cc @mbrailtown @tlou006railtown for visibility, we will be receiving session_id and run_id in the agent runs soon!

PR #467, opened 2025-08-08
122 Failed type conversion returns default

soulFood5632:
## What does this add?

Previously if the a type conversion failed it would return a simple string that caused further issues. Instead we should just return the initial type and hope for the best. 

## Type of changes

### Code Quality

### Testing

### Documentation

### Git & PR Management

### Breaking Changes

---

### DIFF
--- a/packages/railtracks/tests/integration_tests/library/test_function.py
+++ b/packages/railtracks/tests/integration_tests/library/test_function.py
@@ -147,8 +147,9 @@ class TestPrimitiveInputTypes:
                     ]
                 ),
             )
+        resp: str = response.content
 
-        assert response.content == "True"
+        assert resp.lower() == "true"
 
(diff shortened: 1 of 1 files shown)


packages/railtracks/src/railtracks/llm/type_mapping.py
```diff
@@ -80,7 +79,7 @@ def _convert_value(
             if inspect.isclass(target_type) and not hasattr(target_type, "__origin__"):
                 return target_type(value)
         except (TypeError, ValueError):
-            return "Tool call parameter type conversion failed."
+            return value
```
Aryan-Railtown, line 82: we can just `pass` in the except block case. Since the function already defaults to returning the value

soulFood5632: I'd rather be explicit about things here. Its a pedantic decision either way

PR #538, opened 2025-08-13
Async Await Docs

soulFood5632:
## What does this add?

A simple update to add a page outlining async await docs

## Type of changes

### DIFF
--- a/docs/tutorials/guides/async_await.md
+++ b/docs/tutorials/guides/async_await.md
@@ -91,10 +91,10 @@ async def main():
 
 ## 🏗 How RT uses `async/await`
 
-In RT if you want to call another node you should use the `async` logic.
+If you are writing a tool in RT and need to call another tool, use `rt.call(...)`. This ensures the RT backend tracks the tool invocation, enabling logging and visualization.
+
 !!! Note
-    If you are making a node which uses `rt.call(...)` to call another node you must use the `async` keyword in the
-    function definition.
+    Because rt.call(...) returns a coroutine, your function must be declared as async (e.g., `async def my_function(...)`) so you can `await rt.call(...)`.
 
 ```python
 import railtracks as rt
@@ -109,8 +109,8 @@ def split_text(text: str) -> list[str]:
 def alternate_capitalization(text: str) -> str:
     return text.swapcase()
 
-# since the modify_text function calls other nodes, it must be an async function
-# you can use the asyncio library to run the nodes however you want. See below for an example of sequential and parallel execution.
+# Since the modify_text function calls other nodes, it must be an async function
+# You can use the asyncio library to run the nodes however you want.
 @rt.function_node
 async def modify_text(text: str) -> str:
     # Call the split_text node sequentially.


docs/tutorials/guides/async_await.md
```diff
@@ -0,0 +1,124 @@
...
+```python
+async def my_async_function():
+    await another_async_function()
+```
+
+!!! Warning
+    `await` can only be used inside an `async` function.
+
+!!! Tip
+    If you are curious about how Python handles the event loop I would encourage you to check out the [official documentation](https://docs.python.org/3/library/asyncio-eventloop.html).
```
Tristan-railtown: Maybe wouldn't add it here but this feels like a rushed higher-level explanation of async. Makes sense since this is documentation for our own confusing library but maybe nice to link asyc doc as well if they're really new.
https://docs.python.org/3/library/asyncio.html

docs/tutorials/guides/async_await.md
```diff
@@ -0,0 +1,124 @@
...
+
+!!! note "Sequential Execution"
+    Each coroutine will run one after the other, waiting for each to finish before starting the next
+    ```python
+        await coroutine_1()
+        await coroutine_2()
+    ```
+
+## 🏗 How RT uses `async/await`
+
```
Tristan-railtown, line 93: I'm not sure if there is a table of contents or something to indicate to people who know async well to skip to this part but probs something we would want to make sure people who are experienced with async don't skip because this is sort of at the bottom

Tristan-railtown: Yea ok it's in the table of contents so maybe forget about this comment

docs/tutorials/guides/async_await.md
```diff
@@ -0,0 +1,124 @@
...
+In RT if you want to call another node you should use the `async` logic.
+!!! Note
+    If you are making a node which uses `rt.call(...)` to call another node you must use the `async` keyword in the
+    function definition.
```
Tristan-railtown: Maybe draw extra attention to the fact that it must be wrapped in an async function. The wording sort make it seem like you only add async if you're using it in a function. Unless I'm misunderstanding how it works

docs/tutorials/guides/async_await.md
soulFood5632: Yea it does feel a bit out of the blue. It's probably better to keep it high level at the begining.

docs/tutorials/guides/async_await.md
soulFood5632: Either way I think I can understand the core of this comment. We need some flow for people who know their away around async to skim how it applies to RT.

soulFood5632: @Amir-R25 what do you think about this

docs/tutorials/guides/async_await.md
soulFood5632: Yea the wording is kinda bad, let me rework it a bit.

PR #558, opened 2025-08-18
updated serialization to allow for name and node_type

soulFood5632:
## What does this add?

Previously we only provided the `pretty_name`, and we did not provide any semblance of node type. Now we have updated the schema to include 2 new fields. 

`name`: (the previous pretty name)
`node_type`: Literal["Tool", "Agent", "Other"]

This means a very simple change to serialization structure to make this happen.  

## Type of changes

## Background context

There was a need for this change to unlock better visualizations, to make it easier to undertsand the difference between types of nodes and make the rendering as easy as possible. 

### Code Quality

### Testing

### Documentation

### Git & PR Management

### Breaking Changes

### DIFF
--- a/packages/railtracks/tests/unit_tests/state/test_node.py
+++ b/packages/railtracks/tests/unit_tests/state/test_node.py
@@ -29,10 +29,10 @@ def test_linkednode_to_vertex_includes_hierarchy(dummy_node_factory, linked_node
 
     vertex = child_ln.to_vertex()
     assert vertex.identifier == child_ln.identifier
-    assert vertex.node_type == child_ln.node.name()
+    assert vertex.name == child_ln.node.name()
     assert vertex.stamp == child_ln.stamp
     assert vertex.parent.identifier == parent_ln.identifier
-    assert vertex.type == parent_ln.type()
+    assert vertex.node_type == parent_ln.node.type()
 
 def test_linkednode_nodecopyerror_raised_on_safe_copy(monkeypatch, linked_node_factory):
     class UnsafeNode:


packages/railtracks/src/railtracks/nodes/concrete/_llm_base.py
```diff
@@ -310,6 +310,9 @@ def safe_copy(self) -> Self:
 
         return new_instance
 
+    def type(self):
```
Amir-R25: do we want to put a `@property` decorator before this for ease of use?

packages/railtracks/src/railtracks/nodes/concrete/chat_tool_call_llm.py
```diff
@@ -138,3 +138,6 @@ async def invoke(self):  # noqa: C901
                 )
 
         return self.return_output()
+
+    def type(self):
```
Amir-R25: same here

packages/railtracks/src/railtracks/nodes/concrete/function_base.py
```diff
@@ -67,6 +67,9 @@ def prepare_tool(cls, tool_parameters: Dict[str, Any]) -> Self:
         )
         return cls(**converted_params)
 
+    def type(self):
```
Amir-R25: and here :)

packages/railtracks/src/railtracks/state/node.py
```diff
@@ -29,7 +29,8 @@ class LinkedNode(AbstractLinkedObject):
     def to_vertex(self):
         return Vertex(
             identifier=self.identifier,
-            node_type=self.node.name(),
+            node_type=self.node.type(),
```
Amir-R25, line 32: if you do this to do it, the decorator will change here. I also suggest we do the same for name? But not sure how that would work with classmethods

soulFood5632: You can't do them :(. But I did mean to make type a class method 😆

PR #631, opened 2025-08-29
added latency param

soulFood5632:
## What does this add?

Previously we did not include the latency of each individual LLM call in the json payload did not include the latency parameter. This was becuase although we stored it in the `RequestDetails` object we did not properly serilaize it. 

This change modifies the json schema and adds the latency parameter to be serialized. 

## Type of changes

### DIFF
--- a/docs/quickstart/quickstart.md
+++ b/docs/quickstart/quickstart.md
@@ -1,4 +1,6 @@
-Build your first agentic system in just a few steps.
+LLMs are powerful, but they’re not perfect. They’ll trip over simple reasoning tasks that humans can do without thinking. Railtracks is an agentic framework that makes it easier build more intelligent systems.
+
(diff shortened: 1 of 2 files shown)


packages/railtracks/src/railtracks/state/serialize.py
```diff
@@ -134,6 +134,7 @@ def encode_request_details(details: RequestDetails) -> dict[str, Any]:
         "output_tokens": details.output_tokens,
         "total_cost": details.total_cost,
         "system_fingerprint": details.system_fingerprint,
+        "latency": details.latency,
```
jbueza-railtownai, line 137: there's another property outside of llm details which is like latency.total_time -- we keep that since that's the aggregate?

image

soulFood5632: Yep, that is the total latency of the node. These properties are a per LLM call.

PR #652, opened 2025-09-04
Updating `@rt.session` decorator to work without parentheses when using default values + other stability upgrades + typing 

### LINKED ISSUE #629: [General] Proposal: Completely Rethink Session
### Feedback or Inquiry

We are running into various big pains when working with session. I propose a change the interface. 

Currently we do the following:
```python 
@rt.session()
async def some_function(...):
  pass

# or 
async def some_function(...):
   with rt.Session():
      pass
```

This works fine but the top case runs into problems when you want to work with the `ExecutorInfo` object to collect and send the `agent_runs` to conductr or some outside service. 

This issue should be a discussion focused approach. Ideas are welcome in the comments. 

----
# Key Requirements
1. Simple interface to run agents in a lightweight manner. 
2. Ease of use to send state runs after completion. 
3. Ease of use setting configuration params for a session container

For each idea please provide an example of how to do all the above. 

-----

## An option
Instead what if we softly removed the session interface and replace it with something similar but a bit more lightweight. 
1. Running agents 
```python 
async def some_function(...):
  await rt.asession(...)
```
or 
```python 
def some_function(...):
   rt.session(...)
```
Upon completion the method would return an info object containing that matches our current system. The astute few would remember the old method of `run` that was deleted. 

2. Sending runs
```python 
async def some_function(...):
  result  = await rt.asession(...)
  railtownai.send_agent_logs(result.info.serilized())
```
3. Setting conifgrations
- possible but would break typing

### Cons
This works for everything except the typing of the inputs would be impossible since we'd have to use param spec for the inputs which would make it hard to set the config parameters. 

### Additional Context

_No response_

Amir-R25:
## What does this add?

Enhanced the session decorator with proper overloads to support both `@session` and `@session(...)` syntax, and modified it to return a tuple of (result, session) for better access to session state and debugging information.

Improved the Session class to handle invalid characters in session names by implementing a fallback mechanism that uses only the session identifier when the preferred filename fails. Added comprehensive error handling with proper warning messages using the centralized exception message system.

closes #629 

## Type of changes

- [x] 🎨 Code style/formatting (changes that do not affect the meaning of the code)

## Background context

Typing was broken when using `@session` decorator without the parantheses.

### Code Quality

### Testing

### Documentation

### Git & PR Management

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,4 +1,4 @@
-name: Publish to Azure Artifacts on Release
+name: Publish Packages
 
 on:
   release:
@@ -41,11 +41,21 @@ jobs:
           sed -i -E 's/__version__ = ".+"/__version__ = "'"$VERSION_CLEANED"'"/' packages/railtracks/src/railtracks/__init__.py
           echo "Updated railtracks __init__.py to version $VERSION_CLEANED"
 
-      - name: Build and Publish Railtracks Core
+      - name: Build and Publish Railtracks Core to Azure Artifacts
         env:
           FLIT_INDEX_URL: 'https://pkgs.dev.azure.com/railtownai/ML/_packaging/railtownai_rc/pypi/upload/'
           FLIT_USERNAME: 'AZURE'
           FLIT_PASSWORD: ${{ secrets.AZURE_ARTIFACTS_PAT }}
+        run: |
+          cd packages/railtracks
+          flit publish
+
+      - name: Build and Publish Railtracks Core to PyPI
+        if: false
+        env:
+          FLIT_INDEX_URL: 'https://pypi.org/simple'
+          FLIT_USERNAME: '__token__'
+          FLIT_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
         run: |
           cd packages/railtracks
           flit publish
@@ -85,6 +95,16 @@ jobs:
           FLIT_INDEX_URL: 'https://pkgs.dev.azure.com/railtownai/ML/_packaging/railtownai_rc/pypi/upload/'
           FLIT_USERNAME: 'AZURE'
           FLIT_PASSWORD: ${{ secrets.AZURE_ARTIFACTS_PAT }}
+        run: |
+          cd packages/railtracks-cli
+          flit publish
+
+      - name: Build and Publish Railtracks CLI to PyPI
+        if: false
+        env:
+          FLIT_INDEX_URL: 'https://pypi.org/simple'
+          FLIT_USERNAME: '__token__'
+          FLIT_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
         run: |
           cd packages/railtracks-cli
           flit publish
--- a/docs/tools_mcp/mcp/MCP_tools_in_RT.md
+++ b/docs/tools_mcp/mcp/MCP_tools_in_RT.md
@@ -122,6 +122,16 @@ github_server = connect_mcp(
 github_tools = github_server.tools
 ```
 
+
+!!! Warning
+    You must set your github token in your .env file or you will see an error like:
+    
+    Exception in thread Thread-1 (_thread_main):
+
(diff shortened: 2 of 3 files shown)


packages/railtracks/src/railtracks/_session.py
```diff
@@ -249,9 +262,24 @@ def payload(self) -> Dict[str, Any]:
...
-    context: Dict[str, Any] | None = None,
+    func: Callable[_P, Coroutine[Any, Any, _TOutput]],
+) -> Callable[_P, Coroutine[Any, Any, Tuple[_TOutput, Session]]]:
+    """
+    Decorator for async functions without configuration (used as @session).
+
+    This automatically creates and manages a Session context with default settings.
+    The decorated function returns a tuple of (result, session).
+    """
+    ...
```
jbueza-railtownai, line 274: 🤔 what's this `...` ? TIL 😆

Amir-R25: Ellipsis, basically a placeholder. It's how python handles overloaded function definitions. You put placeholder for all implementations except the "main" one.

docs/scripts/session.py
```diff
@@ -15,8 +15,9 @@ async def empty_workflow():
 
 
 # Run your workflow
-result = asyncio.run(empty_workflow())
+result, session = asyncio.run(empty_workflow())
 print(result)  # "Hello, Alice!"
+print(f"Session ID: {session._identifier}")
```
soulFood5632, line 20: marking this an issue to expose _identifier as a public property.

docs/scripts/session.py
```diff
@@ -58,8 +60,10 @@ async def second_workflow():
 
 
 # Run independently
-print(asyncio.run(first_workflow()))  # "Hello, Diana!"
-print(asyncio.run(second_workflow()))  # "Bye, Robert!"
+result1, session1 = asyncio.run(first_workflow())
```
soulFood5632, line 63: This feels slightly pedandic but it might be nice to show that you can go `result1, _ = asyncio.run(...)` if you don't need the session

packages/railtracks/src/railtracks/_session.py
```diff
@@ -50,7 +55,7 @@ class Session:
     - `logging_setting`: "REGULAR"
     - `log_file`: None (logs will not be written to a file)
     - `broadcast_callback`: None (no callback for broadcast messages)
-    - `run_identifier`: None (a random identifier will be generated)
+    - `name`: None
```
soulFood5632: 😄 thanks

packages/railtracks/src/railtracks/_session.py
```diff
@@ -179,10 +176,26 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                        get_message(
+                            ExceptionMessageKey.INVALID_SESSION_FILE_NAME_WARN
+                        ).format(name=self.name, identifier=self._identifier)
+                    )
+                    file_path = railtracks_dir / f"{self._identifier}.json"
+
                 if file_path.exists():
                     logger.warning("File %s already exists, overwriting..." % file_path)
 
+                # file_path.touch()
```
soulFood5632: loose commented out line

packages/railtracks/src/railtracks/_session.py
soulFood5632: @jbueza-railtownai you can also do `pass` does the same thing.

packages/railtracks/src/railtracks/_session.py
```diff
@@ -262,14 +290,17 @@ def session(
     name: str | None = None,
```
soulFood5632: marking this as a ticketb to move name up higher in the signature.

Amir-R25: I'll do it in this PR, to top?

soulFood5632: Yea sure. that works.

Issue #659, opened 2025-09-05
[Bug] `allowable_log_levels` causes typing problems


Amir-R25:
### Description

in `logging/config.py` we define `allowable_log_levels = Literal["VERBOSE", "REGULAR", "QUIET", "NONE"]`

following that, in `utils/config.py` we define `ExecutorCOnfig` as:
```python
class ExecutorConfig:
    def __init__(
        self,
        *,
        timeout: float = 150.0,
        end_on_error: bool = False,
        logging_setting: allowable_log_levels = "REGULAR",
        log_file: str | os.PathLike | None = None,
        broadcast_callback: (
            Callable[[str], None] | Callable[[str], Coroutine[None, None, None]] | None
        ) = None,
        prompt_injection: bool = True,
        save_state: bool = True,
    ):
```

This causes static type checking to complain with the following in `_session.py`:
`Argument of type "str" cannot be assigned to parameter "setting" of type "allowable_log_levels" in function "prepare_logger"`

while this can be fixed by changing line 37 of `utils/config.py` to:
`        self.logging_setting: allowable_log_levels = logging_setting`

let's think on this ticket if we can follow best practices for typehinting (ie naming a type and etc)

### Reproducing Steps (include code examples if applicable)

N/A

PR #686, opened 2025-09-09
[BUG] fix bug that didnt check typing for logging_setting that can tak…

somtirtharoy:
…e from a set of Literal values

## What does this add?
- This adds code to validate the usage of AllowableLogLevels in ExecutorConfig so that it validates during run time.
- It also adds code to remove the static type checking error
- Adds camel style naming for  AllowableLogLevels  and makes those changes everywhere in the repo.

Issue: https://github.com/RailtownAI/railtracks/issues/659 

## Type of changes

- [X ] 🐛 Bug fix (non-breaking change which fixes an issue)
### Code Quality
- [ ] Code follows the project's style guidelines (run `ruff check .` and `ruff format .`)
- [ ] Code is commented, particularly in hard-to-understand areas

### Testing
- [ ] Test coverage maintained

### Documentation

### Git & PR Management

### Breaking Changes

---

### DIFF
--- a/packages/railtracks/src/railtracks/utils/config.py
+++ b/packages/railtracks/src/railtracks/utils/config.py
@@ -42,7 +42,7 @@ class ExecutorConfig:
 
     @property
     def logging_setting(self) -> AllowableLogLevels:
-        return self._logging_setting 
+        return self._logging_setting
 
     @logging_setting.setter
     def logging_setting(self, value: AllowableLogLevels):
@@ -50,7 +50,7 @@ class ExecutorConfig:
             raise ValueError(
                 f"logging_setting must be one of {allowable_log_levels_set}, got {value}"
             )
-        self._logging_setting : AllowableLogLevels = value
+        self._logging_setting: AllowableLogLevels = value
 
     def precedence_overwritten(
         self,


packages/railtracks/src/railtracks/utils/config.py
```diff
@@ -40,12 +40,24 @@ def __init__(
         self.prompt_injection = prompt_injection
         self.save_state = save_state
 
+    @property
+    def logging_setting(self) -> AllowableLogLevels:
+        return self._logging_setting  # type: ignore
```
Amir-R25: I think we ideally want to fix this typing problem

Amir-R25: if you add the line:

```python
        self._logging_setting: AllowableLogLevels
```

inside the `__init__` this should be resolved, let me know what you think

packages/railtracks/tests/unit_tests/utils/test_config.py
```diff
@@ -6,7 +6,7 @@
 from railtracks.utils.config import ExecutorConfig
```
Amir-R25, line 7: can you add "VERBOSE" instead of "DEBUG" for these tests please?

somtirtharoy: @Amir-R25 I get what you are saying. I think putting it as in the setter like so:
`self._logging_setting : AllowableLogLevels = value`

solves the issue of the static type checking. Putting it in the __init__ method defeats the purpose of having it as a private variable accessible through the getter and setter methods. I made the changes and pushed.  let me know

Also added 'VERBOSE' to the tests.

PR #701, opened 2025-09-12
created codeowners file and added dev dependency experince

soulFood5632:
## What does this add?
Now me or Amir has to review any changes

also accidently in this change are the changes to a uv style approach to dev dependencies. The required changes have been made to the docs.


pyproject.toml
```diff
@@ -1,6 +1,48 @@
-[build-system]
-requires = ["flit_core >=3.2,<4"]
-build-backend = "flit_core.buildapi"
```
jbueza-railtownai, line 3: will this break the flit build && flit publish release?

soulFood5632: It should not because this the root pyproject, the release operations run on the on the pyprojects within the `packages` folder

PR #693, opened 2025-09-11
[Feature] Support for Cohere API

somtirtharoy:
## Type of changes

Please check the type of change your PR introduces:

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update (improvements or corrections to documentation)
- [ ] 🎨 Code style/formatting (changes that do not affect the meaning of the code)
- [ ] ♻️ Refactoring (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ Performance improvement (code change that improves performance)
- [ ] ✅ Test update (adding missing tests or correcting existing tests)
- [ ] 🔧 Build/CI changes (changes to build process or continuous integration)
- [ ] 🗑️ Chore (other changes that don't modify src or test files)

### Code Quality
- [ ] Code follows the project's style guidelines (run `ruff check .` and `ruff format .`)
- [ ] Code is commented, particularly in hard-to-understand areas

### Testing
- [ ] Tests added/updated and pass locally (`pytest tests`)
- [ ] Test coverage maintained

### Documentation
- [ ] Documentation updated if needed (bot will verify)

### Git & PR Management
- [ ] PR title clearly describes the change

### Breaking Changes
- [ ] Breaking changes are documented
- [ ] Migration guide provided in documentation (step-by-step instructions for users to update their code/config)

---

Issue #541, opened 2025-08-13
[Feature] Support for Cohere API


Amir-R25:
### Feature Description

We'd like to have support for connecting to Cohere API directly. 

Note: Currently we can access Cohere through Azure and HuggingFace but we'd like to support API as well.

### Proposed Solution

extend the LLM class as we've done to support Cohere's API.

Checkout `packages/railtracks/llm/models/api_providers/openai` for example of how to do it. Basically we will be copying the logic to work with Cohere.

soulFood5632:
This issues is farily stale since cohere has gone out of style

PR #713, opened 2025-09-16
Write an enum type for an the llm providers

Tristan-railtown:
## What does this add?
This creates a providers file that has an enumeration for the different providers we support. This now makes it so that when model_type is called it return the enumeration rather than a string.

One important note is that this enumeration inherits str so is_instance(str, model_provider) will return true and same with something like model_provider.OPENAI == "OpenAI".

## Type of changes

### Code Quality

### Testing

### Git & PR Management

## Additional Notes

I chose to inherit enum and str as well for convenience but potentially it would be preferred to just inherit from enum.


soulFood5632 (approved):
Great work, lets hold off on merging since this is blocked by #693. But after that we are good.

Amir-R25 (approved):
Awesome!

soulFood5632:
@Tristan-railtown check out #541 for one last update to complete.

packages/railtracks/src/railtracks/llm/models/providers.py
```diff
@@ -28,3 +29,4 @@ class ModelProvider(str, Enum):
     HUGGINGFACE = "HuggingFace"
     AZUREAI = "AzureAI"
     OLLAMA = "Ollama"
+    COHERE = "cohere_chat"
```
soulFood5632, line 32: Lets call it "Cohere"

packages/railtracks/src/railtracks/llm/models/api_providers/cohere.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class CohereLLM(ProviderLLMWrapper):
+    """
+    A wrapper that provides access to the Cohere API.
+    """
+
+    @classmethod
+    def model_type(cls):
+        return ModelProvider.HUGGINGFACE
```
soulFood5632: called the wrong thing here it looks

soulFood5632 (changes requested):
Minor changed

packages/railtracks/src/railtracks/llm/models/api_providers/cohere.py
Tristan-railtown: bruh... my bad

packages/railtracks/src/railtracks/llm/models/providers.py
Tristan-railtown: Looking into this right now but is the string not an API dependent thing? If not, should we change "Vertex_ai" to "Gemini" as well?

soulFood5632: I'll leave this a bit to discussion. I like the cleaness of `Cohere`, but I think it is better to describe the API (i.e. it is `Anthropic` not `claude`)

soulFood5632: I think `Cohere` is the blanket term so it should be fine.

soulFood5632 (approved):
LGTM

PR #715, opened 2025-09-17
Updated release pipeline for mkdocs 

soulFood5632:
## What does this add?

Removed an old reference to api docs release pipeline that referenced requirement.dev


Amir-R25 (approved):
LGTM

PR #759, opened 2025-10-08
Increase mcp version

CoronRing:
## What does this add?

Increase mcp required base version from 1.9.0 to 1.16.0. All test and local repo are using 1.16.0 by default so there is no code impact. 

The reason of the change is that some features we used only exist >= 1.12, (mcp.os), hence it is worth updating. The reason for bumping directly to 1.16.0 is that all versions of 1.12 and above are backward safe. We wil also enjoy extra feature and supporting starting with latest version

## Type of changes

- [X ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [X ] 🗑️ Chore (other changes that don't modify src or test files)
\

### DIFF
--- a/packages/railtracks/pyproject.toml
+++ b/packages/railtracks/pyproject.toml
@@ -45,7 +45,7 @@ classifiers = [
 
 dependencies = [
     "colorama >= 0.4.6",
-    "litellm[proxy] >= 1.70.2",
+    "litellm[proxy] >= 1.78.0",
     "mcp >= 1.16.0",
     "openai >= 1.99.5, < 1.100.0",
     "pydantic >= 2.5.3, <3",


packages/railtracks-cli/pyproject.toml
```diff
@@ -37,14 +37,11 @@ classifiers = [
...
-    "Typing :: Typed"
+    "Typing :: Typed",
 ]
 dynamic = ["version", "description"]
 requires-python = ">=3.10"
-dependencies = [
-    "railtracks",
-    "watchdog >= 3.0.0"
-]
+dependencies = ["watchdog >= 3.0.0"]
```
jbueza-railtownai, line 44: railtracks removed as dependency?

Amir-R25: He's investigating the test failures with circular import structures, I don't think the PRs ready for review yet

Amir-R25:
Thanks @Granine, I'll just add this here for tracking: https://github.com/BerriAI/litellm/pull/16307

PR #770, opened 2025-10-16
760 docs improve readme-patch for pypi

CoronRing:
## What does this add?
Minor patch to clean up readme on pypi

### DIFF
--- a/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
+++ b/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
@@ -331,11 +331,9 @@ class NodeBuilder(Generic[_TNode]):
             f"You tried to add prepare_tool_llm to a non LLM Node of {type(self._node_class)}."
         )
 
-        def prepare_tool(cls, tool_parameters: Dict[str, Any]):
+        def prepare_tool(cls, **kwargs):
             # Use the shared implementation in LLMBase
-            message_hist = cls.prepare_tool_message_history(
-                tool_parameters, tool_params
-            )
+            message_hist = cls.prepare_tool_message_history(kwargs, tool_params)
             return cls(message_hist)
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -51,10 +53,16 @@
...
 
-## ✨ What is Railtracks?
+## ✨ What is $${{\color{YellowOrange}\huge\textbf{\textsf{Railtracks}}}}$$?
+
+<div style="display: flex; align-items: flex-start;">
+  <img src="docs/assets/elements/banner.svg" align="left" width="4" height="50">
+  Easy agent building, for no one but <strong>YOU</strong>: While other frameworks force you into rigid workflows or complex APIs, Railtracks lets you create deployable complex agents using Pythonic style with natural control flow.
+</div>
+
+<!-- Add an empty line or a horizontal rule to ensure separation -->
```
soulFood5632, line 61: Probably best to remove this line as it looks very AI generated

README.md
```diff
@@ -2,6 +2,8 @@
 
 <!--Happy Coding ◊-->
 
+<!--Happy Coding ◊ -->
```
soulFood5632, line 3: duplicated

README.md
CoronRing: Easy agent building, for no one but <strong>YOU</strong>: Create deployable complex agents using simple, Pythonic style interface with natural control flow.

How does this sound? I want the first sentence do be a bit stronger so it grabs attentions. We can think of a joke or something else with similar effect

soulFood5632: Oh sorry, I should have been clearer. I just meant the `<!-- --->`. I like both of them. I would lean your second since it is a bit smaller.

CoronRing: Removed

soulFood5632 (approved):
LGTM

PR #776, opened 2025-10-23
Updated Logging Scope

### LINKED ISSUE #754: [Bug] Logging - Session Scoping
### Description

This is not a technical bug but rather a point of discussion.

Currently using the `railtracks` logger has a potential problem/inconsistency in that the logging lines only are displayed/logged when inside a session. We need to rethink and decide where to actually initiate the logger.

If you run the following code block:
```python
import railtracks as rt
import asyncio

ChatAgent = rt.agent_node(
    name="ChatAgent",
    system_message="You are a helpful assistant that summarizes conversations.",
)

async def analysis():
    response = await rt.interactive.local_chat(ChatAgent)

asyncio.run(analysis())
```
you get these logs:
```
[+4.456  s] RT          : INFO     - START CREATED ChatAgent
[+9.555  s] RT          : INFO     - ChatAgent DONE
[+9.557  s] RT.Session  : INFO     - Saving execution info to .railtracks/5e9ec440-9e0d-4b66-a663-1d1cd060dd6d.json
```

however if you include `rt.session` you get:
```python
import railtracks as rt
import asyncio

ChatAgent = rt.agent_node(
    name="ChatAgent",
    system_message="You are a helpful assistant that summarizes conversations.",
)

@rt.session
async def analysis():
    response = await rt.interactive.local_chat(ChatAgent)

asyncio.run(analysis())
```
you get:
```
[+0.919  s] RT.Interactive: INFO     - Connecting with Local Chat Session
[+2.057  s] RT.ChatUI   : INFO     - ChatUI server started at http://127.0.0.1:8000
[+4.148  s] RT          : INFO     - START CREATED ChatAgent
[+10.209 s] RT          : INFO     - ChatAgent DONE
[+12.567 s] RT.Interactive: INFO     - Ended Local Chat Session
[+12.568 s] RT.Session  : INFO     - Saving execution info to .railtracks/32aaa0e1-53ef-4871-88f5-a5d07bf5ab01.json
```

### Additional Context

As mentioned, this is not a bug, the reason this is happening is that in the first case, there's no session until we run `rt.call` and that's why only logs after that point are printed whereas in the second case we initiate the session before entering `interactive.local_chat` and therefore we print the additional logs

### Operating System

_No response_

### railtracks Version

_No response_

### Python Version

_No response_

Amir-R25:
## What does this add?
This PR updates logging configuration. After merge, we will support both global/module-level logging configuration (including environment variable support) and per-session logging overrides.

* Added global/module-level logging initialization via `initialize_module_logging()`, which reads from environment variables (`RT_LOG_LEVEL`, `RT_LOG_FILE` if available) and sets up logging when Railtracks is imported. 
* Updated runtime logging configuration updates allowing users to change logging level and file for the entire module, except when a session override is active. 
* Added logic in `Session` to detect custom logging parameters and override module-level logging for the session's duration
* Updated documentation to explain environment variable configuration and clarified docstrings for logging parameters, emphasizing the scoping behavior between global and session settings. 
* Added new unit tests
* Refactored logger setup logicsimplifying handler management for different logging levels.
* Updated pyproject.toml to handle having identical file names in different modules in our tests directory

Upon merge, this closes #754 

## Type of changes

### Code Quality

### Testing

### Documentation

### Git & PR Management

## Final Product

logger initiated at import time
```python
import railtracks as rt
```

logger configuration updated:
```python
import railtracks as rt
rt.set_config(logging_setting="DEBUG", log_file="my_logs.log")
```

logger configuration in `.env` file:
```
RT_LOG_LEVEL = "CRITICAL"
RT_LOG_FILE = "rt_logs.log"
```

### DIFF
--- a/.github/workflows/pr_tests.yaml
+++ b/.github/workflows/pr_tests.yaml
@@ -43,6 +43,10 @@ jobs:
         with:
           python-version: "${{ env.PYTHON_VERSION }}"
 
+      - name: Ensure assets directory exists
+        run: |
+          mkdir -p packages/railtracks
+
       - name: Use root README for PyPI (railtracks)
         run: |
           cp README.md packages/railtracks/README.md
--- a/.github/workflows/release_package.yaml
+++ b/.github/workflows/release_package.yaml
@@ -24,6 +24,10 @@ jobs:
         with:
           python-version: "3.10"
 
+      - name: Ensure assets directory exists
+        run: |
+          mkdir -p packages/railtracks/docs/assets
+
       - name: Use root README for PyPI (railtracks)
         run: |
           cp README.md packages/railtracks/README.md
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,6 @@ core-licenses-summary.md
 
 # OS generated files
 .DS_Store
+
+# Ignore Files copied during release workflow for PyPI packaging (for Flit)
+packages/railtracks/docs/


packages/railtracks/src/railtracks/context/central.py
```diff
@@ -382,6 +383,11 @@ def set_config(
         )
 
     config = global_executor_config.get()
+
+    if logging_setting or log_file:
```
soulFood5632, line 387: I would do explicit checks to make sure they are non (i.e. `logging_setting is not None or log_file is not None)

packages/railtracks/src/railtracks/utils/logging/config.py
```diff
@@ -190,18 +132,26 @@ def prepare_logger(
...
+    console_handler = logging.StreamHandler()
+    formatter = ColorfulFormatter(fmt=_default_format_string)
+    console_handler.setFormatter(formatter)
+
+    logger = logging.getLogger(rt_logger_name)
+    logger.addHandler(console_handler)
+
     if setting == "VERBOSE":
-        setup_verbose_logger_config()
+        logger.setLevel(logging.DEBUG)
```
soulFood5632: I think we should have some quick further discussion on this. I think it should be the handler that gets a logging level not the logger itself. 

(I'm gonna quickly look into best practices here)

packages/railtracks/src/railtracks/utils/logging/config.py
```diff
@@ -212,3 +162,119 @@ def detach_logging_handlers():
...
+    Mark that a session has overridden module-level logging.
+
+    Stores the current module config for later restoration and applies
+    the session-specific logging configuration.
+
+    Args:
+        session_level: The session's logging level
+        session_log_file: The session's log file (or None)
+    """
+    global _session_has_override, _pre_session_config
```
soulFood5632: There are a couple of cases that worry me.
We need to test this out to see what happens in the following cases:
(each case will be each comment for clarity)

packages/railtracks/src/railtracks/utils/logging/config.py
```diff
@@ -212,3 +162,119 @@ def detach_logging_handlers():
...
+
```
soulFood5632, line 322: 1. Multiple session overrides at the same time

soulFood5632, line 322: 2. module level override during 1 session

soulFood5632, line 322: 3. Module level override during multiple sessions

packages/railtracks/src/railtracks/utils/logging/config.py
soulFood5632: I would say in general I think we should be using `contextvars` to pass around this state like info. globals like these propagate to every async "thread" whereas `contextvars` are scoped and propagate more like a tree.

packages/railtracks/src/railtracks/utils/logging/config.py
```diff
@@ -17,6 +17,14 @@
 
 
 _file_format_string = "%(asctime)s %(levelname)s - %(message)s"
+
+_module_logging_config: Dict[str, AllowableLogLevels | str | os.PathLike | None] = {
+    "level": None,
+    "log_file": None,
+}
+_session_has_override: bool = False
+_pre_session_config: Dict[str, AllowableLogLevels | str | os.PathLike | None] = {}
```
soulFood5632: see my below comments, but I think stateful variables should be `contextvars` to prevent leakage between threads (async stuff)

packages/railtracks/src/railtracks/utils/logging/config.py
soulFood5632, line 322: 4. One session overrides and the other doesn't at the same time.

packages/railtracks/src/railtracks/utils/logging/config.py
```diff
@@ -212,3 +162,119 @@ def detach_logging_handlers():
...
+    _pre_session_config = {
+        "level": _module_logging_config["level"],
+        "log_file": _module_logging_config["log_file"],
+    }
+    _session_has_override = True
+
+    prepare_logger(setting=session_level, path=session_log_file)
+
+
+def restore_module_logging() -> None:
```
soulFood5632, line 332: Walking through one of my above examples. Lets say someone does the following:
1. Sets up a module level config
2. Starts to sessions in parallel
3. One session ends so it restores the module-level
4. Then the other session ends but the _pre_session_config has been reset to none so it resets it to default. 


A concrete situation where this could come up is within an API. They set the module level logging at the top of the file then run many overrides in their session calls. This would break things I believe (from the way I understand things)

packages/railtracks/src/railtracks/_session.py
```diff
@@ -224,7 +228,10 @@ def _close(self):
         """
         # the publisher should have already been closed in `_run_base`
         self.rt_state.shutdown()
-        detach_logging_handlers()
+
```
soulFood5632, line 231: Yep, see my above comment about multithreading concerns here

soulFood5632, line 231: To present another potential angle of annoying behaviour. 

1. Lets say they set module level logging
2. Then they start 2 session concurrently (one with overrides logs and one with not)
3. The session with override logs would propogate its changes to the one that didn't. 
4. In addition when the first session finishes and restores module_logging, the handler for the one that didn't would magically update mid run.

packages/railtracks/src/railtracks/context/central.py
Amir-R25: This was done on purpose, since we have env variables and it's very common for env variables to be `""` when they're not needed

packages/railtracks/src/railtracks/utils/logging/config.py
Amir-R25: I prefer to avoid added complexity of contextvars when they're not need but happy to discuss if there are concerns

packages/railtracks/src/railtracks/utils/logging/config.py
Amir-R25: how can multiple sessions be happening at the same time?

packages/railtracks/src/railtracks/_session.py
Amir-R25: It will be helpful if we have written examples of this happening so it can be addressed, for instance, I thought there's no way for two sessions to run concurrently within the same process

Amir-R25: I'll see if I can create a few

packages/railtracks/src/railtracks/context/central.py
soulFood5632: Ok sounds good 👍

PR #781, opened 2025-10-28
760 docs improve readme

CoronRing:
## What does this add?
bug fix to ensure repo exists before coying text over.


.github/workflows/pr_tests.yaml
```diff
@@ -43,6 +43,10 @@ jobs:
         with:
           python-version: "${{ env.PYTHON_VERSION }}"
 
+      - name: Ensure assets directory exists
```
soulFood5632, line 46: why do we need to do this?

soulFood5632 (approved):
LGTM, I don't see why we needed to change the PR tests as well. but it looks good.

PR #788, opened 2025-10-29
783 bug release failing

CoronRing:
## What does this add?
flit do not take non-python file, it is now resolved.


packages/railtracks/pyproject.toml
```diff
@@ -10,7 +10,7 @@ authors = [
     { name = "Jaime Bueza", email = "jaime@railtown.ai" },
     { name = "Amir Refaee", email = "amir@railtown.ai" },
     { name = "Aryan Ballani", email = "aryan@railtown.ai" },
-    { name = "Tristan Brown", email = "tristan@railtown.ai" }
+    { name = "Tristan Brown", email = "tristan@railtown.ai" },
```
Amir-R25, line 13: add yourself too Guan :D

CoronRing: Next PR I will!

PR #811, opened 2025-11-14
Tristan/vector stores wrappers

Tristan-railtown:
## What does this add?
Vector store capabilities with different providers for local or non-local stores.

## Background context
Wanted an easier way to create and use vector stores for various applications

### Code Quality

### Testing
- [ ] Test coverage maintained

### Documentation

### Git & PR Management

### Breaking Changes

---

### DIFF
--- a/packages/railtracks/src/railtracks/__init__.py
+++ b/packages/railtracks/src/railtracks/__init__.py
@@ -31,6 +31,7 @@ __all__ = [
     "create_mcp_server",
     "ToolManifest",
     "session_id",
+    "vector_stores",
 ]
 
 
@@ -39,7 +40,7 @@ from railtracks.built_nodes.easy_usage_wrappers import (
     function_node,
 )
 
-from . import context, integrations, llm, prebuilt
+from . import context, integrations, llm, prebuilt, vector_stores
 from ._session import ExecutionInfo, Session, session
 from .context.central import session_id, set_config
 from .interaction import broadcast, call, call_batch, interactive


packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,173 @@
...
+    metadata: Dict[str, Any]= field(default_factory=dict)
+
+    def __post_init__(self) -> None:
+        """Ensure metadata is always a dict."""
+        if self.metadata is None:
+            self.metadata = {}
+
+@dataclass
+class SearchResult:
+    """Container for a single search result from a vector store query."""
```
Amir-R25: please expand the docstrings to include the args

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,173 @@
...
+    cosine = "cosine"  # Cosine similarity
+    l2 = "l2"  # Euclidean (L2) distance
+    dot = "dot"  # Dot product similarity
+
+class Document(str):
+    content :str
+
+@dataclass
+class Chunk:
+    """
```
Amir-R25: please expand the docstrings to include the args

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,173 @@
...
+    document: Optional[str] = None
+    metadata: Dict[str, Any] | Mapping[str, Any] = field(default_factory=dict)
+    
+    def __post_init__(self) -> None:
+        """Ensure metadata is always a dict."""
+        if self.metadata is None:
+            self.metadata = {}
+@dataclass
+class FetchResult:
+    """Container for a single fetch result from a vector store fetch."""
```
Amir-R25: please expand the docstrings to include the args

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,173 @@
...
+class Metric(str, Enum):
+    """
+    Enumeration of supported distance/similarity metrics for vector comparisons.
+    """
+
+    cosine = "cosine"  # Cosine similarity
+    l2 = "l2"  # Euclidean (L2) distance
+    dot = "dot"  # Dot product similarity
+
+class Document(str):
```
Amir-R25: nitpick, docstrings :D

packages/railtracks/src/railtracks/vector_stores/milvus.py
```diff
@@ -0,0 +1,49 @@
+from typing import List, Dict, Any, Optional, Callable
```
Amir-R25: If I remember correctly you said this was incomplete so let's exclude it from this PR.

packages/railtracks/src/railtracks/vector_stores/pinecone.py
```diff
@@ -0,0 +1,207 @@
+from typing import List, Dict, Any, Optional, Callable, Union
```
Amir-R25: same here, If I remember correctly you said this was incomplete so let's exclude it from this PR.

packages/railtracks/src/railtracks/vector_stores/chroma.py
```diff
@@ -0,0 +1,212 @@
...
+                metadata=metadata
+                )
+            )
+            
+        return results
+    
+    #There is support for other types of query modalities but for now just list of strings
+    #Should Probably add support for Chunks as well
+    #TODO cross reference and fix typing plust batch querys problems. Could be an issue with the large nested nature of this
+    def search(
```
Amir-R25, line 266: Typing is broken here since parameter types are different than base class

Amir-R25: can probably be added later but we should have `asearch` as well (async search)

packages/railtracks/src/railtracks/rag/chunking_methods.py
```diff
@@ -0,0 +1,599 @@
+from typing import List, Union
+import tiktoken
+import asyncio
+from collections import deque
+import railtracks as rt
+
+def strict_chunk_file(file_path : str, token_encoding : str | None = None, chunk_size : int = 400, overlap : int = 200) -> List[str]:
```
Amir-R25: what types of files does this accept? ie txt, json, etc

packages/railtracks/src/railtracks/rag/chunking_methods.py
```diff
@@ -0,0 +1,599 @@
...
+#Based off Chroma implementation
+def recursive_chunk_text(
+    text: str,
+    chunk_size: int = 400,
+    overlap: int = 200,
+    delta: int = 100,
+    separators: List[str] = ["\n\n", "\n", ".", "?", "!", " ", ""]
+) -> List[str]:
+    """
+    Recursively splits text into chunks using hierarchical separators.
```
Amir-R25: complete docstrings

packages/railtracks/src/railtracks/rag/chunking_methods.py
```diff
@@ -0,0 +1,599 @@
...
+
+            #Update char count and index of last chunk added to window
+            char_count += len(chunked_text[i])
+            last_chunk = i
+            if char_count > window_size:
+                break
+        
+        # If we've reached the end, process everything remaining
+        if last_chunk == len(chunked_text) - 1:
+            split_points = asyncio.run(agent_get_split_points(window_text, agent))
```
Amir-R25: this and line 315 will break stuff if they're in an async context. For example if you have an agent using this function and then you run that agent you'll get this scenario:

```python
async def main_agent(..):
     llm_chunk_text(...)
     rt.call(SomeOtherAgent...)

if __name__=='__main__':
      asyncio.run(main_agent)
```

Amir-R25 (changes requested):
Great work and look forward to the final version!

I know it's a WIP so just provided general feedback. I think it might make it easier to decrease the scale of the PR by removing pincone and milvus and add them in separately.

Also for things you want to complete later, you can create an issue and then link the issue in the code instead of the TODO if you want to scale things back for this

packages/railtracks/src/railtracks/rag/chunking_methods.py
soulFood5632, line 1: Remind me here @Tristan-railtown, these methods can be added to the embedding service. Or are they a bit unrefined?

soulFood5632: Either way i think it would make sense to take these methods out of the PR and create a new branch with these changes in it so we can migrate the logic over.

packages/railtracks/src/railtracks/rag/chunking_methods.py
soulFood5632: Yea this will indeed break. i think we just derisk the current PR and move this stuff into a seperate branch so we can continue refining the methods we expose and chunking algoirthims (see #819)

packages/railtracks/src/railtracks/rag/chunking_methods.py
```diff
@@ -0,0 +1,599 @@
...
+
+    Returns:
+        float: Estimated characters per token.
+    """
+    sample = text[:min(len(text), sample_size)]
+    tokens = len(tokenizer.encode(sample))
+    return len(sample) / tokens if tokens > 0 else 4.0
+
+
+def find_best_boundary(
```
soulFood5632: This is a more general comment, but some of these methods should be private to this file. You can do that by putting an undescore before the method name. so `_find_best_boundary`

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+
+@dataclass
+class FetchResult:
+    """Container for a single fetch result from a vector store fetch."""
+
+    id: str
+    content: str
+    vector: List[float]
+    document: Optional[str] = None
+    metadata: Dict[str, Any] | Mapping[str, Any] = field(default_factory=dict)
```
soulFood5632: A dict should already be a mapping, so just Mapping should work here

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+@dataclass
+class SearchResult:
+    """Container for a single search result from a vector store query."""
+
+    id: str
+    distance: float
+    content: str
+    vector: List[float]
+    document: Optional[str] = None
+    metadata: Dict[str, Any] | Mapping[str, Any] = field(default_factory=dict)
```
soulFood5632: A dict should already be a mapping, so just Mapping should work here

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def __post_init__(self) -> None:
+        """Ensure metadata is always a dict."""
+        if self.metadata is None:
+            self.metadata = {}
+
+
+# For Readability
+class SearchResponse(List[SearchResult]):
+    """Containter for results of search"""
```
soulFood5632: something like this is the correct way to do this type of type aliasing 
```
SearchResponse = list[str]

def hello_world(data: str) -> SearchResponse:
    return data.split()

```

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class FetchResponse(List[FetchResult]):
```
soulFood5632: ditto 

something like this is the correct way to do this type of type aliasing 
```
SearchResponse = list[str]

def hello_world(data: str) -> SearchResponse:
    return data.split()

```

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+    """Container for results of a Fetch."""
+
+
+class VectorStore(ABC):
+    """Abstract base class for vector store implementations."""
+
+    @abstractmethod
```
soulFood5632: By putting this an abstract method it will limit the flexability of future method signatures in concrete base classes. This may limit our ability to handle #820

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns:
+            FetchResponse containing the fetched vectors
+        """
+        pass
+
+    @abstractmethod
+    def search(
+        self,
+        query: List[Chunk] | List[str],
+        n_results: int = 10,
```
soulFood5632: Is this the common term and `top_k`. I very well could be wrong, but that is the term i am more familiar with

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Returns:
+            FetchResponse containing the fetched vectors
+        """
+        pass
+
+    @abstractmethod
+    def search(
+        self,
+        query: List[Chunk] | List[str],
```
soulFood5632: Should we also allow for the input of a single str/chunk?

packages/railtracks/src/railtracks/vector_stores/vector_store.py
soulFood5632, line 1: Overall, this structure is very clean. I think these simple base operations and their corresponding signatures make sense. I left a couple of minor comments around things

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class VectorStore(ABC):
+    """Abstract base class for vector store implementations."""
+
+    @abstractmethod
+    def __init__(
+        self,
+        collection_name: str,
+        embedding_function: Callable[[List[str]], List[List[float]]],
```
soulFood5632: It may be good design to handle the saving of these properties in the base init here. It should be as simple as:

```
self.collection_name = collection_name
self.embedding_function = embedding_function 
```

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+        Args:
+            collection_name: Name of the collection/index
+            embedding_function: Function to convert text to embeddings
+        """
+        pass
+
+    @abstractmethod
+    def upsert(
+        self,
+        content: List[Chunk] | List[str],
```
soulFood5632: Again here, we should also allow for people to input single entries.

packages/railtracks/src/railtracks/vector_stores/vector_store.py
```diff
@@ -0,0 +1,176 @@
...
+        Args:
+            content: List of Chunk objects or raw text strings to be upserted
+
+        Returns:
+            List of IDs of the upserted vectors
+        """
+        pass
+
+    @abstractmethod
+    def fetch(self, ids: List[str]) -> FetchResponse:
```
soulFood5632: see above, but we should also allow for a single string

packages/railtracks/src/railtracks/vector_stores/pinecone.py
```diff
@@ -0,0 +1,221 @@
...
+        deleted = self._collection.delete(
+            ids,
+            delete_all,
+            self._collection_name,
+            filter,
+        )
+
+        return deleted
+
+    # TODO Test
```
soulFood5632: Lets delete this and make sure it is tested in #817

packages/railtracks/src/railtracks/vector_stores/pinecone.py
```diff
@@ -0,0 +1,221 @@
...
+)
+from uuid import uuid4
+
+
+class PineconeVectorStore(VectorStore):
+    """Pinecone implementation of VectorStore."""
+
+    @classmethod
+    def class_init(
+        cls,
```
soulFood5632: I think its a good idea to add type hints here. But this should be handled in a future issue

packages/railtracks/src/railtracks/vector_stores/pinecone.py
soulFood5632, line 1: I'm gonna avoid going to in depth with this file, this will be handled in a seperate issue.

packages/railtracks/src/railtracks/vector_stores/milvus.py
```diff
@@ -0,0 +1,107 @@
...
+from .vector_store import VectorStore, SearchResponse, FetchResponse, Chunk
+
+
+class MilvusVectorStore(VectorStore):
+    """Milvus implementation of VectorStore. Currently only local Vector Stores"""
+
+    @classmethod
+    def class_init(cls, db_name: str):
+        if not hasattr(cls, "_milvus"):
+            # First object initialized so import pinecone and connect to server
```
soulFood5632: Copy pasted pinecone

packages/railtracks/src/railtracks/vector_stores/milvus.py
soulFood5632, line 1: yep agreed with Amir here, What i think the best to do is to remove each of these unused things into their own branches. 

I will make sure there is issues for you to tie those branches to for the completion of milvus and pinecone

packages/railtracks/src/railtracks/vector_stores/chroma.py
```diff
@@ -0,0 +1,227 @@
...
+
+            embeddings.append(embedding)
+            metadatas.append(metadata)
+
+        self._collection.upsert(
+            ids=ids, embeddings=embeddings, metadatas=metadatas, documents=documents
+        )
+        return ids
+
+    # TODO cross reference
```
soulFood5632: mark to delete TODO if complete

packages/railtracks/src/railtracks/vector_stores/chroma.py
```diff
@@ -0,0 +1,227 @@
...
+    from chromadb.api.types import Include
+
+CONTENT = MetadataKeys.CONTENT.value
+
+
+class ChromaVectorStore(VectorStore):
+    """ChromaDB implementation of VectorStore."""
+
+    @classmethod
+    def class_init(cls, path, host, port):
```
soulFood5632: maybe add type hints as well.

packages/railtracks/src/railtracks/vector_stores/chroma.py
```diff
@@ -0,0 +1,227 @@
...
+                        host=host, port=port
+                    )  # Currently this does not provide the extra asycn features that comes with this
+                else:
+                    cls._chroma = chromadb.EphemeralClient()
+            except ImportError:
+                raise ImportError(
+                    "Chroma package is not installed. Please install railtracks[chroma]."
+                )
+
+    def __init__(
```
soulFood5632, line 87: add some quick fail fast type stuff to handle when the user provides an unknown sequence of parameters. 

i.e. path and port, nothing but host, etc.

packages/railtracks/src/railtracks/vector_stores/vector_store.py
Tristan-railtown: Yes I left a note for myself to look into this but Chroma uses a type called OneOrMany which is basically this. I will be looking into this to see if it would make more sense to use that class, if not I'll just union the single version?

packages/railtracks/src/railtracks/vector_stores/vector_store.py
Tristan-railtown: Just to make sure I understand you want me to just leave the init unimplemented so that we do not lock ourselves into inits that require collection_name and embedding_function for concrete classes?

packages/railtracks/src/railtracks/vector_stores/vector_store.py
Tristan-railtown: Also wondering about this with regards to the comment about making something abstract

soulFood5632: Yea its connected. Basically we should make the base class __init__ not abstract. That way the base class still expects the same params but we have complete freedom of the other __init__'s

packages/railtracks/src/railtracks/vector_stores/vector_store.py
soulFood5632: Yea, that makes sense to me. You can easily implement a type alias or use the unions directly.

packages/railtracks/src/railtracks/vector_stores/vector_store.py
soulFood5632: You can implement it (see my other comment), just don't make it abstract so you can add other parameters in the concrete base classes. Basically it will make it so you can call `super().__init__(collection_name, embedding_function)`. But using this approach (non-abstract) it will allow the inits for each base class to be different. and potentially add other params

packages/railtracks/src/railtracks/vector_stores/chroma.py
Tristan-railtown: Yea I think I'll add that as an issue to add to later

docs/tutorials/vector_store_tutorial.md
```diff
@@ -0,0 +1,164 @@
...
+- A Chunk object or list of Chunk objects (if including document data or metadata)
+
+### Simple Insert
+```python
+ids = store.upsert(["Bananas are yellow.", "Apples can be red or green."])
+```
+
+### Insert With Metadata
+```python
+from railtracks.vector_stores.vector_store_base import Chunk
```
soulFood5632: Can you move this code over to `docs/scripts`

docs/tutorials/vector_store_tutorial.md
```diff
@@ -0,0 +1,164 @@
...
+    path="./chroma-data"
+)
+```
+!!! note "Choosing where your vector store is located"
+    You can look at our different integrations to location options for your stores
+
+## Step 2: Insert Your Documents
+
+You can upsert:
+- A string or list of strings
```
soulFood5632, line 21: This isn't rendering properly: should have an extra enter statement here

docs/tutorials/vector_store_tutorial.md
```diff
@@ -0,0 +1,164 @@
...
+```python
+from railtracks.vector_stores.chroma import ChromaVectorStore 
+store = ChromaVectorStore(
+    collection_name="my_collection",
+    embedding_function=my_embedding_function,
+    path="./chroma-data"
+)
+```
+!!! note "Choosing where your vector store is located"
+    You can look at our different integrations to location options for your stores
```
soulFood5632: We should provide a page for each in the future. I will create an issue for this

docs/tutorials/vector_store_tutorial.md
```diff
@@ -0,0 +1,164 @@
...
+    return model.encode(texts).tolist()
+
+
+store = ChromaVectorStore(
+    collection_name="article_archive",
+    embedding_function=my_embedding_function,
+    path="./chroma-data" #Specify local path to get existing store from or create new one
+)
+
+articles_data = [
```
soulFood5632: It would be nice to have an example for how to import a file for this. I will descope this PR and add an issue to complete this.

docs/vector_stores/integrations/chroma.md
```diff
@@ -0,0 +1,79 @@
+# ChromaVectorStore
+
+Railtracks' built-in implementation using ChromaDB.
+```python
+class ChromaVectorStore(VectorStore):
```
soulFood5632: I would avoid putting actual source code like this here. We have the API documentation to handle this type of analysis. 

Honestly, this entire page should be focused on the specific requirements of setting things up with Chroma.

soulFood5632: You do a good job for this below btw just some of this intro is kinda irrelvant

docs/vector_stores/integrations/chroma.md
```diff
@@ -0,0 +1,79 @@
...
+    ...
+```
+
+### Initialization and storage backends
+
+Create a `ChromaVectorStore` by supplying a `collection_name` and an
```
soulFood5632: I would restructure this section:
Basic requirements:
 - collection name
 - embedding function
??? In Memory 
  ...
??? http instance
  ...

etc.

docs/vector_stores/integrations/chroma.md
```diff
@@ -0,0 +1,79 @@
...
+This section briefly documents what `ChromaVectorStore` adds on top of
+the generic vector-store API (the general semantics live in
+`vector_store_info.md`).
+
+- Initialization: `ChromaVectorStore.class_init` chooses between a
+        persistent, HTTP or ephemeral client based on `path` or `(host, port)`.
+        Provide a filesystem `path` (persistent) or both `host` and `port`
+        (HTTP). Passing an invalid/mixed combination raises `ValueError`.
+
+- fetch extras: `fetch(ids=None, where=None, where_document=None)`
```
soulFood5632: This doesn't render porely

docs/vector_stores/integrations/chroma.md
```diff
@@ -0,0 +1,79 @@
...
+        - `where/where_document` : Allows for metadata filtering and document filtering
```
soulFood5632: I am pretty confused by what this section is saying. The limit paramater isn't in the above siganture which makes it confusing

docs/vector_stores/integrations/chroma.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+
+### Chroma-specific behaviour
+
+This section briefly documents what `ChromaVectorStore` adds on top of
+the generic vector-store API (the general semantics live in
+`vector_store_info.md`).
+
+- Initialization: `ChromaVectorStore.class_init` chooses between a
```
soulFood5632: I think this is too detailed information to include here. They really don't need to know about how we do this

docs/vector_stores/integrations/chroma.md
soulFood5632: We should not put method signatures here. It is impossible to mantain if we ever make changes

docs/vector_stores/vector_store_info.md
```diff
@@ -0,0 +1,188 @@
+
+# Vector Store API Reference
+This page documents the core interface for embedding storage and retrieval in Railtracks. All vector store implementations follow this unified API, allowing you to swap backends without changing your code.
+
+## When to Use Vector Stores
+!!! tip "Vector Stores Shine When"
+- You have large collections of text that won't fit in a single prompt
```
soulFood5632: You need indents here

docs/vector_stores/vector_store_info.md
```diff
@@ -0,0 +1,188 @@
...
+- Your application uses RAG (Retrieval Augmented Generation)
+- You want metadata-driven filtering for targeted queries
+- You're building a knowledge base or document search system
+!!! warning "Skip Vector Stores If"
+- Your entire dataset fits comfortably in one prompt (< 100K tokens)
+- You're not doing similarity search or retrieval
+- Your use case requires exact matching (use a traditional database)
+- Approximate nearest neighbor search adds no value
+!!! tip "Vector Stores Shine When"
+    - you have large collections of text
```
soulFood5632: This section reads like it is duplicated. Not super sure

docs/vector_stores/vector_store_info.md
```diff
@@ -0,0 +1,188 @@
...
+    - you want metadata-driven filtering (where)
+
+!!! warning "Skip Vector Stores If"
+    - your whole dataset fits in one prompt
+    - you're not doing retrieval
+    - approximate search adds no value for your case
+---
+
+## Core Methods
+#### `upsert(content: OneOrMany[Chunk] | OneOrMany[str]) -> list[str]`
```
soulFood5632: Again I think it is best to avoid the API reference directly. Especially since things are changing fast

soulFood5632: I think we would be better served to describe the core methos without the signatures in a short table form with a vague description of what they do.

soulFood5632 (changes requested):
Some comments around the docs. I will try to derisk this PR as much as I can by creating issues for non-critical items but there are still some things that need to be modified

docs/vector_stores/vector_store_info.md
soulFood5632: Like to be clear I think we just get rid of this section generally. This page should be a quick and dirty intro into the world of vector stores with a list of the ones we support, and a further link to page describing how to create your own integration.

soulFood5632:
@Tristan-railtown It looks some of the typing is broken in your examples. It could be a symptom of some typing issues in the underlying code. I'll let you untangle the root cause

soulFood5632 (approved):
Looks good to me.

PR #839, opened 2025-11-21
DataPoint class to reprent a single instance of an agent data

Amir-R25:
## What does this add?

`DataPoint` class for representing an "instance" of an agentic interaction.


soulFood5632 (approved):
LGTM

PR #840, opened 2025-11-21
metrics class for storing evaluation results

Amir-R25:
`Metrics` class for storing evaluation results. 

- Currently the two child classes supported are `Categorical` and `Continuous`.
- It was requested that we generate a hash for equality of metrics so please review that portion
- Some test examples for you to try things out


soulFood5632 (approved):
LGTM

PR #857, opened 2025-12-02
Implement Chunking Base class, fixed Token Chunking, and media parser

Tristan-railtown:
## Type of changes

## Background context
This is the beginning of chunking service so there may be some roughness around it. The tokenizer, chunking classes, and media parser will likely all be added to in the future

### Code Quality

### Testing

### Git & PR Management

### DIFF
--- a/packages/railtracks/src/railtracks/vector_stores/chunking/base_chunker.py
+++ b/packages/railtracks/src/railtracks/vector_stores/chunking/base_chunker.py
@@ -161,8 +161,9 @@ class BaseChunker(ABC):
         """
         pass
 
-    @staticmethod
+    @classmethod
     def make_into_chunks(
+        cls,
         text_chunks: list[str],
         document: Optional[str] = None,
         metadata: Optional[dict[str, Any]] = None,
--- a/packages/railtracks/src/railtracks/vector_stores/chunking/media_parser.py
+++ b/packages/railtracks/src/railtracks/vector_stores/chunking/media_parser.py
@@ -1,7 +1,7 @@
 import os
 from typing import Optional
 
-import pymupdf
+import pdfplumber
 from charset_normalizer import from_path
 
 
@@ -13,11 +13,6 @@ class MediaParser:
         - .pdf
     """
 
-    _PARSERS = {
-        ".txt": "_parse_txt",
-        ".pdf": "_parse_pdf",
-    }
-
     @classmethod
     def get_text(cls, path: str, **kwargs) -> str:
         """Return cleaned text extracted from a supported file.
@@ -27,9 +22,13 @@ class MediaParser:
             **kwargs: Parser-specific arguments (e.g., encoding for .txt files)
         """
         ext = cls._get_extension(path)
-        handler_name = cls._PARSERS.get(ext)
+        if ext == ".txt":
+            handler_name = "_parse_txt"
 
-        if not handler_name:
+        elif ext == ".pdf":
+            handler_name = "_parse_pdf"
+
+        else:
             raise ValueError(f"Unsupported file type: {ext}")
 
         parser_function = getattr(cls, handler_name)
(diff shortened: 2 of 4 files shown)


packages/railtracks/src/railtracks/vector_stores/chunking/base_chunker.py
```diff
@@ -0,0 +1,193 @@
...
+        path: str,
+        encoding: Optional[str] = None,
+        document: Optional[str] = None,
+        metadata: Optional[dict[str, Any]] = None,
+    ) -> list[Chunk]:
+        """Split text into list of chunks from a specified file.
+
+
+        Args:
+            path (str): file path of where to retrieve text you want chunked.
```
Amir-R25: will the types of supported file come in inside the classes that implement the abstract class?

packages/railtracks/src/railtracks/vector_stores/chunking/semantic_chunker.py
```diff
@@ -0,0 +1,45 @@
+from typing import Optional
+
+import tiktoken
+
+from .base_chunker import BaseChunker
+
+
+class SentenceChunker(BaseChunker):
```
Amir-R25: Nitpick but I imagine this class will be completed later. Can you please link an issue inside the docstrings for it if that's the case?

packages/railtracks/src/railtracks/vector_stores/chunking/media_parser.py
```diff
@@ -0,0 +1,82 @@
...
+            return "\n".join(extracted)
+
+    @classmethod
+    def _clean_text(cls, text: str) -> str:
+        """Remove null bytes / non-printable characters while preserving whitespace."""
+        if not text:
+            return ""
+        return "".join(char for char in text if char.isprintable() or char in "\t\n\r")
+
+    @staticmethod
```
Amir-R25, line 77: @Tristan-railtown

soulFood5632:
> @Amir-R25 @soulFood5632 I use pymupdf but it's not on our pytoml. Not quite sure how you think I should add this though? Should I just include it for all versions that specify a version of vector stores? ie pip install railtracks[chroma] or do you think it would be good to make it it's own separate thing?

@Tristan-railtown I think an optional dependency makes the most sense to me. Maybe we can add a new optional dependency category called `files` or something like this. Do you see there being more dependencies for this down the road?

packages/railtracks/src/railtracks/vector_stores/chunking/base_chunker.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Args:
+            text (str): Raw text to chunk.
+
+        Returns:
+            list[str]: A list of strings produced by the chunking strategy.
+        """
+        pass
+
+    @staticmethod
```
soulFood5632: This is defiantly in the category of extremely pedantic, but let's make this a `classmethod` instead of a static. It would allow us to access any static properties of the class in a way that works better with OOP if that ever came up

packages/railtracks/src/railtracks/vector_stores/chunking/media_parser.py
```diff
@@ -0,0 +1,82 @@
...
+            path: Path to the file
+            **kwargs: Parser-specific arguments (e.g., encoding for .txt files)
+        """
+        ext = cls._get_extension(path)
+        handler_name = cls._PARSERS.get(ext)
+
+        if not handler_name:
+            raise ValueError(f"Unsupported file type: {ext}")
+
+        parser_function = getattr(cls, handler_name)
```
soulFood5632, line 34: Honestly this logic feels a bit spooky to me. Is there any way for us to refactor this it is a bit more type friendly. Like I think the following logic would be a bit easier to understand 

```
handler_name = cls._PARSERS(...)
if handler_name == ".txt":
  _parse_txt(...)
elif ...
```

what do you think?

I just find the above a bit hard to follow purely from a readability perspective

Tristan-railtown: Yea the idea was just to avoid a bunch of if statements as we get more and more types of files we want to use but I can do it this way for now since we only have 2 and we can make that call as the if statement grows

packages/railtracks/src/railtracks/vector_stores/chunking/base_chunker.py
Tristan-railtown: Class method for all methods in basechunker? Or also make those changes to media parser?

soulFood5632:
> @soulFood5632 Ok I'm thinking I'll make it so that should you install one of our vector store dependencies then you'll get the file parsing as well if that's ok with you. I'll make the change soon with the change in parsing library we use so you can see what it looks like.

Sounds good to me!

Issue #916, opened 2026-01-22
[Bug] Listener gets killed in an aysnc run


Pooria90:
### Description

I face this error when using `async.gather`:
`ValueError: Listener has been killed before receiving the correct message.`

### Reproducing Steps (include code examples if applicable)

The following code minimally simulates my original flow I was using for my own agent:
```python
import railtracks as rt
import asyncio
import random


@rt.function_node
async def random_generator() -> float:
    """
    A dummy tool for random sleeping.
    """
    r = random.random() * 10
    await asyncio.sleep(r)
    return r


dummy_agent = rt.agent_node(
    name="dummy_agent",
    tool_nodes=[random_generator],
    llm=rt.llm.OpenAILLM("gpt-4o-mini"),
    system_message="You are a dummy agent that invokes a random async sleep and returns the sleep duration. Use the random_generator tool to sleep for a random duration, and report how much time you slept.",
    manifest=rt.ToolManifest(
        description="A dummy tool for random sleeping.",
        parameters=[
            rt.llm.Parameter(name="prompt", description="The prompt to the agent.", param_type="str")
        ]
    )
)


async def body_function(n: int = 5):
    futures = []
    for i in range(n):
        futures.append(rt.call(dummy_agent, "Get some sleep!"))
    resp = await asyncio.gather(*futures)
    return resp


async def main():
    with rt.Session(name="dummy_agent"):
        res = await body_function(n = 5)
        print(res)


if __name__ == "__main__":
    asyncio.run(main())
```

This is a visualization from railtracks:

<img width="989" height="731" alt="Image" src="https://github.com/user-attachments/assets/544b1659-5674-467d-b0eb-e8fe3802739b" />

### Additional Context
I also saw that these tests triggered it:
```bash
FAILED packages/railtracks/tests/integration_tests/context/test_internal.py::test_run_id_propagation_multiple_runs_parallel[1] - ValueError: Listener has been killed before receiving the correct message.
FAILED packages/railtracks/tests/integration_tests/context/test_internal.py::test_run_id_propagation_multiple_runs_parallel[5] - ValueError: Listener has been killed before receiving the correct message.
```

### Operating System

Ubuntu 24.04

### railtracks Version

1.1.24

### Python Version

3.12.10

PR #924, opened 2026-01-27
Listener bug loop

Pooria90:
## What does this add?

Closes https://github.com/RailtownAI/railtracks/issues/916

The problem with this issue was that the event loop treats the shutdown like a usual async task. While this is a normal concurrent behavior, it can shutdown the publisher while there are still tasks in the queue pending completion. 

I added a condition for the shutdown and moved the close-publisher mechanism to the `_session.py`.

## Type of changes

### Code Quality

### Testing

### Documentation

### Git & PR Management

### Breaking Changes

---


packages/railtracks/src/railtracks/interaction/_call.py
```diff
@@ -170,7 +169,10 @@ async def wrapped_fut(f: Coroutine):
 
         raise GlobalTimeOutError(timeout=timeout)
     finally:
-        await shutdown_publisher()
+        # Only shutdown publisher if there's no Session context
+        # When Session exists, it manages the publisher lifecycle
+        if not is_context_present():
```
soulFood5632: Correct me if I am wrong but I think this is not needed. The pair of possible scenarios:

Session already exists. In this case the session should handle it. 
If the session does not exist, it should create a new one (see line 95). At which time time the session should be able to handle that.

packages/railtracks/src/railtracks/utils/publisher.py
```diff
@@ -64,6 +64,10 @@ async def __aenter__(self):
         return self
 
     async def start(self):
+        # Make start idempotent - if already running, don't restart
+        if self._running:
+            return
```
soulFood5632, line 69: Good idea

packages/railtracks/src/railtracks/interaction/_call.py
Pooria90: Yeah it's like it never passes the condition. The code works without it and nothing breaks.
My question is do we need to have something in the `finally` block at all?

soulFood5632: Maybe get rid of it and see if anything breaks. Logically I don't see why it would ever be needed.

Issue #939, opened 2026-02-05
[Docs] Update Docs for new Flows


soulFood5632:
### Issue Categorization

- [ ] Missing documentation
- [ ] Incorrect or outdated information
- [ ] Overly verbose or complex explanation
- [ ] Broken links or formatting
- [ ] Missing code examples
- [ ] Typos or grammar issues

### Documentation Problem

Update the documentation for the new changes seen in #933. We should be encouraging everybody to use the new approach. There is a lot of code that uses the previous paradigms that we should address

PR #946, opened 2026-02-05
933 general top layer abstraction difficulties

soulFood5632:
## What does this add?

Big change to the way we invoke agents. This change is backwards compatible but it is designed to entice users to use the new method. The invocation of agents is much simpler now 

```python 
Agent = rt.agent_node(
    system_message=system_prompt,
    name="Tool-Forward Guide",
    tool_nodes=[roll_dice, draw_card, check_weather, find_item, pick_path],
)


def save_payload(payload: dict[str, Any]):
    with open("payload_log.json", "w") as f:
        json.dump(payload, f, indent=4)


        
flow = Flow(
    name="Cozy Tool Adventure",
    entry_point=Agent,
    context={"time_of_day": "morning", "location": "town gate"},
    payload_callback=save_payload,
)


result = flow.invoke("Create a one-shot cozy adventure scene that uses tools.")
print(result)
````

## Type of changes

### DIFF
--- a/packages/railtracks/src/railtracks/orchestration/flow.py
+++ b/packages/railtracks/src/railtracks/orchestration/flow.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-import base64
 import hashlib
 import json
 import os
@@ -104,7 +103,7 @@ class Flow(Generic[_P, _TOutput]):
         """
         config_string = json.dumps(self._get_hash_content(), sort_keys=True)
         return hashlib.sha256(config_string.encode()).hexdigest()
-    
+
     def _get_hash_content(self) -> dict:
         return {
             "name": self.name,

### CODE REVIEW

File: docs/scripts/quickstart.py [hunk 1]
```diff
@@ -9,13 +9,17 @@
...
     result = await rt.call(
         Agent,
         "Hello, what can you do?"
     )
     return result
 
-result = asyncio.run(main())
+flow = rt.Flow("Quickstart Example", entry_point=main)
+
+result =flow.invoke()
```
Amir-R25: nitpick, could you please add an example that takes an input argument so the users can see how to pass an argument upon invoking a flow

docs/index.md
```diff
@@ -27,7 +27,7 @@ In Railtracks, agent behavior is defined directly in Python. There is **no** con
 
 Building a sequential flow is just like any the code you have written in your life. 
 ```python
-@session
+@node
```
Amir-R25: let's chat about this a little bit more, we will have:
- function_node
- tool_node
- agent_node
- and now node

The abstractions are now at risk of seeming unclear to users

packages/railtracks/src/railtracks/orchestration/flow.py
```diff
@@ -0,0 +1,104 @@
...
+    def invoke(self, *args: _P.args, **kwargs: _P.kwargs) -> _TOutput:
+        try:
+            return asyncio.run(self.ainvoke(*args, **kwargs))
+
+        except RuntimeError:
+            raise RuntimeError(
+                "Cannot invoke flow synchronously within an active event loop. Use 'ainvoke' instead."
+            )
+
+    def equality_hash(self) -> str:
```
Amir-R25, line 100: I do recommend we separate out the hashing and the config str generation into different functions and also make them internal

docs/scripts/quickstart.py
soulFood5632: That is a great point. I will attach that to the work in #939

packages/railtracks/src/railtracks/orchestration/flow.py
soulFood5632: I'm lost. Are you suggestinga  function like config_dict and then this function calls it?

docs/index.md
soulFood5632: This is designed to just be psuedocode (not a new abstraction). I think I will change this to function_node for clarity

docs/scripts/quickstart.py
soulFood5632: I lied I will do in this PR

packages/railtracks/src/railtracks/orchestration/flow.py
soulFood5632: I get it

PR #971, opened 2026-02-27
fix(config): allow timeout=None in ExecutorConfig to disable timeout

### LINKED ISSUE #967: [Bug] setting `timeout` to `None` gets overwritten to 150 seconds in `ExecutorConfig`
### Description

We need to make sure that timeout is not set when user passes in None.

ref: https://github.com/RailtownAI/railtracks/blob/ac30cd8b152c09ac1928515d3d89649966b49370/packages/railtracks/src/railtracks/utils/config.py#L9-L23

N/A

### Additional Context

_No response_

### Operating System

_No response_

### railtracks Version

_No response_

### Python Version

_No response_

CoronRing:
## Summary

Fixes #967

## Root Cause

Two compounding issues in `ExecutorConfig`:

1. **Wrong type annotation**: `timeout: float = 150.0` did not allow `None`, making `ExecutorConfig(timeout=None)` a type error even though Python wouldn't enforce it at runtime.

2. **Sentinel collision in `precedence_overwritten`**: The parameter `timeout: float | None = None` used `None` to mean *both* "not provided" and "disable timeout". The guard `timeout if timeout is not None else self.timeout` always fell back to `self.timeout` (e.g. `150.0`) when `None` was explicitly passed, making it impossible to propagate `None` through `precedence_overwritten`.

## Changes

- `config.py`: Added module-level `_UNSET = object()` sentinel
- `config.py`: Changed `timeout: float` to `timeout: float | None` in `__init__`
- `config.py`: Changed `precedence_overwritten` timeout default from `None` to `_UNSET`; updated guard to `is not _UNSET`
- `config.py`: Updated docstring to document `None` behaviour
- `test_config.py`: Added 4 new unit tests covering `None` timeout at init and via `precedence_overwritten`

## Testing

All 21 config unit tests pass. No new lint errors introduced.


packages/railtracks/src/railtracks/utils/config.py
```diff
@@ -5,12 +5,14 @@
...
 
+_UNSET = object()
+
 
 class ExecutorConfig:
     def __init__(
         self,
         *,
-        timeout: float = 150.0,
+        timeout: float | None = 150.0,
```
Amir-R25: This should default to `None` instead of `150.0`

Amir-R25 (changes requested):
Thanks for the quick turnaround on this Guan

There is no need for sentinel parameter as `not provided == disable` from a developer standpoint.

Summary of developer experience:
`timeout=None` => Disable
`timeout` not provided => Disable
`timeout=some number` => Enable

Issue #1027, opened 2026-03-19
[Bug] `AnthropicLLM` Bug


Amir-R25:
### Description

We seem to be missing a parameter for Anthropic llm messages

```
Traceback (most recent call last):
  File "/Users/amirr/dev/demo-agents/.venv/lib/python3.10/site-packages/railtracks/execution/execution_strategy.py", line 52, in execute
    result = await invoke_func()
  File "/Users/amirr/dev/demo-agents/.venv/lib/python3.10/site-packages/railtracks/execution/task.py", line 36, in invoke
    return await self.node.tracked_invoke()
  File "/Users/amirr/dev/demo-agents/.venv/lib/python3.10/site-packages/railtracks/nodes/nodes.py", line 149, in tracked_invoke
    raise e
  File "/Users/amirr/dev/demo-agents/.venv/lib/python3.10/site-packages/railtracks/nodes/nodes.py", line 147, in tracked_invoke
    return await self.invoke()
  File "/Users/amirr/dev/demo-agents/.venv/lib/python3.10/site-packages/railtracks/nodes/nodes.py", line 81, in async_wrapper
    return await method(self, *args, **kwargs)
  File "/Users/amirr/dev/demo-agents/.venv/lib/python3.10/site-packages/railtracks/built_nodes/concrete/_tool_call_base.py", line 317, in invoke
    still_tool_calls, message = await self._handle_tool_calls()
  File "/Users/amirr/dev/demo-agents/.venv/lib/python3.10/site-packages/railtracks/built_nodes/concrete/_tool_call_base.py", line 298, in _handle_tool_calls
    message = await self._on_max_tool_calls_exceeded()
  File "/Users/amirr/dev/demo-agents/.venv/lib/python3.10/site-packages/railtracks/built_nodes/concrete/_tool_call_base.py", line 265, in _on_max_tool_calls_exceeded
    response = await asyncio.to_thread(
  File "/Users/amirr/.pyenv/versions/3.10.18/lib/python3.10/asyncio/threads.py", line 25, in to_thread
    return await loop.run_in_executor(None, func_call)
  File "/Users/amirr/.pyenv/versions/3.10.18/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/amirr/dev/demo-agents/.venv/lib/python3.10/site-packages/railtracks/llm/model.py", line 182, in chat
    self._run_exception_hooks(messages, e)
[...]

PR #1029, opened 2026-03-19
enhance handling of raw litellm messages to preserve provider-specific metadata

Amir-R25:
- Fixing the handling of `raw_litellm_message`
- Adding `litellm.modify_params = True` to solve #1027

### DIFF
--- a/packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
+++ b/packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
@@ -39,6 +39,7 @@ _TBaseModel = TypeVar("_TBaseModel", bound=BaseModel)
 
 # Dropped unsupported parameters from the request to the model.
 litellm.drop_params = True
+litellm.modify_params = True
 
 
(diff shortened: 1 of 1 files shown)


packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
```diff
@@ -727,6 +722,29 @@ def _to_litellm_message(self, msg: Message) -> Dict[str, Any]:
                 )
                 for tool_call in msg.content
             ]
+            # Copy provider-specific metadata (e.g. Gemini thought_signature)
+            # from the raw litellm message without returning it wholesale,
+            # since msg.content may have been truncated and returning the raw
+            # message would re-introduce tool_call_ids that lack responses.
+            raw = getattr(msg, "raw_litellm_message", None)
+            if raw is not None:
+                _standard_fields = {
```
soulFood5632, line 731: this is such a code smell, but I think we solve it later.

Amir-R25: yes I hate it

PR #1056, opened 2026-04-07
Update mkdocs dependency version constraint to prevent breaking changes

Amir-R25:
Following: https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/

While the last release of mkdocs for version 1.X is from Aug 30, 2024, this PR pins our version with the hopes of an update to 1.X and also pins it so we won't automatically update to the incompatible 2.X.


jbueza-railtownai (dismissed):
LGTM!

image

PR #1064, opened 2026-04-13
added [visual] package and fixed import errors

### LINKED ISSUE #1061: [Tech Debt] Does `pip install 'railtracks[cli]'` still make sense?

@Pooria90 discovered a bug where we need FastAPI installed, while we can fix that bug and refactor, the visualizer is a big feature of our framework. Does it still make sense to have a version of our package that doesn't come with it?

### Additional Context

_No response_

Pooria90:
## What does this add?

Closes: https://github.com/RailtownAI/railtracks/issues/1061

Fixes `import railtracks` and the CLI entry point crashing on a base install (no extras) due to FastAPI/uvicorn being imported eagerly at module load time. After this change:

- `pip install railtracks` followed by `import railtracks` works without error.
- CLI commands that don't need a web server (`init`, `update`, `add`) work without installing visual extras.
- `railtracks viz` and `ChatUI` gracefully tell the user to `pip install 'railtracks[visual]'` when FastAPI is missing, instead of showing a raw traceback.

The fix uses Python's module-level `__getattr__` (PEP 562) to defer heavy imports until the code path that actually needs them runs. The monolithic `cli/__init__.py` is split so the FastAPI app and server live in a dedicated `viz_server.py` submodule. The optional extras are renamed from `[cli]`/`[chat]` to `[visual]` (with backward-compatible aliases).

## Type of changes

## Background context

The top-level `railtracks/__init__.py` eagerly imported `interaction.interactive`, which pulled in `human_in_the_loop` → `local_chat_ui` → `FastAPI`. Since FastAPI is an optional dependency (not in core `[project.dependencies]`), any user who ran `pip install railtracks` without extras would hit `ModuleNotFoundError: fastapi` on import. The same problem affected the CLI entrypoint: `railtracks.cli` imported FastAPI and uvicorn at the top of the file, so every subcommand (`init`, `update`, `add`) crashed even though none of them need a web server.

### Code Quality

### Testing

### Documentation

### Git & PR Management

### Breaking Changes

---

## Final Product

**Base install (no extras):**

```
$ pip install railtracks
$ python -c "import railtracks; print(railtracks.__version__)"
1.0.0

$ railtracks add cursor:agent-builder
[railtracks] Installed 'agent-builder' for Cursor → .cursor/rules/agent-builder.mdc

$ railtracks viz
[railtracks] The visualizer requires optional dependencies.
[railtracks] Install with: pip install 'railtracks[visual]'
[railtracks] (or: pip install 'railtracks[cli]' — same dependencies, backward compatible)
```

**Accessing ChatUI without extras:**

```python
>>> from railtracks.human_in_the_loop import ChatUI
railtracks.visual_extra.VisualExtraRequired: The local chat UI requires optional
dependencies. Install with: pip install 'railtracks[visual]'
(or 'railtracks[cli]' for backward compatibility).
```

**With visual extras installed, everything works as before:**

```
$ pip install 'railtracks[visual]'
$ python -c "from railtracks.human_in_the_loop import ChatUI; print(ChatUI)"
<class 'railtracks.human_in_the_loop.local_chat_ui.ChatUI'>
```

## Additional Notes

- **Backward compatibility:** `pip install 'railtracks[cli]'` and `pip install 'railtracks[chat]'` still work; they resolve to the same `[visual]` dependencies. No user action is required unless they want to adopt the new name.
- **Transitive dependency note:** `mcp>=1.9.0` (a core dependency) transitively pulls in `uvicorn` and `starlette`. The visual dependency guard checks for **both** `fastapi` and `uvicorn`, so this doesn't cause false positives. If `mcp` ever drops `uvicorn`, the guard will still work correctly; it will just catch both missing packages instead of only `fastapi`.
- **PEP 562 pattern:** The lazy import mechanism uses module-level `__getattr__`, standardized in Python 3.7. This is the same pattern used by NumPy, SciPy, and other major libraries for optional/heavy imports.


packages/railtracks/src/railtracks/human_in_the_loop/__init__.py
```diff
@@ -1,4 +1,26 @@
...
+
+            raise VisualExtraRequiredError(
+                "The local chat UI requires optional dependencies. "
+                "Install with: pip install 'railtracks[visual]' "
+                "(or 'railtracks[cli]' for backward compatibility)."
+            ) from e
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+def __dir__() -> list[str]:
```
Amir-R25, line 29: what is this used for?

packages/railtracks/src/railtracks/interaction/interactive.py
```diff
@@ -34,27 +38,15 @@ def _process_attachment(attachments: list[UserMessageAttachment]) -> list[str]:
...
-    chat_ui: ChatUI,
+    chat_ui,
     node: Type[LLMBase[_TOutput, _TOutput, Literal[False]]],
     initial_message_to_user: str | None,
     initial_message_to_agent: str | None,
     turns: int | None,
     *args,
     **kwargs,
 ) -> _TOutput:
-    """Handles the interactive session logic using the ChatUI interface.
```
Amir-R25, line 45: why were the docstrings removed?

README.md
```diff
@@ -125,8 +125,10 @@ See the [Observability documentation](https://railtownai.github.io/railtracks/ob
 <summary><b>Installation</b></summary>
 
 ```bash
-pip install railtracks 'railtracks[cli]'
+pip install railtracks 'railtracks[visual]'
 ```
+
+(`railtracks[cli]` is an alias for the same optional dependencies, for backward compatibility.)
```
Amir-R25: let's remove this line

packages/railtracks/pyproject.toml
```diff
@@ -11,6 +11,7 @@ authors = [
     { name = "Amir Refaee", email = "amir@railtown.ai" },
     { name = "Aryan Ballani", email = "aryan@railtown.ai" },
     { name = "Tristan Brown", email = "tristan@railtown.ai" },
+    { name = "Pooria Ashrafian", email = "pooria@railtown.ai" }
```
Amir-R25, line 14: :D

packages/railtracks/src/railtracks/visual_extra.py
```diff
@@ -0,0 +1,16 @@
...
+#   Copyright (c) Railtown AI. All rights reserved.
+#   Licensed under the MIT License. See LICENSE in project root for information.
+#   -------------------------------------------------------------
+
+
+class VisualExtraRequiredError(ImportError):
+    """Raised when FastAPI/uvicorn-backed optional components are used without installing extras."""
+
+    def __init__(self, message: str | None = None) -> None:
+        default = (
```
Amir-R25: this str is repeated multiple times. Let's define it somewhere and import it so we can remove the "backward compatibility" easier or update it in a singular location

packages/railtracks/src/railtracks/interaction/__init__.py
```diff
@@ -1,11 +1,24 @@
...
-from .interactive import local_chat
 
 __all__ = [
     "call",
     "call_batch",
     "broadcast",
     "local_chat",
 ]
+
+
```
Amir-R25, line 19: this is a neat idea, I just fear we will lose "." suggestion if we implement the API this way

Amir-R25 (changes requested):

packages/railtracks/src/railtracks/human_in_the_loop/__init__.py
Pooria90: We use `__getattr__` to lazy-load ChatUI, so it isn’t in the module’s `__dict__` until accessed. dir(module) would otherwise omit ChatUI. Returning __all__ from __dir__ keeps dir() aligned with our public exports.

packages/railtracks/src/railtracks/interaction/interactive.py
Pooria90: Back!

packages/railtracks/src/railtracks/interaction/__init__.py
Pooria90: All clear with `TYPE_CHECKING`!

Amir-R25 (approved):
Great, LGTM
lgtm-looks-good-to-me

PR #1067, opened 2026-04-15
llm guardrails for tool calling llms

Pooria90:
## What does this add?

`LLMGuardrailsMixin` previously raised `NotImplementedError` when `guardrails=` was passed to any tool-calling agent (`ToolCallLLM`, `StreamingToolCallLLM`, `StructuredToolCallLLM`). This was because `OutputLessToolCallLLM.invoke()` drove its own tool-call loop and never called `_pre_invoke` / `_post_invoke`; the hooks that the mixin overrides to run input and output guards.

This PR wires input and output guardrails through the tool-call loop, making the guardrail semantics identical to those on `TerminalLLM` and `StructuredLLM`:

- **Input guard** fires once, before the first `chat_with_tools` call.
- **Output guard** fires once, after the LLM produces its final text reply (not on intermediate tool-call turns).

Specific changes:

- `OutputLessToolCallLLM._handle_tool_calls` return type extended to `tuple[bool, Message | None, Response | None]` so `invoke()` can pass the final `Response` to `_post_invoke`.
- `OutputLessToolCallLLM.invoke()` calls `_pre_invoke` before the loop and `_post_invoke` after, patching `message_hist[-1]` in-place when a TRANSFORM action modifies the final reply.
- `StreamingOutputLessToolCallLLM.invoke()` wires `_pre_invoke` only (output guardrails on the streaming tool-call path are deferred; see notes below).
- `StructuredToolCallLLM.invoke()` wires `_pre_invoke` only (output guardrails deferred pending a dedicated `StructuredOutputGuard` type).
- Three new guarded classes added to `guarded_llm.py`: `GuardedToolCallLLM`, `GuardedStreamingToolCallLLM`, `GuardedStructuredToolCallLLM`.
- `LLMGuardrailsMixin._guardrail_agent_kind` extended with `"tool_call"` and `"structured_tool_call"` branches.
- `helpers/tool_call_llm.py` and `helpers/structured_tool_call_llm.py` updated to accept `guardrails=` and select the guarded base class.
- `agent.py` (`_build_dynamic_agent`) removes the `NotImplementedError` and routes to the correct guarded/unguarded helper.
- Docs updated: the "Current Guardrail Support" callout in `guardrails/overview.md` now covers tool-calling agents.

## Type of changes

## Background context

Tool-calling agents use a different invocation loop (`_handle_tool_calls`) that bypasses the `_pre_invoke` / `_post_invoke` hooks relied on by `LLMGuardrailsMixin`. As a result, guardrails silently had no effect, and the framework explicitly blocked the combination with a `NotImplementedError`.

**Deferred (follow-up PRs):**
- Output guardrails for `StreamingToolCallLLM`; the streaming generator bypasses `LLMBase._gen_wrapper` where `_post_invoke` normally runs; requires dedicated refactoring.
- Output guardrails for `StructuredToolCallLLM`; the final content, which is a pydantic model, not a string; needs a `StructuredOutputGuard` base type to handle it meaningfully.

### Code Quality

### Testing

### Documentation

### Git & PR Management

### Breaking Changes

---

## Final Product

```python
import railtracks as rt
from railtracks.guardrails import Guard, GuardrailDecision, InputGuard, OutputGuard, LLMGuardrailEvent
from railtracks.llm import AssistantMessage

@rt.function_node
async def get_forecast(city: str):
    """Return the forecast for a given city. Args: city (str)"""
    return f"Cloudy and 25°C in {city}"

class TopicBlocker(InputGuard):
    def __call__(self, event: LLMGuardrailEvent) -> GuardrailDecision:
        if "cupcake" in str(event.messages):
            return GuardrailDecision.block(reason="Off-topic", user_facing_message="Stay on topic.")
        return GuardrailDecision.allow()

class SafetyStamp(OutputGuard):
    def __call__(self, event: LLMGuardrailEvent) -> GuardrailDecision:
        content = event.output_message.content + " [VERIFIED]"
        return GuardrailDecision.transform_output(
            output_message=AssistantMessage(content), reason="Safety stamp applied."
        )

agent = rt.agent_node(
    name="WeatherAgent",
    tool_nodes=[get_forecast],
    guardrails=Guard(input=[TopicBlocker()], output=[SafetyStamp()]),
)

### DIFF
--- a/packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/__init__.py
+++ b/packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/__init__.py
@@ -1,7 +1,28 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .agent import agent_node
+    from .function import function_node
+
 __all__ = [
-    "function_node",
     "agent_node",
+    "function_node",
 ]
 
-from .agent import agent_node
-from .function import function_node
+
+def __dir__() -> list[str]:
+    return __all__
+
+
+def __getattr__(name: str):
+    if name == "agent_node":
+        from .agent import agent_node
+
+        return agent_node
+    if name == "function_node":
+        from .function import function_node
+
+        return function_node
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/docs/guardrails/overview.md
+++ b/docs/guardrails/overview.md
@@ -25,6 +25,12 @@ Guardrails are attached where agents are defined. The main entry point is `agent
 To write a custom rule, subclass `InputGuard` or `OutputGuard` and implement `__call__`. Your implementation receives an `LLMGuardrailEvent` (the messages and, for output guards, the model response) and returns a `GuardrailDecision` with an action: `ALLOW`, `TRANSFORM`, or `BLOCK`.
 
 !!! info "Current Guardrail Support"
-    Railtracks currently supports LLM input and output guardrails for non-tool-calling agents. This covers the `agent_node` path for terminal, streaming terminal, structured, and streaming structured agents. Input rails run before the LLM call; output rails run after. If a guardrail blocks the interaction, Railtracks raises `GuardrailBlockedError` so the outcome stays explicit.
+    Railtracks supports LLM input and output guardrails for all agent types created via `agent_node`, including tool-calling agents (`ToolCallLLM`, `StreamingToolCallLLM`, `StructuredToolCallLLM`). Input rails run once before the first LLM call; output rails run once on the final reply (not on intermediate tool-call turns). If a guardrail blocks the interaction, Railtracks raises `GuardrailBlockedError` so the outcome stays explicit.
+
+    **Limitations:**
+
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/__init__.py [hunk 1]
```diff
@@ -1,7 +1,28 @@
...
+
 __all__ = [
-    "function_node",
     "agent_node",
+    "function_node",
 ]
 
-from .agent import agent_node
-from .function import function_node
+
```
Amir-R25, line 14: I don't think we need to change this file

Amir-R25 (changes requested):
Thanks for the quick fix! 🙏 

One small comment

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/__init__.py
Pooria90: There was a circular import bug after subclassing. I can send you a doc for it.

Amir-R25: Alright, thanks for the info!

Amir-R25 (approved):
Thanks!

PR #1085, opened 2026-05-06
Rag/embedding

Amir-R25:
Embedding Module

Summary: 
- `Embedding` abstract base class:
  - `aembed`: Subclasses need to implement
  - `embed`: Defaults to `asyncio.run(aembed)` if not overwritten by subclasses
  - `astream_batches`: streaming embedding batches of chunks for the upcoming rag pipeline
- `SyncEmbedding`: mixin for providers that do not provide an async embedding api
- `LiteLLMEmbedding`: implements the `Embedding` base class to be used with litellm
   - Point of concern: `LiteLLM` does not directly expose embedding costs, so we are using an internal field of theirs called `_hidden_params` which provides cost. This is potentially problematic as its an internal undocumented param
   - This can be a point for a case that maybe for embedding it's best to just directly implement each providers logic. Current design future proofs if we make this decision as `OpenAIEmbedding` can inherit from `Embedding` directly rather `LiteLLMEmbedding`
- `OpenAIEmbedding`, `AzureEmbedding` and `OllamaEmbedding` child classes extending `LiteLLMEmbedding` for now
- AI Generated documentation (Placeholder for now, will need to edit/verify before merging feature branch to main)
- unit tests
- Current Date Model (please provide feedback)
image

I would say please mainly focus your review on `Embedding` base class and see if it make sense from the following two paradigms:
- Extendible embedding wrapper for advanced users
- Part of an event based ingestion/retrieval pipeline


packages/railtracks/src/railtracks/retrieval/embedding/base.py
```diff
@@ -0,0 +1,123 @@
+from __future__ import annotations
+
+import asyncio
+from typing import TypeVar
+import sys
+from abc import ABC, abstractmethod
+from collections.abc import AsyncGenerator, AsyncIterable
+
+from ...utils.logging.create import get_rt_logger
```
Pooria90: Full import path for this might be better as the structure of this module might change

packages/railtracks/src/railtracks/retrieval/embedding/base.py
```diff
@@ -0,0 +1,123 @@
...
+from .models import (
+    EmbeddingFailure,
+    EmbeddingResult,
+    TextEmbeddings,
+)
+
+logger = get_rt_logger(__name__)
+
+_T = TypeVar("_T")
+async def _to_async_iterable(lst: list[_T]) -> AsyncGenerator[_T, None]:
```
Pooria90, line 21: Image

packages/railtracks/src/railtracks/retrieval/embedding/base.py
Amir-R25: Nice thanks will do!

packages/railtracks/src/railtracks/retrieval/embedding/base.py
```diff
@@ -0,0 +1,123 @@
...
+                raise RuntimeError(
+                    "embed() cannot be called from a Jupyter notebook; "
+                    "use `result = await embedder.aembed(texts)` instead."
+                )
+            raise RuntimeError(
+                "embed() cannot be called from an async context; use `await aembed()` instead."
+            )
+        return asyncio.run(self.aembed(texts))
+
+    async def _embed_one_batch(
```
Pooria90, line 60: sick!

Pooria90 (approved):
Solid design!

PR #1089, opened 2026-05-07
Pooria/stores

Pooria90:
## What does this add?

Adds the **Stores** layer to the RAG pipeline — the persistence and retrieval layer that sits after the embedding stage.

**Core data models**

- `StoreEntry` — the atomic unit written to and retrieved from a store. Required fields map directly from `EmbeddedChunk`; all enrichment fields (`abstract`, `summary`, `scope`, `store_category`, `entities`, `valid_from`, `valid_until`) are optional.
- `StoreEntry.from_chunk(embedded_chunk, ...)` — classmethod that builds an entry from an `EmbeddedChunk` automatically; all field mapping is handled, enrichment fields are keyword-only overrides.
- `StoreScope` — frozen dataclass for hard-filter namespacing. Any non-`None` field becomes a mandatory equality filter on every read and clear — a query scoped to `user_id="alice"` will never return entries written under `user_id="bob"`.
- `StoreCategory` — string enum (`EPISODIC`, `SEMANTIC`, `SKILL`, `PROCEDURAL`) for classifying entry knowledge type.
- `DetailLevel` — enum controlling retrieval verbosity: `L0` returns abstract only, `L1` adds summary, `L2` returns full content (default). Enables cheap first-pass ranking before an expensive full-content pass.
- `StoreQuery` / `RetrievedStoreEntry` — query input and scored result wrappers.

**`Store` protocol**

An async `write` / `read` / `delete` / `clear` interface. Any class satisfying it can be used wherever a `Store` is expected.

**`VectorStore`**

The built-in implementation. Delegates low-level index I/O to a swappable `VectorBackend` while owning payload serialization, scope filter construction, and detail-level projection.

**Three backends**

| Backend | Persistence | Install |
|---|---|---|
| `InMemoryBackend` | Optional JSON snapshot | None |
| `ChromaBackend` | Via client mode (ephemeral / on-disk / HTTP) | `railtracks[stores-chroma]` |
| `PgvectorBackend` | Full Postgres durability | `railtracks[stores-vector]` |

All backends accept a `DistanceMetric` (`COSINE`, `L2`, `IP`). `ChromaBackend` and `PgvectorBackend` expose an async `create(...)` factory that initialises the backend in one call.

**Documentation**

Full docs added under `docs/rag/components/stores/`: an overview page covering the data model, querying, the `Store` protocol, and `VectorStore` usage, plus a backends page covering all three backends and the custom-backend recipe. Wired into `mkdocs.yml`.

**Model rename: `Memory*` → `Store*`**

All public model names were changed from `Memory`-prefixed to `Store`-prefixed (`MemoryEntry` → `StoreEntry`, `MemoryScope` → `StoreScope`, etc.) because "memory" is overloaded in the LLM agent literature (LangMem, MemGPT) and these stores are general-purpose retrieval stores, not agent memory systems.

---

## Type of changes

---

## Background context

The stores layer completes the core RAG pipeline (`ingestion → chunking → embedding → stores`). It was designed backend-agnostic from the start: the `VectorBackend` protocol is the only contract between `VectorStore` and the underlying index, so swapping from `InMemoryBackend` to `PgvectorBackend` requires changing one line.

`StoreScope` was modelled as a hard filter (not a soft hint) to make multi-tenant data isolation a first-class concern. The scope fields are serialised as `scope_*` prefixed keys in the payload so they are always filterable at the backend level, not post-hoc in Python.

`DetailLevel` exists to enable multi-stage retrieval pipelines that are common in production RAG: a cheap `L0` pass over abstracts can narrow a large corpus to a short candidate list before a more expensive `L2` pass fetches full content for re-ranking.

The `Memory*` → `Store*` rename was made early to avoid a harder migration later. The term "memory" is already claimed by agent memory systems (LangMem, MemGPT, semantic/episodic agent memory); using it for a retrieval store creates confusing overlap in docs and user mental models.

---

### Code Quality
- [ ] Code is commented, particularly in hard-to-understand areas

### Testing

### Documentation

### Git & PR Management

### Breaking Changes

---

## Final Product

**Creating and indexing entries from an embedding pipeline:**

```python
from railtracks.retrieval.stores import (
    VectorStore, StoreEntry, StoreScope, StoreCategory,
    StoreQuery, DetailLevel,
)
from railtracks.retrieval.stores.vector.backends import InMemoryBackend

store = VectorStore(InMemoryBackend())
scope = StoreScope(user_id="alice", session_id="s-001")

# Index
for embedded_chunk in embedded_chunks:
    entry = StoreEntry.from_chunk(
        embedded_chunk,
        scope=scope,
        store_category=StoreCategory.SEMANTIC,
        abstract=embedded_chunk.chunk.content[:100],
    )
    await store.write(entry)

# Retrieve with detail-level control
query = StoreQuery(
    text="What is the refund policy?",
    scope=scope,
    embedding=query_vector,
    top_k=5,
    detail_level=DetailLevel.L2,
)
results = await store.read(query)
for r in results:
    print(f"(rank={r.rank}, score={r.score:.3f}) {r.entry.content}")
```

**Production backend (Pgvector):**

```python
from railtracks.retrieval.stores import VectorStore, PgvectorBackend

backend = await PgvectorBackend.create(
    dsn="postgresql://user:pass@localhost/mydb",
    table="my_index",
    dim=1536,
)
store = VectorStore(backend)
```

**Persistent development backend (Chroma on-disk):**

```python
from railtracks.retrieval.stores import VectorStore, ChromaBackend

backend = ChromaBackend("my-collection", path="/data/chroma")
await backend.initialize()
store = VectorStore(backend)
```


packages/railtracks/src/railtracks/retrieval/models.py
```diff
@@ -1,116 +1,63 @@
...
-    Attributes:
-        content: The raw text content.
-        id: Unique identifier, auto-generated as a UUID if not provided.
-        source: Origin of the content — file path, URL, database key, etc.
-        metadata: Arbitrary key-value pairs attached by the loader (page number,
-            language, author, etc.).
-    """
-
     content: str
-    type: DocumentType
```
Amir-R25, line 32: this needs to be reverted to DocumentType

packages/railtracks/src/railtracks/retrieval/models.py
```diff
@@ -1,116 +1,63 @@
...
     chunks: list[RetrievedChunk]
-    total_candidates: int | None = None  # considered before top_k selection
-    metadata: dict[str, Any] = field(default_factory=dict)
+    total_candidates: int | None = None
+    metadata: dict = field(default_factory=dict)
 
 
 @dataclass(frozen=True)
 class Cost:
-    """Declarative budget threaded through retrieval."""
```
Amir-R25, line 112: revert this please

Amir-R25 (approved):
Thanks Pooria, let's merge this in, we can handle required changes in the runtime work

PR #1097, opened 2026-05-11
1094 feature Some Retry Logic for llms

### LINKED ISSUE #1094: [Feature] Some Tooling to handle Rate Limit retries
### Feature Request

Right now when I use `agent_node`, if I ever run a lot of calls to the same model i can go over the Token per minute threshold. This is itself is not that big of a problem in a agentic system. However, this becomes a significant problem since there is no easy way for me to build graceful handling of such an error since the calling of the llm's is so deep within railtracks. 

i suggest we implement retries on the railtracks side of things inside the llm module itself which will allow the user to turn it off if the want to. something like:

```python 
rt.llm.OpenAILLM(retry_logic=(max_num=10, method="exponential"))
```

that way the user is able to control how they want to do retries and we provide support for it as required. 

### Additional Context

_No response_

soulFood5632:
## Summary

I have built some important logic in to handle retries of requests in the railtracks LLM module. I have implemented 3 major types and have allowed the user to create their own if they are feeling spicy. 

These new approaches include:
1. Exponential Retry
2. Linear Retry
3. Fixed Retry

I created a new error type is raised if the rate limit fails the prerequisite number of times. 

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

## Notes

To recreate the problem that would require a rate limit 

```python
import railtracks as rt
from dotenv import load_dotenv

load_dotenv()


openaiLLM = rt.llm.OpenAILLM(
    "gpt-4o", retry_approach=rt.llm.retries.ExponentialRetry(max_tries=5, base=2)
)


@rt.function_node
def big_tool() -> str:
    print("called tool once")
    result = ["la" for _ in range(100)]
    return " ".join(result) + "\n\nNot completed"


system_message = "You are a helpful assistant that calls a tool to generate a big string. You should call the tool 1 time, making sure the result is good then call it again making sure the output is good. After completing the first 2 steps, repeat the call check process until the returned response says that it is done. At which time you will return the results. DO NOT return the response until you receive clear instructions from the tool that is complete"
Agent = rt.agent_node(
    name="test_agent",
    system_message=system_message,
    llm=openaiLLM,
    tool_nodes=[big_tool],
)

flow = rt.Flow("Tester", entry_point=Agent)


if __name__ == "__main__":
    print(flow.invoke("give me the goods"))
```

There are various pieces of future work attached to this:
1. Handling rate limit responses where the entire message is too big rather than it being a rapid calls related problem. (i.e. we can only handle 300000 tokens per minute for gpt-4o. if we have a input larger than that it will fail every time even with the backups). 
2. Being able to pull out the recommended wait time from rate limit requests would be nice. I tried to implement it here but there was a closet of skeletons in the way litellm does things. 
3. Add support for review of this somewhere in the visualizer. 

The above will be issued up.


soulFood5632:
I need to review the testing before it is ready for review. Please hold on until I have

soulFood5632 (commented):
Looks good for review

soulFood5632:
Still needs manual verification that it works for non openai models but I would be surprised if it fails

soulFood5632:
This is ready for review now

docs/documentation/agent_design/llms/handling_failures.md
```diff
@@ -0,0 +1,25 @@
+# Handling LLM Failures
+LLM failures are inevitable. In your agent design you must handle these failures accoridingly Railtracks gives you the tooling to handle them gracefully.
```
Amir-R25: typo `accoridingly`

docs/documentation/agent_design/llms/handling_failures.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Types of Failures
+LLM failures fall into two categories, and the right handling differs between them:
+
+1. **Retryable Failures**: Known, anticipated errors you can recover from. Including, rate limits (`429`), timeouts, transient server errors. Safe to retry with backoff.
```
Amir-R25: no comma needed after Including :D

packages/railtracks/src/railtracks/built_nodes/concrete/_tool_call_base.py
```diff
@@ -248,10 +248,15 @@ async def _handle_tool_calls(
...
         """
-
-        response = await asyncio.to_thread(
-            self.llm_model.chat_with_tools, self.message_hist, tools=self.tools()
-        )
+        try:
+            response = await asyncio.to_thread(
+                self.llm_model.chat_with_tools, self.message_hist, tools=self.tools()
+            )
+        except Exception as e:
```
Amir-R25, line 255: nitpick, I'd use `repr(e)` over `str(e)` since we don't we have stack_trace

packages/railtracks/src/railtracks/built_nodes/concrete/_tool_call_base.py
```diff
@@ -292,10 +297,16 @@ class StreamingOutputLessToolCallLLM(
...
-    async def _handle_tool_calls(self):
-        returned_mess = await asyncio.to_thread(
-            self.llm_model.chat_with_tools, self.message_hist, tools=self.tools()
-        )
+    async def _handle_tool_calls(self):  # noqa: C901
+        try:
+            returned_mess = await asyncio.to_thread(
+                self.llm_model.chat_with_tools, self.message_hist, tools=self.tools()
+            )
+        except Exception as e:
```
Amir-R25, line 305: same as above

packages/railtracks/src/railtracks/built_nodes/concrete/terminal_llm_base.py
```diff
@@ -67,7 +67,7 @@ async def invoke(self):
             raise LLMError(
                 reason=f"Exception during llm model chat: {str(e)}",
```
Amir-R25, line 68: same

packages/railtracks/src/railtracks/built_nodes/concrete/terminal_llm_base.py
```diff
@@ -126,6 +126,6 @@ async def invoke(self):
             raise LLMError(
```
Amir-R25, line 126: same, leave the decision to you

packages/railtracks/src/railtracks/llm/retries/__init__.py
```diff
@@ -0,0 +1,20 @@
+"""Retry strategies for LLM API calls.
```
Amir-R25: we don't generally do __init__.py docstrings, up to you if you want to keep them in or not

Amir-R25 (changes requested):
Nice, minor things

docs/documentation/agent_design/llms/handling_failures.md
Pooria90: Also missing punctuation.

packages/railtracks/src/railtracks/llm/retries/base.py
```diff
@@ -0,0 +1,105 @@
...
+                        self.approach_name(),
+                        "Max retries exceeded",
+                        ["Examine exceptions to determine the root cause."],
+                        exceptions,
+                    ) from e
+
+                delay: float | None = None
+
+                if delay is not None:
+                    time.sleep(delay)
```
Pooria90: The use of sync sleep here sounds a bit worrying to me. I think we should switch to async if we consider doing some background work while this is cools off.

packages/railtracks/src/railtracks/llm/retries/base.py
```diff
@@ -0,0 +1,105 @@
...
+                        self.approach_name(),
+                        "Max retries exceeded",
+                        ["Examine exceptions to determine the root cause."],
+                        exceptions,
+                    ) from e
+
+                delay: float | None = None
+
+                if delay is not None:
+                    await asyncio.sleep(delay)
```
Pooria90, line 96: Hah here it is! ^^

packages/railtracks/src/railtracks/llm/retries/base.py
```diff
@@ -0,0 +1,105 @@
...
+                exceptions.append(e)
+                if attempt == self._max_tries - 1:
+                    raise RetryError(
+                        self.approach_name(),
+                        "Max retries exceeded",
+                        ["Examine exceptions to determine the root cause."],
+                        exceptions,
+                    ) from e
+
+                delay: float | None = None
```
Pooria90: Is there any part of the code changing this value? seems like None all the time.

mkdocs.yml
```diff
@@ -83,8 +83,11 @@ nav:
             - Function Tools: documentation/agent_design/tools/function_tools.md
             - Agents as Tools: documentation/agent_design/tools/agents_as_tools.md
             - MCP: documentation/agent_design/tools/mcp.md
-        - Structured Extraction: documentation/agent_design/structured_extraction.md
-        - Modalities: documentation/agent_design/modalities.md
+        - LLMs:
+            - Modalities: documentation/agent_design/llms/modalities.md
+            - Failures: documentation/agent_design/llms/handling_failures.md
+            - Structured Extraction: documentation/agent_design/llms/structured_extraction.md
```
Pooria90: This is an empty page. Is it a placeholder?

Pooria90 (commented):
Sick!

packages/railtracks/src/railtracks/llm/retries/base.py
soulFood5632: Yea that is very reasonable, but since the operations done on this thread are purely the LLM calls I think it is ok. In essence it isn't blocking anything.

mkdocs.yml
soulFood5632: @Pooria90 I am unsure which lline you are referencing in this comment?

packages/railtracks/src/railtracks/llm/retries/__init__.py
soulFood5632: Yea AI artifact. I'll get rid of it

packages/railtracks/src/railtracks/llm/retries/base.py
soulFood5632: Yep, artifiact of previous implementation

mkdocs.yml
soulFood5632: Yep not sure how that made it in

Amir-R25 (approved):
Lovely

PR #1101, opened 2026-05-19
Add PDF OCR loader (closes #1079)

### LINKED ISSUE #1079: [Feature] [Retrieval][Loader] PDF OCR loader
> [!WARNING]
> Please PR into `feature-branch-rag` and not `main`
### Feature Request
Add an OCR-capable PDF document loader that scans each page and outputs our `Document` data model, with a potential `BaseOCRLoader` abstraction.

#### Problem
PDFs are often a collection of scanned images rather than text-based documents. Standard PDF loaders cannot extract text from these, making them unusable in document pipelines.

#### Proposed Solution
Add a new OCR-capable document loader that processes each PDF page using OCR and returns a `Document` object per page. Since OCR is a common functionality, it may be worth introducing a `BaseOCRLoader` first.

#### Motivation
- Scanned PDFs are extremely common in real-world use cases (invoices, contracts, forms, etc.)
- A `BaseOCRLoader` would make it easy to support other OCR-backed sources in the future (e.g. images, TIFFs)

#### Acceptance Criteria
- [ ] `BaseOCRLoader` is defined with a common interface and extends `BaseDocumentLoader`
- [ ] An OCR-capable PDF loader is implemented extending `BaseOCRLoader`
- [ ] Each scanned page is returned as a `Document` object
- [ ] Loader is tested against both text-based and scanned PDFs

#### Additional Context
- Extend `BaseDocumentLoader`: https://github.com/RailtownAI/railtracks/blob/14f5f78ecf6cac2f6356f5373d071d9771b8369f/packages/railtracks/src/railtracks/retrieval/loaders/base.py#L10-L36
- Review `PyPDFLoader`: https://github.com/RailtownAI/railtracks/blob/14f5f78ecf6cac2f6356f5373d071d9771b8369f/packages/railtracks/src/railtracks/retrieval/loaders/pdf_loader.py#L23-L47
- Review `Document` data model: https://github.com/RailtownAI/railtracks/blob/14f5f78ecf6cac2f6356f5373d071d9771b8369f/packages/railtracks/src/railtracks/retrieval/models.py#L20-L35

DharunKosanam:
## Summary

Adds `BaseOCRLoader` (abstract) and `PyPDFOCRLoader` (concrete) to the
retrieval module. Closes #1079.

`PyPDFOCRLoader` handles both text-based and scanned PDFs in a single loader:
- Tries `pypdf` text extraction per page first
- Pages where extraction returns empty are rasterized with `pypdfium2` and
  recognized via Tesseract (`pytesseract`)
- `force_ocr=True` flag OCRs every page unconditionally
- Metadata records whether each page used OCR (`metadata["ocr"]` for page
  strategy; `metadata["ocr_pages"]` for document strategy)

**Design decisions:**
- **pypdfium2** (Apache-2.0 / BSD-3) over Poppler (GPL-2), PyMuPDF (AGPL),
  and Wand/ImageMagick (transitively AGPL via Ghostscript). It was already
  a transitive dependency via `pdfplumber`.
- **Tesseract** as the default OCR engine — open-source, no API key, the
  obvious default for an OSS library. Subclassing `BaseOCRLoader` with a
  different engine (EasyOCR, Textract, LLM vision) is a one-method override.
- New `[ocr]` optional-dependency extra in `pyproject.toml`. Users also
  need the Tesseract binary on PATH — surfaced in the `ImportError` message.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

**Test coverage:**

- **29 unit tests** (mocked, no real binaries needed in CI):
  - `BaseOCRLoader` contract + inheritance (6 tests)
  - `PyPDFOCRLoader` init / errors / text-fast-path / OCR-fallback /
    `force_ocr` / page strategy / document strategy / directory loading
    (23 tests)
- **End-to-end smoke test** in `packages/railtracks/tests/end_to_end/rag/`,
  excluded from regular CI per `pyproject.toml`'s existing `addopts`.
  Generates a synthetic image-only PDF with Pillow, runs the loader,
  asserts text recovery. Verified locally on Windows with Tesseract
  v5.5.0.20241111.
- **Manual verification** against a real scanned typed document (clean
  recovery) and a handwritten document (expected Tesseract limitation —
  intended to be addressed in future via additional `BaseOCRLoader`
  subclasses for handwriting-capable engines).
- Full loaders test suite: 111 passed, no regressions.

**Installation:** users need to `pip install "railtracks[ocr]"` for the
Python packages, and separately install the Tesseract binary and put it
on PATH. The loader's `ImportError` message points to the official
Tesseract install docs.


packages/railtracks/src/railtracks/retrieval/loaders/pdf_ocr_loader.py
```diff
@@ -0,0 +1,236 @@
...
+                    source=source,
+                    metadata={
+                        "page": page_index + 1,
+                        "total_pages": total_pages,
+                        "file_type": ".pdf",
+                        "ocr": used_ocr,
+                    },
+                )
+        finally:
+            await asyncio.to_thread(pdf.close)
```
Pooria90, line 208: Can you also check and see if `reader` needs to be closed?
It might be a potential resource leak.

packages/railtracks/src/railtracks/retrieval/loaders/pdf_ocr_loader.py
```diff
@@ -0,0 +1,236 @@
...
+                        "total_pages": total_pages,
+                        "file_type": ".pdf",
+                        "ocr_pages": ocr_pages,
+                    },
+                )
+                return
+
+            for page_index in range(total_pages):
+                text, used_ocr = await self._resolve_page_text(reader, pdf, page_index)
+                if not text or not text.strip():
```
Pooria90, line 191: So basically empty and white-space-only pages won't return any document object. 
But an empty page will leave a "\n\n\n\n" trace in the `document` strategy. Can we handle both cases consistently?

uv.lock
```diff
@@ -158,18 +158,6 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/6f/12/e5e0282d673bb9746bacfb6e2dba8719989d3660cdb2ea79aee9a9651afb/anyio-4.10.0-py3-none-any.whl", hash = "sha256:60e474ac86736bbfd6f210f7a61218939c318f43f9972497381f1c5e930ed3d1", size = 107213, upload-time = "2025-08-04T08:54:24.882Z" },
 ]
```
Pooria90, line 160: Please reset this file to the original one; it will change every time users sync their dependencies.

Pooria90 (changes requested):
Nice work! Thanks for the PR @DharunKosanam.

I have a few comments on the code.

Also, please add some documentations on installation and usage inside `docs/rag/components/ingestion/pdfs.md` to complete the work.

packages/railtracks/src/railtracks/retrieval/loaders/__init__.py
```diff
@@ -1,11 +1,13 @@
 from railtracks.retrieval.loaders.base import BaseDocumentLoader
+from railtracks.retrieval.loaders.base_ocr import BaseOCRLoader
 from railtracks.retrieval.loaders.csv_loader import CSVLoader
 from railtracks.retrieval.loaders.json_loader import JSONLoader
 from railtracks.retrieval.loaders.text_loader import TextLoader
 from railtracks.retrieval.models import DocumentType
 
 __all__ = [
     "BaseDocumentLoader",
+    "BaseOCRLoader",
```
CoronRing: Perhaps this should be PyPDFOCRLoader? Since BaseOCRLoader is just the template.

packages/railtracks/src/railtracks/retrieval/loaders/pdf_ocr_loader.py
```diff
@@ -0,0 +1,236 @@
...
+
+from railtracks.retrieval.loaders.base_ocr import BaseOCRLoader
+from railtracks.retrieval.models import Document, DocumentType
+
+try:
+    import pypdfium2 as pdfium
+    import pytesseract
+    from pypdf import PdfReader
+except ImportError as exc:
+    raise ImportError(
```
CoronRing: This will throw on instantiation/compile instead of import, which means any use, even not [ocr] will trigger

packages/railtracks/src/railtracks/retrieval/loaders/pdf_ocr_loader.py
DharunKosanam: fixed. Added reader.close() in the finally block.

packages/railtracks/src/railtracks/retrieval/loaders/pdf_ocr_loader.py
DharunKosanam: Fixed. Empty/whitespace pages are now skipped in document strategy too, matching page strategy. Added a regression test for it.

uv.lock
DharunKosanam: Done, reset uv.lock to match the base branch.

Pooria90 (approved):

soulFood5632:
Awesome work @DharunKosanam, thank you for the contribution :).

Amir-R25 (approved):
Very cool, thanks @DharunKosanam

PR #1102, opened 2026-05-19
feat: add remote storage loaders and writers (S3, GCS, Azure Blob, SQL)

### LINKED ISSUE #1090: [Feature] [Retrieval] Loader for remote storage provider
## **Background / Context**
Currently, **Railtracks** needs a robust way to ingest documents from remote cloud storage. Similar to LangChain's Document Loaders ecosystem, we need an abstraction layer that handles cloud authentication, fetches files (or entire directories/containers), and standardizes the output into a common format (e.g., `RailtracksDocument` objects containing `content` and `metadata`). 

This will allow our agents to seamlessly access enterprise knowledge bases hosted in AWS and Azure.

### **Requirements & Acceptance Criteria**

**1. AWS S3 Integration**
**2. Azure Blob Storage Integration**

### **Out of Scope (For Future Tickets)**
* Complex file parsing (e.g., extracting tables from PDFs, OCR for images). *This ticket assumes we are fetching plain text/UTF-8 files or returning raw bytes for a downstream parser to handle.*
* Google Cloud Storage (GCS) integration (will be handled in a separate ticket).

### Additional Context

_No response_

CoronRing:
## Summary

- Adds `S3Loader` /  `GCSLoader` / `AzureBlobLoader` / `SQLLoader` 
- All providers use optional extras (`railtracks[aws]`, `railtracks[gcp]`, `railtracks[azure-blob]`, `railtracks[sql]`) so the core package stays lean
- All loaders expose sync and async interfaces (`load`/`aload`)
- SQL classes include a context-manager (`with SQLLoader(...) as l`) and explicit `close()` for engine lifecycle management
- SQL identifier arguments validated against a strict allowlist at construction time to prevent injection
- Full unit test coverage across all classes
- Comprehensive developer docs
- Writer are not included in the PR, it will be in a separate module

## Security hardening

- SQL table/column names validated at `__init__` time — raises `ValueError` on any metacharacter (`[A-Za-z_][A-Za-z0-9_$]*` allowlist, supports `schema.table`)
- Helpful `ValueError` when `content_column` is missing from query results (was a bare `KeyError`)
- `__repr__` on all classes exposes only non-sensitive fields (bucket/container name); credentials never appear in repr
- All `ImportError` messages include both `pip install` and `uv add` forms

## Limitations documented

- CTE (`WITH …`) queries not supported as `table_or_query`; workaround shown in docs
- `aload` are thread-backed (`asyncio.to_thread`) not true-async; noted in docs with guidance for high-concurrency cases

## Test plan

- [x] 127 unit tests passing across all 4 providers × loader + writer
- [x] SQL tests use real in-memory SQLite (no mocks for correctness)
- [x] Cloud tests (S3/GCS/Azure) use provider SDK mocks
- [x] Async variants covered for all classes

## Additional file
Tutorial file are added to ipynb file on colab, not included per requirements for single-location consistency.


uv.lock
Pooria90, line 1: 1. Please reset this one to the one on main.
2. Add uv.lock to the .gitignore so we don't push its changes.

pyproject.toml
```diff
@@ -5,15 +5,18 @@ version = "0.0.0"
 description = "Root workspace for dev tools and workspace packages"
 readme = "README.md"
 requires-python = ">=3.10"
-dependencies = ["railtracks"]
+dependencies = [
```
Pooria90, line 8: I think moto[s3] needs to go the optional dependencies in the toml file inside railtracks.

.gitignore
```diff
@@ -45,4 +45,10 @@ core-licenses-summary.md
 .DS_Store
 
 # Ignore Files copied during release workflow for PyPI packaging (for Flit)
-packages/railtracks/docs/
\ No newline at end of file
+packages/railtracks/docs/
```
Pooria90: These look like local temp folders. Please remove them.

CoronRing: thanks for noticing!

pyproject.toml
CoronRing: REmoved!

uv.lock
CoronRing: nv.lock is removed!

Pooria90 (changes requested):
Good work @CoronRing 

I went over the code. A few substantial changes needs to be applied:
1. We need to move `loaders` and `writers` modules to `packages/railtracks/src/railtracks/retrieval/loaders` as they are being built on top of Amir's work on loaders. I recommend having all classes for a specific provider under one single module: for example having `loaders/s3.py` and `writers/s3.py` all under one `s3.py`.
Or we can also have them under a `loaders/cloud` folder. We can discuss the structure.
2. The modules are returning `Chunk` which is located in our old `vector_stores` module. That module is deprecated now and will be removed from the framework. The new type that we use for our retrieval module is called `Document` which is located in `packages/railtracks/src/railtracks/retrieval/models.py`. Please refer to Amir's loaders for examples.
3. Please adjust the Base classes, data models, and the docs accordingly.

packages/railtracks/src/railtracks/integrations/writers/storage_writers.py
Pooria90, line 1: We already have these scripts under `docs/scripts`. Also, I don't see any references for Writer classes. So, either we should move or remove.

CoronRing: This class is removed entirely,it should be gone

packages/railtracks/src/railtracks/integrations/writers/storage_writers.py
```diff
@@ -0,0 +1,311 @@
...
+
+writer = S3Writer("my-bucket", region_name="us-east-1")
+
+# Write raw text at an explicit key
+uri = writer.write_key("reports/summary.txt", "Today's executive summary ...")
+print(uri)  # s3://my-bucket/reports/summary.txt
+
+# Write a list of Document objects -- key is derived from document.source / id
+uris = writer.write(documents, prefix="generated/")
+# --8<-- [end:s3_write_basic]
```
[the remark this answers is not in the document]
CoronRing: Removed component

packages/railtracks/src/railtracks/integrations/writers/storage_writers.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+# ===========================================================================
+# Azure Blob Storage
+# ===========================================================================
+
+# --8<-- [start:azure_write_basic]
+from railtracks.integrations.writers import AzureBlobWriter
+
+# DefaultAzureCredential resolves credentials automatically
```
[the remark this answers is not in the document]
CoronRing: Writer modules are removed

packages/railtracks/src/railtracks/retrieval/loaders/cloud/azure_blob.py
```diff
@@ -0,0 +1,127 @@
...
+            f"AzureBlobLoader("
+            f"account_url={self._account_url!r}, "
+            f"container_name={self._container_name!r})"
+        )
+
+    def _list_keys(self) -> list[str]:
+        kwargs: dict[str, Any] = {}
+        if self._prefix is not None:
+            kwargs["name_starts_with"] = self._prefix
+        return [blob.name for blob in self._container_client.list_blobs(**kwargs)]
```
Pooria90, line 98: Azure supports async listing. Please check out this: https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blobs-list-python#list-blobs-asynchronously

packages/railtracks/src/railtracks/retrieval/loaders/cloud/azure_blob.py
```diff
@@ -0,0 +1,127 @@
...
+    async def astream(self) -> AsyncGenerator[Document, None]:
+        """Stream documents one at a time as each blob is downloaded.
+
+        Yields:
+            Document: The next loaded document.
+        """
+        if self._keys is not None:
+            names = self._keys
+        else:
+            names = await asyncio.to_thread(self._list_keys)
```
Pooria90, line 124: This is kind of against the aync generator idea. If we implement aync list_keys, we can do this better.

packages/railtracks/src/railtracks/retrieval/loaders/__init__.py
```diff
@@ -1,13 +1,23 @@
 from railtracks.retrieval.loaders.base import BaseDocumentLoader
+from railtracks.retrieval.loaders.cloud import (
+    AzureBlobLoader,
+    GCSLoader,
+    S3Loader,
+    SQLLoader,
+)
```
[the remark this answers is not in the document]
CoronRing: This is updated, the PR message was for old code, its now under retrievals.loaders

packages/railtracks/src/railtracks/retrieval/loaders/cloud/azure_blob.py
CoronRing: This was noted by copilot as well, the issue is that azure_blob for python do not support its usage like this out of the box (stream getting the key). I need to look into if there is an alternative way

packages/railtracks/src/railtracks/retrieval/loaders/cloud/s3.py
```diff
@@ -0,0 +1,131 @@
...
+            aws_access_key_id=aws_access_key_id,
+            aws_secret_access_key=aws_secret_access_key,
+            aws_session_token=aws_session_token,
+            endpoint_url=endpoint_url,
+        )
+
+    def __repr__(self) -> str:
+        return f"S3Loader(bucket={self._bucket!r})"
+
+    def _list_keys(self) -> list[str]:
```
Pooria90, line 94: same comment as Azure; `aiobotocore` supports true async operations.
Check out this link for examples: https://pypi.org/project/aiobotocore/

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
```diff
@@ -0,0 +1,234 @@
...
+    ) -> None:
+        try:
+            import sqlalchemy  # noqa: F401
+        except ImportError:
+            raise ImportError(
+                "sqlalchemy is required for SQL loading. "
+                "Install it via `pip install railtracks[sql]` or `uv add railtracks[sql]`."
+            )
+
+        import sqlalchemy as sa
```
Pooria90: redundant; you can do it in the `try` block

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
```diff
@@ -0,0 +1,234 @@
...
+            stmt = sa.text(self._table_or_query)
+        else:
+            stmt = sa.text(f"SELECT * FROM {self._table_or_query}")  # noqa: S608
+
+        with self._engine.connect() as conn:
+            result = conn.execute(stmt)
+            return [dict(row._mapping) for row in result]
+
+    def _fetch_by_ids(self, keys: list[str]) -> list[dict[str, Any]]:
+        import sqlalchemy as sa
```
Pooria90, line 200: Does this have to be here in addition to the one in `__init__`?

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
```diff
@@ -0,0 +1,234 @@
...
+            return [dict(row._mapping) for row in result]
+
+    def _fetch_rows(self) -> list[dict[str, Any]]:
+        if self._keys is not None:
+            if not self._keys:
+                return []
+            return self._fetch_by_ids(self._keys)
+        return self._fetch_all()
+
+    async def astream(self) -> AsyncGenerator[Document, None]:
```
Pooria90, line 225: same comment as s3 and azure: sqalchemy support async operations!

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def close(self) -> None:
+        """Dispose the underlying SQLAlchemy engine.
+
+        Only disposes engines created internally by this loader.
+        """
+        if self._owns_engine:
+            self._engine.dispose()
+
+    def __enter__(self) -> SQLLoader:
```
Pooria90, line 147: please add `__aenter__` and `__aexit__` for async context management.

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
```diff
@@ -0,0 +1,234 @@
...
+            available = sorted(row.keys())
+            raise ValueError(
+                f"content_column {self._content_column!r} was not found in the "
+                f"query results. Available columns: {available}. "
+                "Check that 'content_column' matches a column returned by the query."
+            )
+        content = str(row[self._content_column])
+
+        if self._metadata_columns is not None:
+            meta = {col: row[col] for col in self._metadata_columns if col in row}
```
Pooria90, line 164: This part is a little bit worrying to me. Let's say the user asked for a col that is not in row. They end of with a data with a missing meta data column, which they were not notified about. Maybe some warning would be nice here, but I'll leave the handling to you.

packages/railtracks/tests/unit_tests/integrations/writers/__init__.py
Pooria90, line 1: i think should be removed!

Pooria90 (changes requested):

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
CoronRing: Moved inside t-c

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
CoronRing: I think the SQLAlchemy engine is synchronous, and connections are opened and closed within each _fetch_* call already.

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
CoronRing: This one is harder, since if a col is missing, it can be either at a col level or row level, depending on the architecture and where it's loaded. If we detect this at each row, it will be expensive and require two iterations. I added a check at the table level, which solves 50% of the issue.

packages/railtracks/tests/unit_tests/integrations/writers/__init__.py
CoronRing: Nice catch, removed

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
CoronRing: Let me know if you like the fix, if so you can resolve it.

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
CoronRing: This was not done because I want to make the sql system generic, the async system is different across providers (largely on auth): Ie
asyncpg for PostgreSQL
aiosqlite for SQLite, 
boto, azure etc.

Which works, but is a lot of work to make it work, which seems like a tradeoff not needed at this time. Since sql was not a primary target anyways.

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
CoronRing: If I do that the sa need to be passed around as a variable, claude said both operations are fine and I guess this one is more debugger friendly and easy to read so I added to each.

packages/railtracks/src/railtracks/retrieval/loaders/cloud/s3.py
CoronRing: All operations are actually already async for azure and s3 in content, its only the key fetch are sync, the reason why list was not streamed was isolation of responsiblity, if I fetch and stream in the same function, we cannot track progress (ie percent loaded) as well as multi-thread properly (ie if I subthread 3 times, each takes 1/3 of task I need to compute full list ahead of time. This logic was used in write and not read, but I think it need to be added later so lets keep it if not too concering?)

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
```diff
@@ -0,0 +1,234 @@
...
+
+    async def astream(self) -> AsyncGenerator[Document, None]:
+        """Stream documents one at a time as rows are read.
+
+        Yields:
+            Document: The next row as a document.
+        """
+        rows = await asyncio.to_thread(self._fetch_rows)
+        for row in rows:
+            yield self._row_to_document(row)
```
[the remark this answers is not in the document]
CoronRing: This was not done because I want to make the sql system generic, the async system is different across providers (largely on auth): Ie
asyncpg for PostgreSQL
aiosqlite for SQLite,
boto, azure etc.

Which works, but is a lot of work to make it work, which seems like a tradeoff not needed at this time. Since sql was not a primary target anyways.

@copilot is this reasonable?

packages/railtracks/src/railtracks/retrieval/loaders/cloud/azure_blob.py
```diff
@@ -0,0 +1,127 @@
...
+
+        for name in names:
+            yield await asyncio.to_thread(self._fetch_document, name)
```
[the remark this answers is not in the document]
CoronRing: All operations are actually already async for azure and s3 in content, its only the key fetch are sync, the reason why list was not streamed was isolation of responsiblity, if I fetch and stream in the same function, we cannot track progress (ie percent loaded) as well as multi-thread properly (ie if I subthread 3 times, each takes 1/3 of task I need to compute full list ahead of time. This logic was used in write and not read, but I think it need to be added later so lets keep it if not too concering?)

@copilot is this reasonable?

packages/railtracks/src/railtracks/retrieval/loaders/cloud/gcs.py
```diff
@@ -0,0 +1,108 @@
...
+        Yields:
+            Document: The next loaded document.
+        """
+        if self._keys is not None:
+            names = self._keys
+        else:
+            names = await asyncio.to_thread(self._list_keys)
+
+        for name in names:
+            yield await asyncio.to_thread(self._fetch_document, name)
```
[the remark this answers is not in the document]
CoronRing: All operations are actually already async for azure and s3 in content, its only the key fetch are sync, the reason why list was not streamed was isolation of responsiblity, if I fetch and stream in the same function, we cannot track progress (ie percent loaded) as well as multi-thread properly (ie if I subthread 3 times, each takes 1/3 of task I need to compute full list ahead of time. This logic was used in write and not read, but I think it need to be added later so lets keep it if not too concering?)

https://gh.io/copilot-coding-agent-docs is this reasonable?

packages/railtracks/src/railtracks/retrieval/loaders/cloud/s3.py
```diff
@@ -0,0 +1,131 @@
...
+        Yields:
+            Document: The next loaded document.
+        """
+        if self._keys is not None:
+            keys = self._keys
+        else:
+            keys = await asyncio.to_thread(self._list_keys)
+
+        for key in keys:
+            yield await asyncio.to_thread(self._fetch_document, key)
```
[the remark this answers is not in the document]
CoronRing: All operations are actually already async for azure and s3 in content, its only the key fetch are sync, the reason why list was not streamed was isolation of responsiblity, if I fetch and stream in the same function, we cannot track progress (ie percent loaded) as well as multi-thread properly (ie if I subthread 3 times, each takes 1/3 of task I need to compute full list ahead of time. This logic was used in write and not read, but I think it need to be added later so lets keep it if not too concering?)

@copilot  is this reasonable?

Pooria90 (approved):

CoronRing (commented):
?

packages/railtracks/src/railtracks/retrieval/loaders/cloud/sql.py
CoronRing: test

CoronRing:
closes #1090

PR #1104, opened 2026-05-19
fix: auto-download UI in viz when bundle is missing

### LINKED ISSUE #1103: bug: railtracks viz serves {"error":"File not found"} with no indication that init is required
## Problem

Running `railtracks viz` without first running `railtracks init` starts the server successfully (port 3030, no terminal errors, browser opens automatically), but every browser request returns:

```json
{"error":"File not found"}
```

There is **no error message** in the terminal indicating that the UI package is missing or that `railtracks init` needs to be run first. The server logs look completely healthy, so the user has no way of knowing what went wrong.

## Root cause

`railtracks viz` calls `create_railtracks_dir()` (creates `.railtracks/`) but never downloads the UI bundle. The server's catch-all route in `viz_server.py` silently falls through to a 404 JSON response when `.railtracks/ui/index.html` does not exist.

## Steps to reproduce

```bash
# Fresh clone, or after deleting .railtracks/
railtracks viz
# → browser opens to http://localhost:3030
# → shows {"error":"File not found"} — no hint in the terminal
```

## Expected behaviour

Either:
- `railtracks viz` should auto-download the UI if missing (no extra step needed), or
- a clear error message should be shown: _"UI bundle not found — run `railtracks init` first"_

The current silent failure is especially confusing because the server reports itself as running successfully.

## Fix

Check for `index.html` in `viz` before starting the server and auto-download if missing — making `railtracks init` optional rather than an undocumented prerequisite.

CoronRing:
## Summary

Fixes #1103

- `railtracks viz` now checks for `.railtracks/ui/index.html` before starting the server and automatically downloads the UI bundle if it is missing, making `railtracks init` an optional step rather than a silent prerequisite
- Switched the path-validity check from `FileNotFoundError` to `OSError` so the invalid-filename fallback actually fires on Windows (where bad chars raise `OSError` errno 22, not `FileNotFoundError`)

## Test plan

- [ ] Delete `.railtracks/ui/` and run `railtracks viz` — bundle should download automatically, browser should load the UI
- [ ] Run `railtracks viz` with UI already present — should start immediately with no download
- [ ] Run `railtracks init` then `railtracks viz` — still works as before


packages/railtracks/src/railtracks/cli/__init__.py
```diff
@@ -465,6 +465,11 @@ def main():
 
         create_railtracks_dir()
 
+        ui_index = Path(f"{cli_directory}/ui/index.html")
```
Pooria90: Let's wait until #1110 is merged, then change this line to 
```python
ui_index = resolve_railtracks_home() / "ui" / "index.html"
```
As now we rely on absolute path generate by the `paths/resolve_railtracks_home` function.

Pooria90 (changes requested):

packages/railtracks/src/railtracks/cli/__init__.py
CoronRing: Updated per-request, please check!

packages/railtracks/src/railtracks/cli/__init__.py
```diff
@@ -468,6 +468,11 @@ def main():
 
         create_railtracks_dir()
 
+        ui_index = resolve_railtracks_home() / "ui" / "index.html"
+        if not ui_index.exists():
+            print_status("UI not found — downloading...")
+            download_and_extract_ui()
```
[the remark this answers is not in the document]
CoronRing: Test was added

Pooria90 (approved):

PR #1106, opened 2026-05-19
fix: prevent empty session files when payload serialization fails

### LINKED ISSUE #1105: bug: session files written as 0 bytes when payload serialization fails
## Problem

Session JSON files in `.railtracks/data/sessions/` are created as 0-byte empty files whenever an error occurs during payload serialization. These empty files persist permanently and cause the viz server to log errors on every startup:

```
[railtracks] Error reading session file _xyz.json: Expecting value: line 1 column 1 (char 0)
```

The viz `/api/sessions` endpoint silently skips all corrupt files, so the UI shows far fewer sessions than expected with no indication of the problem.

## Root cause

In `_session.py`, `__exit__` does this:

```python
file_path.touch()          # 1. empty file created on disk
...
file_path.write_text(...)  # 2. raises (e.g. AssertionError in graph_serialization)
except Exception as e:
    logger.error(...)      # 3. error swallowed — empty file stays forever
```

`file_path.touch()` creates the file unconditionally before the content is ready. If anything between `touch()` and `write_text()` raises an exception, the outer `except Exception` catches it, logs a minimal message, and exits — leaving a permanent empty file behind.

A secondary issue: the filename-fallback `except FileNotFoundError` does not fire on Windows for invalid filename characters, which raise `OSError` (errno 22), not `FileNotFoundError`.

## Impact

- Every failed session write leaves a zombie empty file
- `railtracks viz` logs an error for every empty file on every startup
- The UI silently shows incomplete session history

## Fix

Validate the path with `touch()` + immediate `unlink()` so no file is left if the subsequent write fails. Broaden the filename-fallback to catch `OSError`. Improve the error log to include the full traceback so the root cause (e.g. `AssertionError` in `graph_serialization`) is visible.

CoronRing:
## Summary

Fixes #1105

`file_path.touch()` was creating an empty file before `write_text()` was called. If `payload()` or `json.dumps()` raised an exception (e.g. `AssertionError` inside `graph_serialization()`), the outer `except Exception` swallowed it and the empty file remained permanently on disk.

**Changes:**
- Replace `touch()` with `touch()` + `unlink()` to probe path validity without leaving a file behind if the write later fails
- Broaden filename-fallback catch from `FileNotFoundError` → `OSError` to cover Windows invalid-character errors (errno 22)
- Improve error log to pass the exception message and `exc_info=True` so the full traceback is visible

## Test plan

- [ ] Trigger a session where `graph_serialization()` raises — confirm no empty file is created
- [ ] Trigger a normal session — confirm file is written correctly as before
- [ ] Use a flow name with invalid filename characters on Windows — confirm fallback to identifier-only filename fires


Amir-R25 (approved):
Great, this was an annoying bug!

PR #1109, opened 2026-05-20
`RetrievalRuntime`: Streaming pipeline for ingestion and retrieval

Amir-R25:
## Summary

Adds the `RetrievalRuntime` orchestrator and the supporting `Store` / loader
changes needed to drive the full ingest → retrieve flow end-to-end. Also
removes the legacy `railtracks.vector_stores` package now that
`railtracks.retrieval` supersedes it.

```
┌────────┐   ┌─────────┐   ┌──────────┐   ┌────────┐   ┌───────────┐
│ Loader │ → │ Chunker │ → │ Embedder │ → │ Store  │ → │ Retrieval │
└────────┘   └─────────┘   └──────────┘   └────────┘   └───────────┘
     ▲             ▲             ▲             ▲              ▲
     └─────────────┴─────────────┴─────────────┴──────────────┘
                                 │
                       ┌─────────────────────┐
                       │  RetrievalRuntime   │
                       │  (the orchestrator) │
                       └─────────────────────┘
```

### `RetrievalRuntime`

The orchestrator that wires a chunker + embedder + `Store` (+ optional `scope`)
into the ingest/retrieve flow.

* **Loader is passed to `ingest()`, not the constructor**: one runtime captures
  *how* to process (chunker/embedder/store/scope); the loader decides *what*.
  A single runtime can ingest from many sources and re-ingest to update.
* **Streaming + aggregate APIs:** `ingest(loader)` is an async generator yielding
  per-batch events; `ingest_all(loader)` drains it and returns `IngestionStats`.
* **Events:** `BatchIngested` (carries per-batch `EmbeddingMetrics` — tokens, cost,
  latency, vector count), `EmbeddingFailure`, `DocumentFailed`, `DocumentSkipped`.
  `batch_index` is **per-document**, not run-global.
* **Upsert semantics:** before writing the *first* chunk of a document the runtime
  fires `store.delete_where({"document_id": str(doc.id)})` to clear the prior
  version. The delete only runs once a batch succeeds, so a total embedding
  failure preserves the previous version. Writes are per-chunk and **not
  transactional** — a crash mid-write leaves a partial document (recovered on the
  next ingest, see below).
* **Count-aware staleness (skip unchanged docs):** a document is skipped only when
  the store already holds a *complete* copy — matched on `source_path` +
  `content_hash` and the persisted `doc_chunk_count`. A partially-written document
  (fewer chunks than expected after an interrupted run) is re-ingested rather than
  left broken. Counting is done via `find()` rather than a `count()` call so the
  runtime depends only on the `Store` protocol. 
* **Token-size guard:** when `max_tokens` is set, chunks over the per-item limit
  are dropped *before* embedding and surfaced as `EmbeddingFailure` instead of
  causing provider 4xx errors. Uses `TiktokenTokenizer` by default. (Partial fix
  for the embedding per-item token-cap gap — see Known limitations.)
* **Embedding-model consistency:** the model is captured from the first successful
  batch; a later `retrieve()` with a different embedder raises
  `EmbeddingModelMismatchError` (cross-model similarity scores are meaningless).
  Note: capture is **in-process only** — a fresh runtime over an existing store
  won't enforce until its first ingest.
* `on_ingest` / `on_retrieve` callbacks for logging/observability;
  `delete_document(id)` convenience wrapper.

### `stores` module

`Store` protocol:
* added `delete_where(filters)` and `find(filters, limit=1)` (metadata-only
  lookup, no vector search) — both required by the runtime's upsert/staleness paths.

`StoreEntry`:
* `vector` is now `list[float] | None`. Read results no longer round-trip the
  vector (was `[]`, now `None`) — the backend owns the stored vector; callers must
  not rely on this field on retrieved entries.

`StoreQuery`:
* `scope` is now optional (`StoreScope | None`) for single-tenant callers.
* `metadata_filters` retyped `dict[str, Any]` (was `dict[str, str]`).
* removed the unused `strategies` field and the `RetrievalStrategy` enum.

`VectorStore` (base) / `VectorBackend`:
* `VectorBackend` protocol gained `list_where(filters, limit)` and `count(filters)`.
  **`count` lives on the backend, not `Store`** — keeps the runtime's dependency
  surface to the `Store` protocol alone.
* `VectorStore` now implements `find`, `delete_where`, and `count`.
* Payload encoding spreads scalar `chunk_metadata` values to the top level (in
  addition to the JSON-encoded blob) so flat-equality `metadata_filters` / `find`
  work against them.

Backend implementations (`chroma`, `in_memory`, `pgvector`) all implement
`list_where` + `count`. Plus:
* **pgvector** `_build_where` now compares JSONB-to-JSONB
  (`payload->$k::text = $v::jsonb`) so non-string scalars (int/bool/None) keep
  their JSON type instead of being stringified. Filters are parameterized; `LIMIT`
  is `int`-cast before interpolation. Added `pool_kwargs` passthrough to
  `asyncpg.create_pool` for tuning `min_size`/`max_size`/etc.
* **in_memory** `_flush` is now async — JSON encode happens under the lock, the
  disk write is offloaded to a thread so the event loop isn't blocked. Search now
  sanitizes non-finite scores (NaN/inf from a misbehaving embedder): they're
  logged and sorted/dropped to the end instead of corrupting the ranking.

### `loaders` module

`Document`:
* **`id` is now derived deterministically from `source` via
  `uuid5(NAMESPACE_URL, source)`** so re-ingesting the same source yields the same
  id across processes. Fixes a silent upsert bug where modified files left their
  prior chunks orphaned in the store, because `delete_where({"document_id": ...})`
  was keyed on a fresh random UUID each pass. Sourceless documents fall back to
  `uuid4()` (no stable identity ⇒ no upsert semantics).
* added `content_hash` (SHA-256, computed by the runtime at ingest time; loaders
  leave it `None`) used by staleness detection. `type` now defaults to
  `DocumentType.TEXT`.

* `Sanitizer` protocol for PII redaction (sync **or** async `sanitize`; errors
  propagate, no logic baked into the framework).
* `SanitizingLoader` wraps any `BaseDocumentLoader` + a `Sanitizer`, running every
  yielded document through it.

### Removals / cleanup

* Deleted the legacy `railtracks.vector_stores` package (`chroma`, `chunking/`,
  `filter`, `vector_store_base`) and its tests — fully superseded by
  `railtracks.retrieval.stores` and `railtracks.retrieval.chunking` (~7.5k lines).
* `retrieval.__init__` now exports the public surface: `RetrievalRuntime`, the
  ingestion event/stats types, `Store`, `StoreEntry`, `StoreQuery`, `StoreScope`,
  `VectorStore`, `EmbeddingFailure`, `EmbeddingModelMismatchError`.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

**Review callouts**
* Ingest upsert is not transactional (per-chunk writes); count-aware staleness is
  what makes an interrupted ingest self-heal on the next run.
* `_captured_model` is in-process only — model-mismatch enforcement doesn't
  survive a fresh runtime over a pre-populated store until its first ingest.
* pgvector `list_where` interpolates `LIMIT {int(limit)}` (int-cast, safe); all
  filter values stay parameterized.

**Known limitations / follow-ups**
* The `max_tokens` guard enforces a **per-item** token cap (drops oversized chunks
  pre-embedding); it does **not** do batch-level token-budget packing. Batches are
  still sized by count (`default_batch_size`), so a batch of in-spec chunks can
  still exceed a provider's per-request token limit (e.g. OpenAI's 8191). Worth a
  follow-up for token-aware batch packing.


Amir-R25:
A few example scripts to play with. All three share the same wiring — only the
backend and whether you ingest differ.

### In-memory ingest + retrieve

```python
from rich import print

from railtracks.retrieval import RetrievalRuntime, VectorStore
from railtracks.retrieval.chunking import SentenceChunker
from railtracks.retrieval.loaders import TextLoader
from railtracks.retrieval.stores import InMemoryVectorBackend
from railtracks.retrieval.embedding import OpenAIEmbedding
from railtracks.retrieval.embedding.models import EmbeddingFailure
from railtracks.retrieval.runtime import BatchIngested, DocumentFailed, DocumentSkipped


async def main() -> None:
    docs_path = "path/to/directory"
    rr = RetrievalRuntime(
        chunker=SentenceChunker(chunk_size=5, overlap=2),
        embedder=OpenAIEmbedding(model="text-embedding-3-small"),
        store=VectorStore(InMemoryVectorBackend()),
        batch_size=64,
    )

    loader = TextLoader(str(docs_path))

    async for event in rr.ingest(loader):
        match event:
            case BatchIngested(document_id=did, embedded_chunks=ch, batch_index=i):
                print(f"  + doc={str(did)[:8]} batch={i} chunks={len(ch)}")
            case EmbeddingFailure(errors=errs):
                print(f"  ! embedding failed: {errs[0]}")
            case DocumentFailed(document_id=did):
                print(f"  ! doc {str(did)[:8]} partially failed")
            case DocumentSkipped(source=src):
                print(f"  ~ skipped (unchanged): {src}")

    result = await rr.retrieve("query text")
    print(f"\nQuery: {result.query}")
    for hit in result.chunks:
        snippet = hit.chunk.content.replace("\n", " ")
        print(f"  [score={hit.score:.3f}] {snippet}")


if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```

### Persistent ingest + retrieve with Chroma

```python
from rich import print

from railtracks.retrieval import RetrievalRuntime, VectorStore
from railtracks.retrieval.chunking import SentenceChunker
from railtracks.retrieval.loaders import TextLoader
from railtracks.retrieval.stores import ChromaBackend
from railtracks.retrieval.embedding import OpenAIEmbedding
from railtracks.retrieval.embedding.models import EmbeddingFailure
from railtracks.retrieval.runtime import BatchIngested, DocumentFailed, DocumentSkipped


async def main() -> None:
    docs_path = "path/to/directory"
    vsb = ChromaBackend("my_collection", path="retrieval-demos/stores")
    await vsb.initialize()

    rr = RetrievalRuntime(
        chunker=SentenceChunker(chunk_size=5, overlap=2),
        embedder=OpenAIEmbedding(model="text-embedding-3-small"),
        store=VectorStore(vsb),
        batch_size=64,
    )

    loader = TextLoader(str(docs_path))

    async for event in rr.ingest(loader):
        match event:
            case BatchIngested(document_id=did, embedded_chunks=ch, batch_index=i):
                print(f"  + doc={str(did)[:8]} batch={i} chunks={len(ch)}")
            case EmbeddingFailure(errors=errs):
                print(f"  ! embedding failed: {errs[0]}")
            case DocumentFailed(document_id=did):
                print(f"  ! doc {str(did)[:8]} partially failed")
            case DocumentSkipped(source=src):
                print(f"  ~ skipped (unchanged): {src}")

    result = await rr.retrieve("query text")
    print(f"\nQuery: {result.query}")
    for hit in result.chunks:
        snippet = hit.chunk.content.replace("\n", " ")
        print(f"  [score={hit.score:.3f}] {snippet}")


if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```

### Retrieve against a previously-ingested Chroma store

Re-running ingest on unchanged sources is a no-op (see count-aware staleness), so
a retrieve-only script just opens the existing collection — no loader needed.

```python
from rich import print

from railtracks.retrieval import RetrievalRuntime, VectorStore
from railtracks.retrieval.chunking import SentenceChunker
from railtracks.retrieval.stores import ChromaBackend
from railtracks.retrieval.embedding import OpenAIEmbedding


async def main() -> None:
    vsb = await ChromaBackend.create("my_collection", path="retrieval-demos/stores")

    rr = RetrievalRuntime(
        chunker=SentenceChunker(chunk_size=5, overlap=2),
        embedder=OpenAIEmbedding(model="text-embedding-3-small"),
        store=VectorStore(vsb),
        batch_size=64,
    )

    result = await rr.retrieve("query text")
    print(f"\nQuery: {result.query}")
    for hit in result.chunks:
        snippet = hit.chunk.content.replace("\n", " ")
        print(f"  [score={hit.score:.3f}] {snippet}")


if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```

docs/rag/components/design.md
```diff
@@ -33,127 +35,146 @@ flowchart LR
...
-    class C1 chunk;
-    class B1,B2,B3 embed;
-    class Store store;
+## Module layout
 
-    style LoaderPool fill:transparent,stroke:#60A5FA,stroke-width:1px
-    style ChunkerPool fill:transparent,stroke:#FBBF24,stroke-width:1px
-    style EmbedPool fill:transparent,stroke:#FECACA,stroke-width:1px
+```
+railtracks/retrieval/
```
soulFood5632, line 74: Lets flag this. I don't really like putting this type of things in docs, its really hard mantian if we ever change file names

docs/rag/components/design.md
```diff
@@ -33,127 +35,146 @@ flowchart LR
...
 
 ### Chunkers
-`BaseChunker` exposes `chunk(doc) → list[Chunk]` and `chunk_many(docs) → list[Chunk]`. Each `Chunk` carries a `document_id` for full lineage back to the source `Document`. Three strategies ship out of the box: `FixedCharChunker`, `FixedTokenChunker`, and `RecursiveChunker`.
-
-### Embedding Service
-`BaseEmbeddingService` requires both `embed()` and `aembed()`. Every response is an `EmbeddingResponse` containing the vectors **and** an `EmbeddingStats` object with model name, token count, cost, and latency — cost tracking is not optional.
 
-The default implementation, `LiteLLMEmbeddingService`, handles batching automatically and reads cost from litellm's `_hidden_params["response_cost"]`.
+`Chunker.chunk(document) → list[Chunk]` is the sync split primitive;
+`achunk` and `astream_documents` are derived. Subclasses delegate to a
```
soulFood5632, line 100: again for future but this is a really wordy. I feel like we can improve the language here.

docs/rag/components/design.md
```diff
@@ -33,127 +35,146 @@ flowchart LR
...
-pipeline = RAGPipeline(
-    loader=PyPDFLoader("docs/"),
-    chunker=RecursiveChunker(chunk_size=512),
-    embedding_service=LiteLLMEmbeddingService("text-embedding-3-small"),
-    vector_store=InMemoryVectorStore(),
-)
-
-stats: IndexingStats = await pipeline.aindex()
-results: SearchResult = await pipeline.aquery("What is the return policy?", top_k=5)
+class Store(Protocol):
```
soulFood5632, line 117: same comment as above, lets avoid putting code like this in our docs becuase it is such a pain

docs/rag/components/stores/overview.md
soulFood5632, line 1: From a high level, the docs here a bit dense and defiently need some human touch, it feels more like an info dump that a helpful guide on how I move forward with it. This should be issues up for future work

docs/rag/components/stores/overview.md
```diff
@@ -27,7 +27,7 @@ class StoreEntry:
     # Required — sourced from EmbeddedChunk
     id: UUID
     content: str
-    vector: list[float]
+    vector: list[float] | None
```
soulFood5632, line 30: Same here, lets not put code here. Its so hard to mantain if we ever make code changes.

docs/rag/components/design.md
```diff
@@ -33,127 +35,146 @@ flowchart LR
...
+
+The runtime always converts back to `RetrievedChunk` (a thin shape around
+`Chunk`) so the user-facing `RetrievalResult` doesn't expose store-internal
+fields like `scope` or `embedding_version`.
+
+## Upsert and staleness
+
+Two protocol additions make ingestion safe to re-run:
+
+- **`delete_where`** lets the runtime clear prior chunks for a document
```
soulFood5632, line 146: to my comment above here, this is just hella verbose. We need to just provide simple conscise instruction here for better results.

docs/rag/components/design.md
```diff
@@ -33,127 +35,146 @@ flowchart LR
...
+Mixing vectors from different embedding models produces meaningless
+similarity scores. The runtime captures the embedder's model name on the
+first successful batch and raises `EmbeddingModelMismatchError` at retrieve
+time if the embedder later reports a different model. This is in-process
+only — cross-process consistency (different agent restart, same store) is
+out of scope today.
+
+## Multi-tenancy
+
+`StoreScope(user_id, agent_id, session_id, run_id)` is a hard-filter
```
soulFood5632, line 166: The language here is just so dense. It really should be more readable

docs/rag/RAG.md
```diff
@@ -1,97 +1,129 @@
...
-## Two Ways to Get Started
-
-We offer **two approaches** to integrate RAG into your application:
+Pass the same runtime to `RagConfig` to attach it to an agent, or call
+`runtime.retrieve()` directly when you want raw control.
 
-!!! tip "Choose Your Adventure" 
-    1. **Quick & Easy**: Use the prebuilt `rag_node` for instant setup 
-    2. **Full Control**: Build a custom RAG node using the `RAG` class for maximum flexibility
+!!! info "Migrated from older modules"
```
soulFood5632, line 9: warning here instead of a info

soulFood5632 (commented):
Docs comments,

packages/railtracks/src/railtracks/prebuilt/__init__.py
```diff
@@ -1,5 +1,3 @@
 ######## This package contains a suite of pre-built ready to use agents designed to help you build faster #########
 
-__all__ = ["rag_node"]
-
-from railtracks.prebuilt.rag_node import rag_node
+__all__: list[str] = []
```
soulFood5632, line 3: so we are not exposing it. That fine just flagging this

packages/railtracks/src/railtracks/nodes/nodes.py
```diff
@@ -143,7 +144,9 @@ async def tracked_invoke(self) -> _TOutput:
         start_time = time.time()
         try:
             for func in self.pre_invokes:
-                func(self)
+                result = func(self)
+                if inspect.iscoroutine(result):
```
soulFood5632, line 148: Honestly, this logic should be migrated as part of the middleware initative. It should all be in one

docs/rag/RAG.md
Pooria90, line 1: Multiple empty code blocks; review needed before merging to main.

soulFood5632 (commented):
Some more comments

packages/railtracks/src/railtracks/nodes/nodes.py
soulFood5632: No action for this PR, just flagging it

docs/rag/components/stores/overview.md
soulFood5632, line 1: Make sure to include pdf in the .toml

soulFood5632, line 1: Include pypdfLoader in the init of loaders

soulFood5632: Start a fesh envoornment and make sure requirements are synced

soulFood5632: Start a fesh envoornment and make sure requirements are synced

soulFood5632, line 1: I don't really love the difference entrance into a chroma DB versus in memory store one is a corountine and requires the .create function where the other is just a constructor:

```python 
await ChromaBackend.create()
```

vs. 

```python
InMemoryStore()
```

I feel like it should be consistent

soulFood5632, line 1: no chroma in the pytoml either

soulFood5632: also what do you think about making it chroma instead of stores-chroma

soulFood5632 (commented):
Some general comments here

packages/railtracks/src/railtracks/retrieval/runtime.py
```diff
@@ -0,0 +1,488 @@
+"""High-level orchestration of the retrieval pipeline.
```
Pooria90: I think we should include these details in class docstring, not module details.

packages/railtracks/src/railtracks/retrieval/runtime.py
```diff
@@ -0,0 +1,488 @@
...
+class DocumentSkipped:
+    """A document skipped during ingest because the store already has an
+    entry with the same ``source_path`` and ``content_hash``."""
+
+    document_id: UUID
+    source: str
+    reason: str = "unchanged"
+
+
+IngestionEvent = Union[BatchIngested, EmbeddingFailure, DocumentFailed, DocumentSkipped]
```
Pooria90: I guess we should reorganize data models. For example `EmbeddingFailure` is coming from another module in here.

packages/railtracks/src/railtracks/retrieval/runtime.py
```diff
@@ -0,0 +1,488 @@
...
+                f"{type(embedder).__name__} does not declare a "
+                "default_batch_size. Pass batch_size= to RetrievalRuntime "
+                "or set default_batch_size on the embedder class."
+            )
+        return bs
+
+    async def ingest(
+        self, loader: BaseDocumentLoader
+    ) -> AsyncGenerator[
+        BatchIngested | EmbeddingFailure | DocumentFailed | DocumentSkipped, None
```
Pooria90, line 206: Can this be simplified to `IngestionEvent`?

Pooria90 (changes requested):
Thanks for the awesome work, Amir!

I have minor comments on the code. We need to dedicate some time to organize and polish the documents.

docs/rag/components/design.md
Amir-R25: Documented in #1123 to handle before merging into main

docs/rag/components/design.md
Amir-R25: Documented in #1123 to handle before merging into main

docs/rag/components/design.md
Amir-R25: Documented in #1123 to handle before merging into main

docs/rag/components/stores/overview.md
Amir-R25: Documented in #1123 to handle before merging into main

docs/rag/components/stores/overview.md
Amir-R25: Documented in #1123 to handle before merging into main

docs/rag/components/design.md
Amir-R25: Documented in #1123 to handle before merging into main

docs/rag/components/design.md
Amir-R25: Documented in #1123 to handle before merging into main

docs/rag/RAG.md
Amir-R25: Documented in #1123 to handle before merging into main

docs/rag/RAG.md
Amir-R25: Documented in #1123 to handle before merging into main

docs/rag/components/stores/overview.md
Amir-R25: Yeup, will make sure these are all polished before merging the feature branch into main, thanks for flagging

Amir-R25: Documented in #1123 to handle before merging into main

packages/railtracks/src/railtracks/retrieval/runtime.py
Amir-R25: I actually went the other way and removed `IngestionEvent` since the public API needs to use `BatchIngested | EmbeddingFailure | DocumentFailed | DocumentSkipped` so users can see what the expected outputs are

PR #1110, opened 2026-05-21
Pooria/1024 railtracks folder

### LINKED ISSUE #1024: [Visualizer] `.railtracks` folder location
### Issue Type

Bug/Error

### Description

I noticed a potentially problematic issue with how we are storing the json files. Let's say the user has ran `pip install railtracks[cli]` and they have the following folder structure:

```
project-root/
├── agent_1.py
├── first_folder/
│   └── agent_2.py
└── venv/
```

if they run either agent from `project-root`, by running `python agent_1.py` and `python first_folder/agent_2.py`, their directory will look like this:

```
project-root/
├── .railtracks/
├── agent_1.py
├── first_folder/
│   └── agent_2.py
└── venv/
```

However, if they run either agent from within `first_folder` by running `python ../agent_1.py` and `python agent_2.py`, then their directory will look like this:

```
project-root/
├── agent_1.py
├── first_folder/
│   ├── .railtracks/
│   └── agent_2.py
└── venv/
```
What if they'd done both throughout their dev cycle? Then they'll have this:
```
project-root/
├── .railtracks/
├── agent_1.py
├── first_folder/
│   ├── .railtracks/
│   └── agent_2.py
└── venv/
```
And if we're being honest, it's highly likely developers are not paying attention on the directory they're running their `python` scripts from.

So what happens now? Some of their data will be in `project-root/.railtracks` and some of it will be in `project-root/first_folder/.railtracks` and the visualizer will also behave similarly based on where `railtracks init` is run. This has the potential to cause significant confusion for the users.

The only way I can see of addressing this is to require `RAILTRACKS_HOME` to be set by the users for one unified location for storage.

cc: @jbueza-railtownai @soulFood5632

Pooria90:
## Summary

Fixes inconsistent `.railtracks` directory placement when agents are run from different working directories (closes #1024).

Previously, each call site resolved the data directory independently — `_session.py` read `RAILTRACKS_HOME` but fell back to `cwd()`, while `evaluations/utils.py`, `cli/__init__.py`, and `cli/viz_server.py` hardcoded `".railtracks"` relative to `cwd()`. Running scripts from a subdirectory silently created a second `.railtracks` folder, splitting session and evaluation data across locations.

The fix introduces a single `resolve_railtracks_home()` function in `railtracks/paths.py` used by all four sites. Resolution order:
1. `RAILTRACKS_HOME` env var, if set
2. Walk up from `cwd()` until an existing `.railtracks` directory is found (git-style traversal)
3. Fall back to `cwd()/.railtracks` with a `UserWarning` prompting the user to run `railtracks init` from their project root

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

## Notes

**Migration:** No breaking change for users who already set `RAILTRACKS_HOME` — that path takes priority and behaviour is unchanged. Users relying on the implicit `cwd()` fallback will see a `UserWarning` on first run until they run `railtracks init` from their project root; after that, traversal finds the directory and the warning is not emitted again.

**`railtracks init` behaviour:** If a `.railtracks` directory already exists in a parent folder, `init` reports the existing location rather than creating a second one. The `.gitignore` entry is now written adjacent to the `.railtracks` directory rather than always at `cwd()`.

**Removed:** `EVALS_DIR` module-level constant from `evaluations/utils.py` and `UI_VERSION_FILE` constant from `cli/constants.py` — both are now computed at call time via `resolve_railtracks_home()`. Tests that patched `EVALS_DIR` directly have been updated to patch `resolve_railtracks_home` instead.

### DIFF
--- a/packages/railtracks/src/railtracks/cli/__init__.py
+++ b/packages/railtracks/src/railtracks/cli/__init__.py
@@ -31,9 +31,10 @@ from pathlib import Path
 
 from colorama import Fore, Style
 
+from railtracks.paths import resolve_railtracks_home
+
 from .constants import (
     DEFAULT_PORT,
-    UI_VERSION_FILE,
     cli_directory,
     cli_name,
     latest_ui_url,
@@ -111,32 +112,34 @@ def is_port_in_use(port):
 
 def create_railtracks_dir():
     """Create .railtracks directory if it doesn't exist and add to .gitignore"""
-    railtracks_dir = Path(cli_directory)
+    railtracks_dir = resolve_railtracks_home()
     if not railtracks_dir.exists():
         print_status(f"Creating {cli_directory} directory...")
-        railtracks_dir.mkdir(exist_ok=True)
-        print_success(f"Created {cli_directory} directory")
+        railtracks_dir.mkdir(parents=True, exist_ok=True)
+        print_success(f"Created {railtracks_dir}")
 
-    gitignore_path = Path(".gitignore")
-    if gitignore_path.exists():
-        with open(gitignore_path) as f:
-            gitignore_content = f.read()
+        gitignore_path = railtracks_dir.parent / ".gitignore"
+        if gitignore_path.exists():
+            with open(gitignore_path) as f:
+                gitignore_content = f.read()
 
-        if cli_directory not in gitignore_content:
-            print_status(f"Adding {cli_directory} to .gitignore...")
-            with open(gitignore_path, "a") as f:
-                f.write(f"\n{cli_directory}\n")
-            print_success(f"Added {cli_directory} to .gitignore")
+            if cli_directory not in gitignore_content:
+                print_status(f"Adding {cli_directory} to .gitignore...")
+                with open(gitignore_path, "a") as f:
+                    f.write(f"\n{cli_directory}\n")
+                print_success(f"Added {cli_directory} to .gitignore")
+        else:
+            print_status("Creating .gitignore file...")
+            with open(gitignore_path, "w") as f:
+                f.write(f"{cli_directory}\n")
+            print_success(f"Created .gitignore with {cli_directory}")
     else:
-        print_status("Creating .gitignore file...")
-        with open(gitignore_path, "w") as f:
-            f.write(f"{cli_directory}\n")
-        print_success(f"Created .gitignore with {cli_directory}")
+        print_status(f"Using existing {railtracks_dir}")
 
 
 def get_stored_ui_version():
     """Get the stored UI version (ETag) from disk"""
-    version_file = Path(UI_VERSION_FILE)
+    version_file = resolve_railtracks_home() / ".ui_version"
     try:
         if version_file.exists():
             return version_file.read_text().strip()
@@ -147,7 +150,7 @@ def get_stored_ui_version():
 
 def save_ui_version(version: str):
     """Save the UI version (ETag) to disk"""
-    version_file = Path(UI_VERSION_FILE)
+    version_file = resolve_railtracks_home() / ".ui_version"
     try:
         version_file.write_text(version)
     except Exception:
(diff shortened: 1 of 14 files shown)


packages/railtracks/src/railtracks/paths.py
```diff
@@ -0,0 +1,34 @@
...
+    Resolution order:
+    1. RAILTRACKS_HOME env var — .railtracks is created inside this directory.
+    2. Walk up from cwd() looking for an existing .railtracks directory.
+    3. Fall back to cwd()/.railtracks with a UserWarning.
+    """
+    env = os.environ.get("RAILTRACKS_HOME")
+    if env:
+        return Path(env) / _DIRNAME
+
+    current = Path.cwd()
```
jbueza-railtownai, line 20: This should be fine in Windows too, right? (Cross-operating system file paths)

jbueza-railtownai (approved):
LGTM

lgtm

packages/railtracks/src/railtracks/cli/constants.py
```diff
@@ -2,7 +2,6 @@
 
 cli_name = "railtracks"
 cli_directory = ".railtracks"
-UI_VERSION_FILE = f"{cli_directory}/.ui_version"
```
Amir-R25, line 5: This constant is used by the CLI to detect the user's version of the visualizer UI vs the most recently published version and warns the user to update to get new features

packages/railtracks/src/railtracks/cli/__init__.py
```diff
@@ -111,32 +112,34 @@ def is_port_in_use(port):
...
-        with open(gitignore_path, "w") as f:
-            f.write(f"{cli_directory}\n")
-        print_success(f"Created .gitignore with {cli_directory}")
+        print_status(f"Using existing {railtracks_dir}")
 
 
 def get_stored_ui_version():
     """Get the stored UI version (ETag) from disk"""
-    version_file = Path(UI_VERSION_FILE)
+    version_file = resolve_railtracks_home() / ".ui_version"
```
Amir-R25, line 142: Same comment here and below regarding UI_VERSION constant
it's a file inside `.railtracks`

packages/railtracks/src/railtracks/cli/viz_server.py
```diff
@@ -93,7 +95,7 @@ async def serve_ui_or_404(full_path: str):
     if full_path.startswith("api/"):
         return JSONResponse(content={"error": "Not Found"}, status_code=404)
 
-    ui_dir = Path(f"{cli_directory}/ui")
+    ui_dir = get_railtracks_dir() / "ui"
```
Amir-R25, line 98: will this make it `dir/.railtracks/ui`, or `dir/ui` cause only the first is correct, the second one will break the visualizer

Amir-R25 (changes requested):

packages/railtracks/src/railtracks/paths.py
Pooria90: Yes, works the same on Windows!

packages/railtracks/src/railtracks/cli/constants.py
Pooria90: `UI_VERSION_FILE` as a module-level constant was removed, but the functionality is fully preserved: it's now computed at call time. `get_stored_ui_version()` and `save_ui_version()` both use `resolve_railtracks_home() / ".ui_version"`, so the file still lives inside `.railtracks` and the version-checking/update-warning flow is unchanged.

packages/railtracks/src/railtracks/cli/viz_server.py
Pooria90: It's the first one!

Amir-R25 (approved):

Issue #1114, opened 2026-05-21
[Docs][Feature] Minor Doc Grammar and content fixes


CoronRing:
### Issue Categorization

- [X] Missing documentation
- [X] Incorrect or outdated information
- [X] Overly verbose or complex explanation
- [x] Broken links or formatting
- [X] Missing code examples
- [x] Typos or grammar issues

### Documentation Problem

This parent ticket collects minor grammar, content, and structural changes made to the documentation. They may or may not need a ticket, if the task is bigger, a ticket is advised. If the change is local, minor, or along with another fix/PR, one can simply leave this parent ticket (#1114) as reference, but do not close it.

Feel free to leave any grammar/wording issues you see in comments below

### Additional Context

This ticket is expected to be long-present and not closed even when no tracking.

CoronRing:
Batch 3
1. **Medium** — Fix broken link in `tutorials/walkthroughs/agents_as_tools.md:6` pointing to `system_internals/node.md`.
2. **Medium** — Decide fate of 26 orphaned docs: add to nav, move out of `docs/`, or explicitly exclude.
3. **Medium** — Investigate `railtracks.interactive` missing from `__all__` (pdoc Warning #1) and circular import for `_AsyncNodeAttachedFunc` (Warnings #4–5).
4. **Low** — Fix nav label `Email` → `Notion` in `mkdocs.yml:151`.
5. **Low** — Resolve `docs/js/api_collapsible.js` — reference it or remove it.

CoronRing:
packages\railtracks\src\railtracks\llm\tools\tool.py: Typos "Please use a cutom made function.",

CoronRing:
docs\documentation\agent_design\tools\agents_as_tools.md: 

```
One end-to-end test through Tool(name, desc, schema) with {"type": "bool"}, asserting the error names 'enabled'. That's the issue's actual repro path and it covers "dict-schema path" and "names the parameter" in a single test. Cheap.
```

Is a none working failing example.

soulFood5632:
I don't think this is an epic. its more of a feature.

CoronRing:
[resolved] README example missing `"` await flow.ainvoke("What's the weather in Paris?)

CoronRing:
README lacks steps to put openai key, should not assume user know

CoronRing:
README: Installation is redundant

Bash`pip install railtracks 'railtracks[visual]'`

but railtracks[visual] already include ` railtracks`

[...]

PR #1115, opened 2026-05-21
[Docs] Fix markdown list rendering in agent design overview

CoronRing:
part of  #1114.\n\nAdds a paragraph break so the ordered list in the agent design overview renders as separate items instead of collapsing onto one line in markdown renderers.


Amir-R25 (approved):
Thanks Guan!

PR #1118, opened 2026-05-25
Add LangChainLoaderAdapter (closes #1080)

### LINKED ISSUE #1080: [Feature] [Retrieval] `Langchain` Adapter
> [!WARNING]
> Please PR into `feature-branch-rag` and not `main`
### Feature Request
Add a `LangChainLoaderAdapter` that wraps LangChain's community document loaders and outputs our `Document` data model.

#### Problem
LangChain's community has contributed an extensive list of document loader modules. Re-implementing these loaders would be redundant when we could instead adapt them directly.

#### Proposed Solution
Add an adapter called `LangChainLoaderAdapter` that allows users to use any LangChain-compatible loader while outputting our `Document` data model.

#### Motivation
- Gives users access to LangChain's extensive loader ecosystem out of the box
- Avoids redundant re-implementation of loaders we don't need to own
- Keeps our `Document` data model as the standard output regardless of the source loader

#### Acceptance Criteria
- [ ] `LangChainLoaderAdapter` is implemented and extends `BaseDocumentLoader`
- [ ] Adapter accepts any LangChain-compatible loader as input
- [ ] Output is normalized to our `Document` data model
- [ ] Documented with a usage example in the README

#### Additional Context
- Review LangChain's loaders: https://reference.langchain.com/python/langchain-community/document_loaders
- Extend `BaseDocumentLoader`: https://github.com/RailtownAI/railtracks/blob/14f5f78ecf6cac2f6356f5373d071d9771b8369f/packages/railtracks/src/railtracks/retrieval/loaders/base.py#L10-L36
- Review `Document` data model: https://github.com/RailtownAI/railtracks/blob/14f5f78ecf6cac2f6356f5373d071d9771b8369f/packages/railtracks/src/railtracks/retrieval/models.py#L20-L35

DharunKosanam:
## Summary

Adds `LangChainLoaderAdapter`, which wraps any LangChain `BaseLoader` and normalises its output to the railtracks `Document` model. This unlocks LangChain's community loader ecosystem (Wikipedia, Notion, Confluence, S3, …) without re-implementing loaders we don't need to own. Closes #1080.

Key design points:
- **Duck-typed** on `alazy_load` / `lazy_load` / `load` — the adapter does not import `langchain` itself, so it stays a fully optional dependency.
- **Streams natively** when the wrapped loader exposes `alazy_load` or `lazy_load`; falls back to eager `load()` only when neither is available.
- `page_content` → `Document.content`, `metadata["source"]` is popped into `Document.source`, remaining metadata is preserved.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [x] Tests added/updated and pass locally (`pytest tests`) — 13 new tests, all green
- [x] Docs updated if user-facing behavior changed — new page at `docs/rag/components/ingestion/langchain.md` and mkdocs nav entry
- [ ] Breaking changes include migration notes — N/A

## Notes

Tests use in-file stub LangChain loaders so the test suite itself does not require LangChain to be installed.


soulFood5632:
Closes #1080

packages/railtracks/src/railtracks/retrieval/loaders/langchain_loader.py
```diff
@@ -0,0 +1,126 @@
...
+            return
+
+        if hasattr(self._loader, "lazy_load"):
+            iterator = await asyncio.to_thread(lambda: iter(self._loader.lazy_load()))
+            while True:
+                lc_doc = await asyncio.to_thread(next, iterator, _SENTINEL)
+                if lc_doc is _SENTINEL:
+                    return
+                yield self._convert(lc_doc)
+            return
```
Amir-R25: is this `return` line ever reached?

packages/railtracks/src/railtracks/retrieval/loaders/langchain_loader.py
```diff
@@ -0,0 +1,126 @@
...
+
+        docs = await asyncio.to_thread(self._loader.load)
```
Amir-R25: I know we check for the existence of one of the three in the constructor but we can add a `if hasattr(self._loader, "load"):` here for safety and future-proofing

Amir-R25 (changes requested):
Thank you so much for the great PR @DharunKosanam. I have two very small comments, let me know what you think! Please also update your branch cause there's a small merge conflict in `loaders/__init__.py`

packages/railtracks/src/railtracks/retrieval/loaders/langchain_loader.py
DharunKosanam: removed it. The while loop only exits via the return inside it, so that one was dead code.

packages/railtracks/src/railtracks/retrieval/loaders/langchain_loader.py
DharunKosanam: Done — added an `if hasattr(self._loader, "load"):` guard so astream() doesn't assume the constructor check.

DharunKosanam:
Hi @Amir-R25, thanks for the review! Quick update:

Both code comments are fixed** (commit a0ce8d6):
- Removed the unreachable `return` after the `lazy_load` loop.
- Added the `if hasattr(self._loader, "load"):` guard on the fallback path.

**On the merge conflict:** the conflict in `loaders/__init__.py` is just because PR #1109 added `SanitizingLoader` to the same file I added `LangChainLoaderAdapter` to. Easy fix — I'll keep both.

**One heads-up though:** when I went to update my branch, I noticed `feature-branch-rag` itself doesn't import right now:

ImportError: cannot import name 'Cost' from 'railtracks.retrieval.models'


Looks like #1109 removed the `Cost` class from `retrieval/models.py` but `retrieval/__init__.py` still imports it (lines 15 and 41). This is unrelated to my PR, so I didn't want to touch it. Could you patch the base branch (restore `Cost` or drop the stale import)? Once it imports cleanly I'll resolve the conflict and push so CI goes green. Thanks!

Amir-R25:
Thanks @DharunKosanam for pointing out the stale imports. Should be fixed now, let me know.

Also feel free to join our discord and we can chat there easier if you'd like: https://discord.gg/mMButrUU

Amir-R25 (approved):

Amir-R25:
@DharunKosanam thanks for resolving the conflicts, is this ready to merge from your side? I can merge it

DharunKosanam:
Thanks @Amir-R25! Yep, ready to merge from my side ,conflict's resolved.
I see your approval. Go ahead and merge whenever.

PR #1121, opened 2026-05-25
Fix prompt injection literal brace handling

### LINKED ISSUE #1116: [Bug] Prompt Injection Bug
### Description
When `prompt_injection` is enabled (the default), railtracks runs `str.format()`-style templating over **all** message content, not just template strings with intended `{variables}`. Any message whose text contains literal curly braces — arXiv/LaTeX text, code snippets, JSON, math like `{x : x > 0}` — gets parsed as format fields and crashes before the LLM is ever called.

### Environment
- railtracks 1.3.9
- Python 3.10
- OS: macOS

### Root cause

`inject_context` → `inject_values` → `Message.fill_prompt` unconditionally runs `KeyOnlyFormatter().vformat()` over the message content. This applies `.format()` semantics (field/attribute/index parsing) to arbitrary content, so any literal `{ }` is misinterpreted.

### Workarounds

- Disable injection globally: `railtracks.set_config(prompt_injection=False)`
- Or escape per-message before sending: `content.replace("{", "{{").replace("}", "}}")`

### Suggested fixes

- Only run templating when a message is explicitly marked as a template, or
- Restrict substitution to known context keys and treat unmatched/invalid `{ }` as literals (don't raise on attribute/index parsing), or
- Document this clearly and provide an opt-out at the `rt.call`/message level (not just session/global config).

### Steps to reproduce

```python
import asyncio
import railtracks as rt

class Out(BaseModel):
    text: str

Agent = rt.agent_node(
    name="A",
    system_message="Summarize the message.",
    output_schema=Out,
)

# Realistic content: text with literal braces (e.g. extracted from a PDF/LaTeX)
content = "We define the set {x. some property} and note (e.g. {a, b})."
asyncio.run(rt.call(Agent, content))
```

### Expected behavior

Message content passed to `rt.call` should be treated as data and delivered to the model verbatim. Literal `{ }` in content should not be interpreted as template fields (or there should be safe handling/escaping).

### Actual behavior

Crashes in the pre-LLM hook (before any API request) with:

```
File ".../railtracks/built_nodes/concrete/_llm_base.py", line 254, in _pre_llm_hook
    return inject_context(message_history)
File ".../railtracks/prompts/prompt.py", line 30, in inject_context
    inject_values(message_history, _ContextDict())
File ".../railtracks/utils/prompt_injection.py", line 19, in inject_values
    message.fill_prompt(value_dict)
File ".../railtracks/llm/message.py", line 174, in fill_prompt
    self._content = KeyOnlyFormatter().vformat(self._content, (), value_dict)
File ".../string.py", line 276, in get_field
    obj = getattr(obj, i)
AttributeError: 'str' object has no attribute ' '
```

The `AttributeError` is raised because the standard `Formatter.get_field` interprets a brace group such as `{x. }` as an attribute access (`getattr(value, ' ')`).

### Additional Context

_No response_

### Operating System

_No response_

### railtracks Version

_No response_

### Python Version

_No response_

shayan919293:
## Summary
- replace prompt injection formatting with key-only brace scanning
- preserve unknown/literal brace groups such as JSON, LaTeX-like text, and unbalanced braces
- add regression coverage for literal braces and lookup-only context mappings

Fixes #1116

## Validation
- uv run --group test --with rich pytest packages/railtracks/tests/unit_tests/utils/test_prompt_injection.py packages/railtracks/tests/unit_tests/prompt/test_prompt.py

### DIFF
--- a/packages/railtracks/src/railtracks/llm/prompt_injection_utils.py
+++ b/packages/railtracks/src/railtracks/llm/prompt_injection_utils.py
@@ -6,6 +6,44 @@ class KeyOnlyFormatter(string.Formatter):
     A simple formatter which will only use keyword arguments to fill placeholders.
     """
 
+    def vformat(self, format_string, args, kwargs):
+        result: list[str] = []
+        index = 0
+
+        while index < len(format_string):
+            char = format_string[index]
+
+            if char == "{":
+                if index + 1 < len(format_string) and format_string[index + 1] == "{":
+                    result.append("{")
+                    index += 2
+                    continue
+
+                end_index = format_string.find("}", index + 1)
+                if end_index == -1:
+                    result.append(char)
+                    index += 1
+                    continue
+
+                key = format_string[index + 1:end_index]
+                try:
+                    result.append(str(kwargs[key]))
+                except KeyError:
+                    result.append(f"{{{key}}}")
+                index = end_index + 1
+                continue
+
+            if char == "}":
+                if index + 1 < len(format_string) and format_string[index + 1] == "}":
+                    result.append("}")
+                    index += 2
+                    continue
+
+            result.append(char)
+            index += 1
+
+        return "".join(result)
+
     def get_value(self, key, args, kwargs):
         try:
             return kwargs[str(key)]
--- a/packages/railtracks/tests/unit_tests/utils/test_prompt_injection.py
+++ b/packages/railtracks/tests/unit_tests/utils/test_prompt_injection.py
@@ -18,6 +18,32 @@ def test_formatter_missing_key_returns_placeholder():
     formatted = f.format("Hello, {name}")
     assert formatted == "Hello, {name}"
 
+def test_formatter_treats_unknown_attribute_like_field_as_literal():
+    f = KeyOnlyFormatter()
+    formatted = f.format("We define the set {x. some property} and keep it.")
+    assert formatted == "We define the set {x. some property} and keep it."
+
+def test_formatter_treats_json_like_content_as_literal():
+    f = KeyOnlyFormatter()
+    formatted = f.format('Payload: {"items": [1, 2], "ok": true}')
+    assert formatted == 'Payload: {"items": [1, 2], "ok": true}'
+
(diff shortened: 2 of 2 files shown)


Amir-R25:
Thanks a lot for digging into this, @shayan919293 and for adding regression tests. Your patch does stop the crash which is one of the criteria of the issue.

I'm going to close this one rather than merge it, though, because I don't think it fully resolves #1116, and I want to save you a rework cycle by explaining where it lands.

The issue's stated expectation is that content passed to `rt.call` reaches the model verbatim. Two gaps remain with this approach:

1. Silent rewriting on collision. A `{token}` in user content that matches a real context key still gets substituted:
 When you call `rt.call(Agent, "some text")`, that string is wrapped in a `UserMessage` with `inject_prompt=True` (the default). Before the request goes to the model, `inject_values` runs the formatter over every `inject_prompt=True` message and replaces any `{key}` with the matching context variable. So braces in the input are *always* treated as template placeholders and there's no distinction between "a template the developer wrote" and "data the caller passed in." Suppose a context var named `name` exists somewhere (e.g. `rt.context.put("name", "Bob")`, or it was seeded into the session context). Now a caller does:

```python
rt.call(Agent, "Use the {name} field in the schema.")
```
The caller means that literally `{name}` is part of their data. But because templating runs over it, `{name}` matches the context key and the model actually receives:
`Use the Bob field in the schema.`

2. The formatter is the wrong layer. I prototyped this locally: any fix that lives in the formatter (the hand-rolled `vformat` here, or a defensive `get_field`) still can't make content verbatim, because `.format()` semantics interpret `:, !, ., []` inside braces. For example a JSON blob `{"a":1}` gets mangled to `{"a"}` (the `:1` is parsed as a format spec). You can stop the crash there, but not the corruption.

The root cause is really a data-vs-template decision at the message level, which is the issue's first suggested fix. The direction that works. 

That being said, this is a core framework issue and requires some knowledge of the inner workings of our code. I'd suggest you start with one of the issues tagged with "Help Wanted" to get started.

Happy to chat more, looking forward to having you around as a contributor :)

PR #1158, opened 2026-06-02
fix: flow.invoke() works inside a running event loop (notebook support)

### LINKED ISSUE #1134: [Feature] uniform interface for notebook vs python
## intro 
Currently, we require the use of the async version of code like Flow in notebook; however, this requires a different set of interface and documentation (never shown an example of). It will be easier to learn and use if we can uniformize this (ie: make invoke work in notebook context). 

In addition, if a regular flow. invoke is imported from a Python into a notebook, it will break unless the tool itself is modified or hacks are used. This should be addressed so that pre-built tools and agents are accessible in both common contexts. 

### example
```
import railtracks as rt
def sample(t1):
    # To create your agent, you just need a model and a system message. 
    Agent = rt.agent_node(
        system_message="You are a helpful AI assistant."
    )

    # Create a function node that will be the entry point of our flow. 
    # This is where we will call our Agent.
    @rt.function_node
    async def main(message: str):

        # Now to call the Agent, we just need to use the `rt.call` function
        result = await rt.call(
            Agent,
            message,
        )
        return result

    # Create your flow and set the entry point to the function we just created. 
    # Then we can invoke the flow with a the input to the function node. 
    flow = rt.Flow("Quickstart Example", entry_point=main)

    result = flow.invoke(t1)
    print(result)
    
if __name__ == "__main__":
    sample("Hello, what can you do?")

```
Will fail if imported and called in an event loop such as notebook

### Additional Context

_No response_

CoronRing:
## Summary

- `flow.invoke()` previously raised `RuntimeError` when called from a Jupyter notebook or any other async host (FastAPI, async test runner) because `asyncio.run()` rejects calls from within an already-running event loop
- Fixed by detecting the running loop and dispatching to a `ThreadPoolExecutor` with its own fresh event loop; `contextvars.copy_context()` ensures Session and logging `ContextVar`s are visible inside the worker thread
- Removed `nest_asyncio` (monkey-patches the live loop globally — unnecessary side-effect); replaced with a stdlib-only approach using `concurrent.futures` + `contextvars`
- Docstring 

## Behaviour

| Caller context | Before | After |
|---|---|---|
| Plain script / pytest | ✅ works | ✅ works |
| Jupyter notebook | ❌ `RuntimeError` | ✅ works |
| FastAPI / async test runner | ❌ `RuntimeError` | ✅ works |
| `await flow.ainvoke()` anywhere | ✅ works | ✅ works (unchanged) |

## Test plan

- [ ] `test_flow_invoke_in_an_event_loop` — updated: previously asserted `RuntimeError`, now asserts the correct return value is produced when called from inside a running loop
- [ ] `test_flow_invoke_in_event_loop_preserves_context` — new: verifies that Flow-level context is visible to nodes when `invoke()` dispatches through the worker thread
- [ ] All existing `test_flows.py` tests continue to pass unchanged

### DIFF
--- a/packages/railtracks/src/railtracks/orchestration/flow.py
+++ b/packages/railtracks/src/railtracks/orchestration/flow.py
@@ -24,40 +24,29 @@ _P = ParamSpec("_P")
 class Flow(Generic[_P, _TOutput]):
     """A reusable, configured entry point for running an agent graph.
 
-    A ``Flow`` binds an entry-point node to a fixed set of runtime options
-    (context, timeout, callbacks, etc.) so the same configuration can be
-    invoked repeatedly without repeating arguments.  Each invocation is fully
-    isolated — flows never share state between runs.
+    Binds an entry-point node to a fixed set of runtime options so the same
+    configuration can be invoked repeatedly.  Each invocation is fully isolated.
 
     Typical usage::
 
         flow = Flow("my-agent", entry_point=my_node, context={"user": "alice"})
-
-        # async (preferred — native event loop, no overhead)
-        result = await flow.ainvoke(query)
-
-        # sync (works in scripts and Jupyter; dispatches to a worker thread
-        # when an event loop is already running)
-        result = flow.invoke(query)
+        result = await flow.ainvoke(query)   # async (preferred)
+        result = flow.invoke(query)          # sync
 
     Args:
         name: Unique human-readable name used in logging and state filenames.
         entry_point: The node (or decorated function) that starts the graph.
-        context: Key/value pairs made available to every node via
-            ``rt.context`` for the duration of the run.  Deep-copied at
-            invocation time so mutations inside a run never affect later runs.
-        timeout: Maximum seconds to wait for the top-level call to complete.
-            ``None`` means no limit.
-        end_on_error: When ``True``, the first unhandled exception aborts the
-            run immediately.
-        broadcast_callback: Called with each broadcast string emitted by
-            ``rt.broadcast()``.  May be sync or async.
-        prompt_injection: When ``True``, prompt text is automatically injected
-            from context variables before the run starts.
-        save_state: When ``True``, the session state is persisted to
-            ``.railtracks/data/sessions/`` after the run completes.
-        payload_callback: Called with the final result payload once the run
-            finishes successfully.
+        context: Key/value pairs available to every node via ``rt.context``.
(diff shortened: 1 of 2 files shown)


packages/railtracks/src/railtracks/orchestration/flow.py
```diff
@@ -73,14 +97,41 @@ def __init__(
...
+
+        Returns:
+            A new :class:`Flow` instance with the merged context.
         """
         new_obj = deepcopy(self)
         new_obj._context.update(context)
         return new_obj
 
     async def ainvoke(self, *args: _P.args, **kwargs: _P.kwargs) -> _TOutput:
+        """Run the flow asynchronously and return the entry-point result.
```
soulFood5632, line 104: I don't want this to be as verbose. I.e. I dont want to mention the Session class in this. I think the comment about context makes sense and we should include it but just a little smaller so it shows up inline.

packages/railtracks/src/railtracks/orchestration/flow.py
```diff
@@ -20,21 +22,43 @@
 
 
 class Flow(Generic[_P, _TOutput]):
-    """
-    Initializes a Flow object with a provided entry point and a unique name.
+    """A reusable, configured entry point for running an agent graph.
+
+    A ``Flow`` binds an entry-point node to a fixed set of runtime options
```
soulFood5632: Like below i don't want to expose `Session` as part of this docstring.

soulFood5632 (changes requested):
Looks good, one minor comment around the docstring. In future try and keep PRs around one single topic. If you want to update the docs that are not relevant you should do ot in a separate PR

packages/railtracks/src/railtracks/orchestration/flow.py
CoronRing: I guess I added this on purpose so it align with the docs: 

https://docs.railtracks.org/system_internals/node/#ii-within-a-session-context

Ie how session is referenced many placed acrossed docs but not mention how to create (ie not Session class directly). 

Perhaps we should update docs to match instead?
@soulFood5632 

Edit: I guess system_internals is hidden on the public version of the doc which is fine, will remove session reference.

soulFood5632: Yea a lot of systems internals is really out of date now. Good point on #1159

soulFood5632 (approved):
LGTM

Issue #1132, opened 2026-05-28
[Tech Debt] Backend Node Improvement to Prep for Generalized Node Enhancements


soulFood5632:
### Description (and/or) Solution

TBD

soulFood5632:
Closed in #1151

PR #1168, opened 2026-06-04
Reimplement context injection at the LLM/gateway level

### LINKED ISSUE #1141: [Feature] Reimplement context injection
### Feature Request

Previously we had a module which completed context injection. This module preformed admirably but used a old functionality which was removed in #1132 so it will need to be reimplemented. 

This should utilize the LLM level wrappers and should allow for users to manage their context injection through that interface. 

Concretely:
- Reimplement previous functionality of context injection 
- Provide clear guidance on how to turn it on and off both at the message level (previously capable of this) and at the LLM level (new functionality)

The exact implementation is still to be determined. 

### Additional Context

_No response_

CoronRing:
## Summary

Reimplements the **context injection** feature that was removed in #1132 (the NodeBuilder unification). Context injection fills `{placeholder}` templates in prompts with values from the run context (`rt.context` / `Flow.update_context`).

Previously this lived in the LLM base class''s pre-LLM hook, which #1132 removed. It is now reimplemented as a **gateway pre-mapper** wired in at `NodeBuilder.llm()` — the single construction point for *all* LLM nodes — so it applies universally by default, faithfully matching the old behaviour rather than only covering `agent_node()`.

## Control levels

Injection is enabled by default and can be disabled at three scopes:

| Scope | How | Status |
| --- | --- | --- |
| **Session / Flow** | `prompt_injection=False` on `rt.Flow(...)` / `rt.set_config(...)` | existing |
| **Agent / node** | `context_injection=False` on `rt.agent_node(...)` / `NodeBuilder.llm(...)` | **new** |
| **Message** | `inject_prompt=False` on a message, or `{{escaped}}` braces | existing |

## Design notes

- `ModelGateway` stays generic — it has no knowledge of context injection. It only gains a generic `add_pre_mapper()` helper.
- The injection decision lives in `NodeBuilder.llm(context_injection=...)`; `agent_node()` exposes and forwards the same flag.
- The pre-mapper runs after message assembly, so it sees the system message (required for the common `{secret}`-in-system-message case).

## Changes

- `prompts/prompt.py` — new `context_injection_gateway_pre_mapper`
- `built_nodes/llm_helpers.py` — generic `ModelGateway.add_pre_mapper()`
- `built_nodes/_node_builder.py` — `context_injection` param on `NodeBuilder.llm()`, wires the pre-mapper
- `built_nodes/easy_usage_wrappers/agent.py` — `context_injection` param on `agent_node()`
- Docs: walkthrough + example script updated to document all three control levels
- Tests: added `test_prompt_injection_node_level_bypass`

## Testing

All 6 prompt tests pass (5 existing + 1 new). The 2 unrelated failures in `built_nodes` pre-exist on the base branch (old NodeBuilder API in `test_agent.py`, unrelated `tool_info`/`issubclass` assertions).


CoronRing:
@soulFood5632 This version should be the expected add-on version

CoronRing:
I added a add_pre_mapper, but it may not be designed as intended because there is currently no clear mapper order loggic, so I kept it as simple as possible.

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
```diff
@@ -158,13 +160,26 @@ def llm(
...
         casted_instance._node_name = name
         casted_instance._node_class = "Agent"
 
+        # Context injection is the default behaviour for all LLM nodes: rt.context
+        # variables are filled into prompt templates before the model call. It is
+        # wired in as a gateway pre-mapper here so it applies regardless of which
+        # higher-level wrapper (agent_node, etc.) created the node. Disable per-node
+        # with context_injection=False, or session-wide via prompt_injection=False.
+        if context_injection:
+            from railtracks.prompts.prompt import (
```
soulFood5632: Shouldn't be a local import here.

docs/scripts/prompts.py
```diff
@@ -26,10 +26,21 @@
     prompt_injection=False
 )
 
-# or globally via 
+# or globally via
 rt.set_config(prompt_injection=False)
```
soulFood5632, line 30: I'll create a ticket to changes the language here so it is consistent

docs/tutorials/walkthroughs/prompts_and_context.md
```diff
@@ -8,17 +8,33 @@ In this example, the system message will be expanded to: "You are a technical as
 
 ### Enabling and Disabling Context Injection
 
-Context injection is enabled by default but can be disabled if needed:
+Context injection is enabled by default but can be disabled at three levels, from broadest to narrowest scope:
+
+| Scope | How | When to use |
+| --- | --- | --- |
+| **Session / Flow** | `prompt_injection=False` on `rt.Flow(...)` / `rt.set_config(...)` | Turn injection off for an entire run or globally |
```
soulFood5632: Session should be a internal concept so probably shouldn't mention it here. 

There is actually a 4th level we need to worry about as well. 

rt.set_config() global
rt.flow() for any instantiation of that flow
rrt.agent_node() for any instationation of that node
Message for the message

soulFood5632: we should also be clear about the precedence explicitly

soulFood5632 (changes requested):
Looks great. we may want to be careful when we expose the mappers as pass in able to make sure the order of mapper is done right but we can't deal with that anyway. 

other than that a couple of minor comments

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
CoronRing: Updated

docs/tutorials/walkthroughs/prompts_and_context.md
CoronRing: Excluded, I think 4 level is a bit overkill, I even wanted to remove it at a global level as that behaviour seem never "global-able", but lets keep it for consistency. 

4th level added

CoronRing: "Context injection is enabled by default but can be disabled at four levels, from broadest to narrowest scope:" added

soulFood5632: To me that is a really clunky way to describe it. It's kind of tricky to describe but lets say someone sets it to False at the node level and then sets it to true at a message level. Its not clear to me what behavior I would get (I actually have no idea what actually happens in our codebase tbh). 

However, we handle it we need be clear about the cases for precedence and for overriding.

soulFood5632 (commented):
See my above response to previous comment. I don't actually now the behavior we have in the situation I mentioned but we should determine and present it clearly. If we think we have bad design we can design a 3 state which better reflects the precedence we want.

packages/railtracks/src/railtracks/built_nodes/llm_helpers.py
```diff
@@ -92,6 +92,11 @@ def __init__(
         self._pre_mapping = pre_mappers or []
         self._post_mapping = post_mappers or []
 
+    def add_pre_mapper(self, mapper: GatewayPreMapper) -> None:
+        """Register an additional pre-mapper, applied after any existing ones."""
+        if mapper not in self._pre_mapping:
+            self._pre_mapping.append(mapper)
```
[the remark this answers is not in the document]
CoronRing: @soulFood5632 Copilot noticed the same issue, but I guess this will be paused until we figure out a better sequence management system (fifo or others). However I do think modified mapper should be visible to user, whats worse than modifying user added mapper is now tell them its modified. 

But add_pre_mapper( may not be the intended interface anyways. Just placeholder

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
```diff
@@ -158,13 +161,22 @@ def llm(
...
         casted_instance._node_name = name
         casted_instance._node_class = "Agent"
 
+        # Context injection is the default behaviour for all LLM nodes: rt.context
+        # variables are filled into prompt templates before the model call. It is
+        # wired in as a gateway pre-mapper here so it applies regardless of which
+        # higher-level wrapper (agent_node, etc.) created the node. Disable per-node
+        # with context_injection=False, or session-wide via prompt_injection=False.
+        if context_injection:
+            model_gateway.add_pre_mapper(context_injection_gateway_pre_mapper)
```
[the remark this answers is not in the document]
CoronRing: @soulFood5632 This is true, if the user really defined and reuses, it's an issue. No idea how to address this, sees more live a middleware design pattern we need to discuss

CoronRing:
Some thoughts on gate management:
1. Create a custom gateway management object for internal use. User can pass a list or GatewayManager in the model_gateway (<- need to rename to gateway to be generic) field. If a user passes a list, it gets converted to this internally. 
2. GatewayManager have the following property:
  - All is the same as List/iterable on the interface for user. 
  - However, internal gates are stored as a 2d array, 
  - User passed object stored as second array [[SYS]. [USER], [SYS]]
  - When user only accesses and modify the 1 to n-1 array, the first (0) and last (n) layers are managed by the system or other components and not intended to be modified, the order of execution is first to last, left to right; this way, the system always manages the first and last layers. The user-passed object is never modified. 
  
3. This Object can also check through the user passed gateway list to ensure its valid, its almost like gateway wrapper for gateway wrapper =3

CoronRing:
@soulFood5632 This ticket makes less sense to implement alone, I will create a new ticket on gateway and finish gateway + manager with this PR.

soulFood5632:
Sounds good to ne

CoronRing:
It maybe a good idea to rename wrapper_outer to just wrapper while keeping wrapper_inner, so it is easy to understand, and user do not always need the inner feature.

CoronRing:
Potential Name for MiddlewareSet:
1. Pipeline
2. Wrappers
3. Container
4. Surface

For some reason chat do not like gateway (claims that it implies it does not transform) as a name, but it shoulds OK to me.

docs/advanced_usage/node_builder.md
soulFood5632, line 1: The code in this block don't follow the correct style. For the sake of PR lets just remove this docs page so we don't have to mantain it

docs/documentation/advanced/middleware/internals.md
```diff
@@ -0,0 +1,212 @@
...
+  wrappers, and raw functions are ambiguous there, so the decorator is required).
+- `_fresh_copy()` / `copy_user_only()` produce a copy carrying only the **user** layers,
+  with system layers reset — so a `MiddlewareSet` reused across nodes gives each site its
+  own independent system layers and the caller's object is never mutated.
+
+### The execution engine
+
+`run(core, args, kwargs)` composes the onion and awaits it:
+
+```python
```
soulFood5632: code should be in scripts

docs/documentation/advanced/middleware/internals.md
```diff
@@ -0,0 +1,212 @@
...
+## Attach-site integration
+
+### Node middleware
+
+The base `Node` carries a class-level `frozen_middleware: MiddlewareSet` default. Each
+instance takes a fresh copy in `__init__` (`self.middleware = self.frozen_middleware
+._fresh_copy()`), and runs through it via:
+
+```python
+async def wrapped_invoke(self, *args, **kwargs):
```
soulFood5632: ditto

docs/documentation/advanced/middleware/internals.md
```diff
@@ -0,0 +1,212 @@
...
+`wrapped_invoke` is what the executor actually calls, so **every** node type goes through
+its node-level middleware exactly once per call. `NodeBuilder` sets `frozen_middleware`
+from the `middleware=` argument via `MiddlewareSet.coerce`.
+
+### Model middleware
+
+`model_middleware` is owned by `ModelInvoker`, which wraps the *raw* model call:
+
+```python
+async def _core_llm_call(messages, schema, tools):
```
soulFood5632: ditto

docs/documentation/advanced/middleware/usage.md
```diff
@@ -0,0 +1,345 @@
...
+
+---
+
+## `Wrapper` — execution control
+
+A wrapper is an **async** function whose first argument is `call`, the inner
+(already-wrapped) callable. Because the wrapper *owns* the call, it can retry it, time
+it, fall back to an alternative, or skip it entirely.
+
+```python
```
soulFood5632, line 28: ditto

docs/documentation/advanced/middleware/usage.md
soulFood5632, line 1: In general these docs are really verbose. we can in general clean it up a bit. espscially since we haven't even fully defined the use cases as of yet. I would almost prefer we leave these docs out since we have to go through them anyway to polish them for consumption

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
```diff
@@ -69,10 +72,23 @@ def function_node(
...
     name: str | None = None,
     manifest: ToolManifest | None = None,
+    middleware: MiddlewareSet | list | None = None,
 ) -> List[RTFunction[..., Any]]:
     pass
 
 
+@overload
+def function_node(
+    func: None = None,
```
soulFood5632, line 82: I don't understand what this overload is adding

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
```diff
@@ -202,21 +223,56 @@ def function_node(
...
+            parametrized-decorator form, which returns a decorator that takes the function.
         name (str, optional): Human-readable name for the node/tool.
         manifest (ToolManifest, optional): The details you would like to override the tool with.
+        middleware (MiddlewareSet | list | None): Middleware applied around the node boundary.
+            Accepts a MiddlewareSet or a bare list of Wrapper/Gateway (check-only gateways act as guardrails).
     """
 
+    # No function yet -> parametrized-decorator form: bind the options and return
+    # a decorator that finishes the job once the function is supplied.
+    if func is None:
```
soulFood5632, line 254: This is a case we have previously handled so I'm not sure what this adds. I don't think this branch is ever entered.

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
soulFood5632: I can see where is coming from now, but I think this is actually not a use case as the previous overloads handle it.

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
```diff
@@ -132,15 +131,15 @@ def llm(
         name: str,
         class_name: str | None = None,
         *,
-        model_gateway: ModelGateway,
+        model: ModelBase,
```
soulFood5632: So we do have minor design flaw here in some functionality we want to achieve. 

How in the current system would we adjust the llm model to call a different one?

soulFood5632: Maybe we could either allow them to pass in a model or a callable with no args that accesses a model

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
soulFood5632, line 233: do we actually run the middleware on this function properly?

packages/railtracks/src/railtracks/middleware/primitives.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+class _GatewayArgs:
+    """Tagged container returned by :func:`gateway.args` to state an entry gateway's
+    new ``(args, kwargs)`` unambiguously — including multiple positional args and/or
+    keyword args, which a bare return value cannot express."""
+
+    __slots__ = ("args", "kwargs")
+
+    def __init__(self, args: tuple, kwargs: dict) -> None:
```
soulFood5632: kwargs can probably be dict[str, Any]

packages/railtracks/src/railtracks/middleware/primitives.py
```diff
@@ -0,0 +1,268 @@
...
+    loss of the run's context is not acceptable).
+    """
+
+    def __init__(self, fn: WrapperFn) -> None:
+        _require_async(fn, "Wrapper function")
+        self._fn = fn
+        functools.update_wrapper(self, fn, updated=())
+
+    def wrap(
+        self, inner: Callable[..., Awaitable[Any]]
```
soulFood5632: we should use paramspec here instead so the wrapped function outputs are typed to matched the inputs

soulFood5632 (commented):
Some more for me to go through but I thought I would get started here.

docs/advanced_usage/node_builder.md
CoronRing: Yeah, this PR as a whole will be dropped and reformated, including advanced usage and internal doc stuff, its more for us to understand. I did not do cleaning

docs/documentation/advanced/middleware/internals.md
CoronRing: Yup, all the docs is read only, not "merge" able, i got lazy in case it needs rework and I need to update it many times.

packages/railtracks/src/railtracks/built_nodes/llm_helpers.py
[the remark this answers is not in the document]
CoronRing: obsoleted by the GatewayManager refactor — user lists are copied into the managed layers, system gateways live in separate slots

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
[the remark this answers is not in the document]
CoronRing: Same as above, obsoleted by the GatewayManager refactor — user lists are copied into the managed layers, system gateways live in separate slots, close for cleaness

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
CoronRing: The base branch's `function_node` requires `func` positionally, the parametrized decorator form `@rt.function_node(middleware=[guard])` didn't exist. Also for comment below,  `function_node(func=None, ...)`, which is what the `None` overload types and the `if func is None` branch implements

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
CoronRing: `test_function_node_middleware_runs` in `tests/unit_tests/middleware/test_node_middleware.py` was added to check this execution

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
CoronRing: Good point, will widening the `NodeBuilder.llm()` / `agent_node()` annotations to expose it

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
CoronRing: execution goes through `node.wrapped_invoke` (`execution/task.py:37`), which runs node middleware around `invoke` (`nodes/nodes.py:125`). Nothing covered this before,  adding `tests/unit_tests/middleware/test_node_middleware.py` with a tracing wrapper to cover

packages/railtracks/src/railtracks/middleware/primitives.py
CoronRing: Updated, including otehr palced

packages/railtracks/src/railtracks/middleware/primitives.py
CoronRing: Good point, updated to match

CoronRing: `wrap` now carries the wrapped function's ParamSpec and return type through, so composing middleware no longer erases the core callable's signature. 
 
`WrapperFn` itself ( `fn(call, *args, **kwargs)`) stays `Callable[..., Coroutine]`. Typing it fully needs `Concatenate[Callable[_P, Awaitable[_R]], _P]`, which would force every user-authored wrapper to be generic over the wrapped signature, perhaps another ticket, but updated docstring

CoronRing:
@soulFood5632 Renamed wrapper_outer to wrappers (cannot name as wrapper already ownd the name, middleware is still to discuss.

docs/documentation/advanced/middleware/internals.md
soulFood5632: Sounds good.

soulFood5632 (approved):
LGTM, let's just merge and keep it in the feature branch for now. we can always make adjustments need be.

CoronRing:
@soulFood5632 Merge after some final docs update

PR #1151, opened 2026-06-01
1132 tech debt backend node improvement to prep for generalized node enhancements

soulFood5632:
Some things (reverse design doc generated below)
# PR 1132: Node Builder Unification — Maintainer Summary

**Branch**: `1132-tech-debt-backend-node-improvement-to-prep-for-generalized-node-enhancements`
**Scope**: Structural refactor with some breaking changes to public exports and internal APIs.
**Why it matters**: This is PR 1 of a multi-PR overhaul. It lays the foundation for composable middleware (retry, guardrails, caching, tracing) without touching node internals.

**Breaking changes in this PR**:
- `RTSyncFunction`, `RTAsyncFunction`, `SyncDynamicFunctionNode`, `AsyncDynamicFunctionNode`, `DynamicFunctionNode` removed from `concrete/__init__.py` exports — use `RTFunction` instead
- `NodeBuilder` API is completely replaced (old API took a concrete base class; new API uses `.llm()` / `.function()` factory classmethods)
- `Node.tracked_invoke()` renamed to `Node.wrapped_invoke(*args, **kwargs)` — args are now passed explicitly rather than stored on the instance
- LLM-as-tool argument formatting changed: parameters are now formatted as a single coherent instruction message instead of separate messages (behavioral improvement, but observable change in message history)

---

### Function nodes: two redundant parallel classes

Before this PR, sync and async functions required separate concrete classes:

```
@rt.function_node          @rt.function_node (async)
        │                           │
SyncDynamicFunctionNode    AsyncDynamicFunctionNode
  own __init__               own __init__
  own invoke()               own invoke()
  RTSyncFunction             RTAsyncFunction   ← two separate exported types
```

### LLM nodes: combinatorial class explosion

Every combination of output type × capability got its own concrete class:

```
Output type  │  No tools       │  With tools           │  With guardrails
─────────────┼─────────────────┼───────────────────────┼────────────────────────────
String        │  TerminalLLM    │  ToolCallLLM           │  GuardedTerminalLLM
              │                 │                        │  GuardedStreamingTerminalLLM
              │                 │                        │  GuardedStreamingToolCallLLM
              │                 │                        │  GuardedToolCallLLM
─────────────┼─────────────────┼───────────────────────┼────────────────────────────
Structured    │  StructuredLLM  │  StructuredToolCallLLM │  GuardedStructuredLLM
              │                 │  ChatToolCallLLM       │  GuardedStreamingStructuredLLM
[...]

PR #1169, opened 2026-06-04
pooria/1154 reimplement mcp

Pooria90:
## Summary

Updates `rt_mcp` for compatibility with the NodeBuilder unification (PR #1151) and replaces the deprecated `streamablehttp_client` with the current MCP SDK API.

**NodeBuilder unification (`MCPToolNode`):**
- Removed `__init__(**kwargs)` / `self.kwargs` — the old "node-as-factory" pattern where args were baked into the instance at construction time. Under the new system `_create_node_and_request` calls `node()` with no args; invocation args arrive via `wrapped_invoke(**kwargs)`
- Changed `invoke(self)` → `async def invoke(self, **kwargs)` so args are received explicitly, matching the new `Node` base contract
- Replaced blocking `future.result()` with `await asyncio.wrap_future(future)` — non-blocking bridge between the caller's event loop and the MCP background loop
- Removed `prepare_tool` — the tool invocation path in `llm_helpers.py` now uses `prepare_args` (inherited passthrough from `Node`)
- Fixed parameter type: `tool: Tool` → `tool: MCPTool` (`mcp.types.Tool` vs `railtracks.llm.Tool` name collision)
- Replaced the manual `MCPToolNode(Node)` class definition with `NodeBuilder.function()`, eliminating ~30 lines of boilerplate

**`NodeBuilder.function()` — added `tool_info` param:**
- Added `tool_info: Tool | None = None` to `NodeBuilder.function()` for callers that already have a fully-constructed `Tool` (e.g. converted from an external schema). When provided it bypasses `Tool.from_function()` entirely; otherwise the existing inference path is unchanged
- `from_mcp` now passes `Tool.from_mcp(tool)` through this param instead of monkey-patching `builder._tool_info` after construction

**`streamablehttp_client` deprecation:**
- Replaced with `streamable_http_client`. The new API takes an `httpx.AsyncClient` instead of raw `headers`/`timeout`/`auth` kwargs; `httpx.Timeout` maps both `timeout` and `sse_read_timeout` from `MCPHttpParams`
- Added `auth: httpx.Auth | None = None` to `MCPHttpParams` (was referenced via `hasattr` guard but never declared)
- Added `assert isinstance(self.config, MCPHttpParams)` in `_init_http` and `assert self.session is not None` in `list_tools`/`call_tool` for type narrowing

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

`prepare_tool` removal is technically a breaking change for any caller constructing `MCPToolNode` instances manually, but that was never part of the public API — `connect_mcp` → `server.tools` → `rt.call(ToolNode, **kwargs)` is the only documented path and is unaffected.

Verified end-to-end with the Filesystem MCP server: 14 tools discovered, all 11 exercised tool calls pass.

### DIFF
--- a/packages/railtracks/src/railtracks/rt_mcp/main.py
+++ b/packages/railtracks/src/railtracks/rt_mcp/main.py
@@ -427,6 +427,9 @@ def from_mcp(
                 f"Tool invocation failed: {type(e).__name__}: {str(e)}"
             ) from e
 
-    builder = NodeBuilder.function(invoke, class_name=tool.name, name=tool.name)
-    builder._tool_info = lambda: Tool.from_mcp(tool)
-    return builder.build()
+    return NodeBuilder.function(
+        invoke,
+        class_name=tool.name,
+        name=tool.name,
+        tool_info=Tool.from_mcp(tool),
+    ).build()
--- a/packages/railtracks/src/railtracks/rt_mcp/main.py
+++ b/packages/railtracks/src/railtracks/rt_mcp/main.py
@@ -4,12 +4,14 @@ from contextlib import AsyncExitStack
 from datetime import timedelta
 from typing import Any
 
+import httpx
 from mcp import ClientSession, StdioServerParameters
+from mcp.types import Tool as MCPTool
 from mcp.client.sse import sse_client
 from mcp.client.stdio import stdio_client
-from mcp.client.streamable_http import streamablehttp_client
+from mcp.client.streamable_http import streamable_http_client
 from pydantic import BaseModel
-from typing_extensions import Self, Type
+from typing_extensions import Type
 
 from railtracks.llm import Tool
 from railtracks.nodes.nodes import Node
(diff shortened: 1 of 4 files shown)


packages/railtracks/src/railtracks/rt_mcp/main.py
soulFood5632, line 1: I don't think I was clear enough in the tikcet, but you should try and use the new nodebuilder style approach as seen for `nodebuilder.function()` instead of creating subtype for node

soulFood5632 (changes requested):
See comment, my bad on the explanations

packages/railtracks/src/railtracks/rt_mcp/main.py
Pooria90: Please check out the new code and see if it fits the new style.

packages/railtracks/src/railtracks/rt_mcp/main.py
```diff
@@ -374,104 +390,43 @@ def tools(self) -> list[Type[Node]]:
...
+                client.call_tool(tool.name, kwargs), loop
+            )
+            return await asyncio.wrap_future(future)
+        except Exception as e:
+            raise RuntimeError(
+                f"Tool invocation failed: {type(e).__name__}: {str(e)}"
+            ) from e
+
+    builder = NodeBuilder.function(invoke, class_name=tool.name, name=tool.name)
+    builder._tool_info = lambda: Tool.from_mcp(tool)
```
soulFood5632: probably better if we pass the tool details into the `function` method. if there is no way to override we should add that logic into `function`

soulFood5632 (changes requested):
Yea looks great and clean. One comment about accessing protected attr

packages/railtracks/src/railtracks/rt_mcp/main.py
Pooria90: I added an argument in `.function`.  If it's None, the method uses details and params.

soulFood5632 (approved):
LGTM

PR #1174, opened 2026-06-08
Retrieval loader fixes: per-row Document IDs + JSONL support

Amir-R25:
## Summary

- **Per-row Document IDs for row-based loaders.** `CSVLoader`, `JSONLoader`, and `HuggingFaceDatasetLoader` previously set `Document.source` to the file/dataset path, so every row in a file produced the same `Document.id` breaking upsert (`delete_where` on `document_id`). Source is now `"{path}#{row_id}"`, where `row_id` comes from a new `id_column` / `id_key` parameter, falling back to the row's positional index.
- **JSONL support in `JSONLoader`.** `.json` and `.jsonl` are both accepted (single file or directory). JSONL is parsed line by line in a worker thread, so corpora larger than memory stream cleanly. New `DocumentType.JSONL` and `.jsonl → JSONL` mapping in the cloud loader common table.
- **`HuggingFaceDatasetLoader` cleanup.** `streaming=True` is now always set (warns if passed via `dataset_kwargs`); `IterableDatasetDict` is unwrapped to the requested split with a clear error if missing; sentinel is a typed class so `isinstance` checks don't snag on dict rows. Docs gain a warning about the upstream `datasets`/`pyarrow` early-exit hang ([huggingface/datasets#8176](https://github.com/huggingface/datasets/pull/8176)) with the `gc.collect()` workaround.
- **Docs + examples** updated for JSONL and the new id parameter plus a new notebook in the tutorials.

## Test plan

- [ ] `pytest packages/railtracks/tests/unit_tests/retrieval/loaders/` passes
- [ ] Ingest a CSV/JSON/HF dataset twice with `id_column` set; confirm upsert replaces rows instead of duplicating
- [ ] Ingest a `.jsonl` file larger than RAM; confirm streaming (no OOM) and per-line `Document.source` suffix
- [ ] Ingest a directory mixing `.json` and `.jsonl`; confirm both are picked up

### DIFF
--- a/docs/retrieval/components/ingestion/base.md
+++ b/docs/retrieval/components/ingestion/base.md
@@ -28,7 +28,7 @@ to something stable.
 |---|---|---|
 | `TextLoader` | `.txt`, `.md` files & directories | None |
 | `CSVLoader`  | `.csv` files & directories | None |
-| `JSONLoader` | `.json` files & directories | None |
+| `JSONLoader` | `.json` and `.jsonl` files & directories | None |
 | `PyPDFLoader` | `.pdf` files & directories (embedded text only) | `pip install "railtracks[pdf]"` |
 | `PyPDFOCRLoader` | `.pdf` files & directories with OCR fallback | `pip install "railtracks[ocr]"` + [Tesseract](https://pypi.org/project/pytesseract/) |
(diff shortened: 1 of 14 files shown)


soulFood5632 (approved):
LGTM

PR #1210, opened 2026-06-18
1208 pdf support and dependabot fixes

### LINKED ISSUE #1208: [Feature] Add `.pdf` as a supported file extension for attachments
### Feature Request

As explained in the title, openai supports it, need to make sure if other models/providers support it as well

### Additional Context

_No response_

Amir-R25:
## Summary

Adds PDF attachment support to `rt.llm.UserMessage`. Previously a base64 PDF or `.pdf` path raised in `Attachment.__init__` because `detect_source` only knew image magic bytes and the extension map was image-only.

- `attachment_formats` (renamed from `image_formats`): adds `%PDF` to the magic-byte table.
- `encoding`: `_is_base64_attachment`, `_validate_data_uri_header`, and `ensure_data_uri` accept `application/pdf` data URIs and bare base64 payloads. `encode()` takes a `timeout` for URL fetches.
- `Attachment`: stores `mime_type` and `filename`, derives `modality` (`image` or `document`) across local files, URLs, and data URIs. `.pdf` is a supported local extension. Data-URI MIME is lowercased before modality lookup so mixed-case headers don't silently fall through to `image`.
- `_to_litellm_message`: documents serialize as `{"type": "file", "file": {"file_data", "filename"}}` (unified litellm shape for OpenAI / Anthropic native PDF input); images keep `image_url`. Rejects PDF attachments early when `litellm.utils.supports_pdf_input` is false.

### URL PDFs are opt-in (`trust_urls=True`)

Unlike image URLs (which the provider fetches), PDFs are passed inline as base64, so railtracks downloads them in-process. That's an SSRF surface for end-user-supplied URLs, so the path is gated:

- `.pdf` URL without `trust_urls=True` → raises with a pointer to download out-of-band.
- Unknown-extension URL without `trust_urls=True` → no HEAD probe, handed to the provider as `image_url` as today.
- With `trust_urls=True`: HEAD-probe unknown extensions, download PDFs into the prompt.

`attachment_timeout` (default 10s) covers both the HEAD probe and the PDF download.

### Other

- Bumps `litellm` to `>= 1.84.0, <= 1.89.0` for **GHSA-4xpc-pv4p-pm3w / CVE-2026-49468** (Host-header auth bypass in the litellm proxy).
- Re-encodes `uv.lock` from UTF-16 LE / CRLF to UTF-8 / LF so `uv` can read it.

Closes #1208.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [x] Lint & format pass
- [x] Docs updated (`modalities.md` covers the PDF path and the `trust_urls` warning)

## Notes

Provider coverage: litellm's `file` block works for OpenAI (`gpt-4o` / `gpt-5.x`) and Anthropic Claude. Other providers will 400 — render PDF pages to PNGs for those.

```python
rt.llm.UserMessage(
    content="What is this PDF?",
    attachment="1706.03762v7.pdf",
)

rt.llm.UserMessage(
    content="Summarize this.",
    attachment="https://arxiv.org/pdf/1706.03762",
    trust_urls=True,
    attachment_timeout=30,
)

### CODE REVIEW

File: docs/documentation/agent_design/llms/modalities.md
Comment (Pooria90) [inline docs/documentation/agent_design/llms/modalities.md]: Nice to have a sample image path in here; and also moving to docs/scripts for consistency.

File: docs/documentation/agent_design/llms/modalities.md [hunk 1]
```diff
@@ -78,4 +78,37 @@ rt.llm.UserMessage(
...
+- HTTPS URLs ending in `.pdf` — **only with `trust_urls=True`** (see below)
+
+!!! warning "URL PDFs require `trust_urls=True`"
+    Unlike image URLs (which the provider fetches), PDFs are passed to the provider inline as base64. That means railtracks downloads the URL **in-process** before the call. End-user-supplied URLs are an SSRF surface once their bytes are fetched on your server and embedded in the prompt, so this path is opt-in:
+
+    ```python
+    --8<-- "docs/scripts/multimodal.py:pdf_url"
+    ```
+
+    Set `trust_urls=True` only when every URL in `attachment` is developer-controlled (a hard-coded path, an internal doc, a trusted CDN). For user-supplied URLs, download the PDF out-of-band first and pass the local path or base64 payload instead.
```
Pooria90: I think this is redundant as something similar is commented in the code block.

docs/scripts/multimodal.py
```diff
@@ -41,6 +41,48 @@
...
+    [
+        rt.llm.UserMessage(
+            content="Summarize the key findings in this paper.",
+            # attachment="path/to/local/paper.pdf",  # Uncomment and provide a valid local PDF path
+        )
+    ]
+)
+# --8<-- [end: pdf_local]
+
+# --8<-- [start: pdf_url]
```
Pooria90, line 55: A comment on this in the previous file

Pooria90 (changes requested):
Good job Amir. I just have minor comments on documentations.

Pooria90 (approved):
Image

PR #1212, opened 2026-06-22
pooria/1205 memory tool

### LINKED ISSUE #1205: [Feature] Memory Tool
### Feature Request

It would be nice to build on some of the work in #1203. Build a persistent memory toolset for simple:
- key value storage
- free form embedding storage. 
- Simple knowledge graph 

For each of those layers we should provide a backend of choice using the vector stores we provide. 

### Additional Context

_No response_

Pooria90:
## Summary

Adds a prebuilt key-value memory toolset for agents — the memory analog of `ToDoToolSet`. Introduces a reusable, stdlib-only async key-value store under `railtracks.retrieval.stores.key_value` (`KeyValueStore` protocol + `InMemoryKeyValueStore`, with optional JSON `snapshot_path` persistence), kept deliberately separate from the vector `Store` protocol since it's exact-key get/set with no similarity/embeddings. On top of it, `rt.prebuilt.KeyValueMemoryToolSet` exposes six agent tools (`remember`, `recall`, `forget`, `list_keys`, `list_memories`, `search_memories`), an `on_change` callback for outer-system hooks, and a `prompt()` guidance helper. The default store is lazy-imported via a private helper so `import railtracks` stays free of the retrieval stack (numpy/scikit-learn); `[all]` now bundles `[retrieval]` so the standard install and CI exercise the toolset. This is the first (key-value) layer of the persistent memory toolset.

closes #1205

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

## Notes

**Extended example — a memory-backed tour-guide assistant.** Learns about a traveler once and personalizes every later answer; the persistent store means what it learns survives across runs.

```python
import railtracks as rt
from railtracks.llm import AssistantMessage, UserMessage
from railtracks.prebuilt import KeyValueMemoryToolSet
from railtracks.retrieval.stores.key_value import InMemoryKeyValueStore

# Persistent store — traveler facts survive across runs.
memory = KeyValueMemoryToolSet(
    store=InMemoryKeyValueStore(snapshot_path="memory.json"),
)

# Persona + the toolset's own usage guidance + app-specific key conventions.
SYSTEM_PROMPT = (
    "You are a tour-guide assistant. You help users plan travel, hikes, and "
    "day trips.\n\n"
    + memory.prompt()
    + "\n\nStore durable traveler facts under separate, stable keys — e.g. "
    "home_city, dietary_restrictions, fitness_level, budget, travel_dates, "
    "destinations_of_interest. Recall the relevant keys before recommending "
    "so every answer stays personalized."
)

agent = rt.agent_node(
    name="Tour-Guide",
    system_message=SYSTEM_PROMPT,
    llm=rt.llm.AnthropicLLM("claude-sonnet-4-6"),
    tool_nodes=[*memory.tool_set()],
)


@rt.function_node
async def conversation():
    history = []
    for user_text in [
        "I'm vegetarian, based in Vancouver, and I love hiking.",
        "Suggest a weekend trip for me.",  # answered from memory, no re-asking
    ]:
        history.append(UserMessage(user_text))
        response = await rt.call(agent, history)
        history.append(AssistantMessage(response.content))
        print(f"User: {user_text}\nAssistant: {response.content}\n")

    # snapshot_path means these facts are also on disk for the next run.
    print("Stored memory:", await memory.store.items())


rt.Flow("Tour-Guide-Chat", entry_point=conversation).invoke()

### DIFF
--- a/packages/railtracks/src/railtracks/prebuilt/__init__.py
+++ b/packages/railtracks/src/railtracks/prebuilt/__init__.py
@@ -1,6 +1,8 @@
 ######## This package contains a suite of pre-built ready to use agents designed to help you build faster #########
+from .tools.memory import KeyValueMemoryToolSet
 from .tools.todos import ToDoToolSet
 
 __all__ = [
+    "KeyValueMemoryToolSet",
     "ToDoToolSet",
 ]
--- a/packages/railtracks/src/railtracks/prebuilt/tools/memory/key_value.py
+++ b/packages/railtracks/src/railtracks/prebuilt/tools/memory/key_value.py
@@ -121,6 +121,21 @@ class KeyValueMemoryToolSet(ToolSet):
             return "No memories stored."
         return "\n".join(f"- {key}: {value}" for key, value in items.items())
 
+    async def list_keys(self) -> str:
+        """List just the keys currently held in memory, without their values.
+
+        Prefer this over list_memories() to see what is stored without pulling
+        every value into context; then recall(key) only the ones you need.
+
+        Returns:
+            A newline-separated list of keys, or a message saying memory is
+            empty.
+        """
+        keys = await self.store.keys()
+        if not keys:
+            return "No memories stored."
+        return "\n".join(f"- {key}" for key in keys)
+
     async def search_memories(self, query: str) -> str:
         """Search stored memories for a substring across keys and values.
 
@@ -151,7 +166,8 @@ class KeyValueMemoryToolSet(ToolSet):
             "Call remember(key, value) to save a fact under a short, stable, descriptive key; "
             "re-calling remember() with the same key overwrites the old value. "
             "Call recall(key) to read a fact back, and forget(key) to delete one. "
-            "If you are unsure of the exact key, call search_memories() to find by substring "
+            "If you are unsure of the exact key, call search_memories() to find by substring, "
+            "list_keys() to see what is stored without pulling in every value, "
             "or list_memories() to see everything stored. "
             "Save anything the user tells you that may be useful later (preferences, names, goals, "
             "constraints), and recall before asking the user to repeat themselves."
@@ -162,6 +178,7 @@ class KeyValueMemoryToolSet(ToolSet):
             self.remember,
             self.recall,
             self.forget,
+            self.list_keys,
             self.list_memories,
             self.search_memories,
         ]
--- a/packages/railtracks/src/railtracks/prebuilt/tools/memory/key_value.py
+++ b/packages/railtracks/src/railtracks/prebuilt/tools/memory/key_value.py
@@ -1,0 +1,168 @@
+from __future__ import annotations
+
+from typing import Callable
+
+import railtracks as rt
+from railtracks.built_nodes.concrete.function_base import RTFunction
+from railtracks.retrieval.stores.key_value import (
+    InMemoryKeyValueStore,
+    KeyValueStore,
+)
+from railtracks.utils.logging.create import get_rt_logger
+
+from .._base import ToolSet
+
+logger = get_rt_logger(__name__)
+
+
+class KeyValueMemoryToolSet(ToolSet):
+    """Prebuilt key-value memory tools for an agent.
+
+    Gives an agent a persistent, exact-match scratch pad: save a fact under a
+    key, read it back later, forget it, list everything, or search. State lives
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: packages/railtracks/src/railtracks/prebuilt/tools/memory/key_value.py [hunk 1]
```diff
@@ -0,0 +1,188 @@
...
+
+from typing import TYPE_CHECKING, Callable
+
+import railtracks as rt
+from railtracks.built_nodes.concrete.function_base import RTFunction
+from railtracks.utils.logging.create import get_rt_logger
+
+from .._base import ToolSet
+
+if TYPE_CHECKING:
```
soulFood5632, line 11: Good management of importing here. it is slightly akward to have to have import a vector store idea into the core but it does have to happen here

packages/railtracks/src/railtracks/prebuilt/tools/memory/key_value.py
```diff
@@ -0,0 +1,188 @@
...
+        Args:
+            query: Substring to look for in keys and values.
+
+        Returns:
+            Matching "key: value" entries, or a message saying nothing matched.
+        """
+        needle = query.casefold()
+        items = await self.store.items()
+        matches = [
+            f"- {key}: {value}"
```
soulFood5632, line 157: I'll add an issue here do some sort of vector search capability here. won't be used in most cases but it could come up

packages/railtracks/src/railtracks/prebuilt/__init__.py
```diff
@@ -1,6 +1,8 @@
 ######## This package contains a suite of pre-built ready to use agents designed to help you build faster #########
+from .tools.memory import KeyValueMemoryToolSet
 from .tools.todos import ToDoToolSet
 
 __all__ = [
```
soulFood5632, line 5: I thini this makes sense for now but I did want to ask, do you think importing:
```
rt.prebuilt.tools.MemoryToolSet()
# or 
rt.prebuilt.MemoryToolSet() 
```

I'm split on it. I think i cleanness of option 2 but I would be curious to hear your take

packages/railtracks/src/railtracks/retrieval/stores/key_value/protocol.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
+
+from typing import Protocol, runtime_checkable
+
+
+@runtime_checkable
+class KeyValueStore(Protocol):
```
soulFood5632, line 7: I'll add as a future option to add different types of backend strores.

soulFood5632 (approved):
LGTM

packages/railtracks/src/railtracks/prebuilt/tools/memory/key_value.py
soulFood5632: I'll deal with this

packages/railtracks/src/railtracks/prebuilt/tools/memory/key_value.py
Pooria90: The stores don't have to be *vector* stores. I can make a few changes like making a separate store protocol for each type and structure them as separate submodules in a `stores` module, taking some of the attention off the vector stores. Something like this:
store
|___ kv
|___ vector
|___ graph
...
I'd appreciate your input on this.

packages/railtracks/src/railtracks/prebuilt/__init__.py
Pooria90: Same. I think second one is better as the module classes already contain a `ToolSet`.

packages/railtracks/src/railtracks/retrieval/stores/key_value/protocol.py
Pooria90: I'll make an issue.

packages/railtracks/src/railtracks/prebuilt/tools/memory/key_value.py
soulFood5632: Yea that was my thought as well. I like that breakdown you provided and we should implement a couple of examples. I'm thinking some common examples like redis might be good start in addition to the InMemory stuff for key values. 

i like the idea of a base class for each type of store.

PR #1213, opened 2026-06-22
Typing Middleware

soulFood5632:
## Summary

I clairfied and fixed up some typing in the middlware module. The only remaining hole is the output types of input gateways. Those remain a `tuple[tuple, dict[str, Any]]`. everything else is typed up nice, please see below examples and play around with it yourself
## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [x] Breaking changes include migration notes

## Notes

Some cool typing stuff

incorrect input types for a input gateway
image

correct type

image

mixing up inputs and outputs

image

coereced type in exit gateway
image

correct exit gateway
image

correct typing for wrapper
image

```python


from curses import wrapper
from typing import Awaitable, Callable

import railtracks as rt

@rt.function_node
def hello_world(suffix: str) -> str:
    return f"hello world {suffix}"

@rt.gate
def log_inputs(*args, **kwargs):
    print(f"Gateway received inputs: args={args}, kwargs={kwargs}")
    return 


@rt.gate
def add_gateway(x: int, y: int):
    if x < 0 or y < 0:
        raise ValueError("Error: Negative numbers are not allowed.")
    
    return (x, y), {}

@rt.gate
async def add_gateway_exit(result: int):
    return result * 2

@rt.wrapper
async def half_fail(call: Callable[[int, int], Awaitable[int]], x: int, y: int) -> int:
        import random

        if random.random() < 0.5:
            raise Exception("Random failure occurred.")
        return await call(x, y)


@rt.wrapper
async def retry(call, *args, **kwargs):
    max_retries = 3
    for attempt in range(max_retries):
        try:
        except Exception as e:
            if attempt == max_retries - 1:
                raise e
            else:
                print(f"Attempt {attempt + 1} failed: {e}. Retrying...")
        
    raise Exception("All retry attempts failed.")



@rt.function_node(middleware=rt.MiddlewareChain(wrappers=[retry, half_fail], entry_gate=[add_gateway, log_inputs], exit_gate=[add_gateway_exit]))
async def add(x: int, y: int) -> int:
    return x + y
```

code to play with

### DIFF
--- a/docs/advanced_usage/node_builder.md
+++ b/docs/advanced_usage/node_builder.md
@@ -10,7 +10,7 @@ Every builder method returns a builder; call `.build()` to get the node class.
 !!! note "Middleware lives in the middleware guide"
     `NodeBuilder.llm()` and `NodeBuilder.function()` take the same `middleware` /
     `model_middleware` parameters as the public helpers. See
-    [Middleware: Wrappers & Gateways](../documentation/advanced/middleware/usage.md) for
+    [Middleware: Wrappers & Gates](../documentation/advanced/middleware/usage.md) for
     the full model — this page only shows where they attach.
 
 ---
@@ -69,7 +69,7 @@ Sync functions are not supported directly — wrap them with `asyncio.to_thread`
 
 Both builders accept `middleware` (around the node boundary, `wrapped_invoke`). LLM nodes
 additionally accept `model_middleware` (around each raw model call, inside the
-tool-calling loop). Each takes a `MiddlewareSet` or a bare list:
+tool-calling loop). Each takes a `MiddlewareChain` or a bare list:
 
 ```python
 --8<-- "docs/scripts/node_builder.py:middleware"
--- a/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
+++ b/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
@@ -96,6 +96,9 @@ def safe_create_node(
     return type(class_name + "Node", (Node,), class_dict)
 
 
+
+
+
 class NodeBuilder(Generic[_P, _T]):
     def __init__(self) -> None:
         self._class_name: str | None = None
@@ -226,7 +229,7 @@ class NodeBuilder(Generic[_P, _T]):
         class_name: str | None = None,
         name: str | None = None,
         *,
-        middleware: MiddlewareSet | list | None = None,
+        middleware: MiddlewareSet[_P2, _T2] | None = None,
         tool_details: str | None = None,
         tool_params: list[Parameter] | Type[BaseModel] | dict[str, Any] | None = None,
         tool_info: Tool | None = None,
--- a/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
+++ b/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
@@ -96,9 +96,6 @@ def safe_create_node(
     return type(class_name + "Node", (Node,), class_dict)
 
 
-
-
-
 class NodeBuilder(Generic[_P, _T]):
     def __init__(self) -> None:
         self._class_name: str | None = None
@@ -126,7 +123,10 @@ class NodeBuilder(Generic[_P, _T]):
         tool_details: str | None = None,
         tool_params: list[Parameter] | None = None,
         middleware: MiddlewareSet[[UserInput], StringResponse] | None = None,
-        model_middleware: MiddlewareSet[[MessageHistory, type[BaseModel] | None, list[Tool] | None], Response] | None = None,
+        model_middleware: MiddlewareSet[
+            [MessageHistory, type[BaseModel] | None, list[Tool] | None], Response
+        ]
+        | None = None,
         context_injection: bool = True,
     ) -> NodeBuilder[[UserInput], StringResponse]: ...
 
@@ -143,8 +143,12 @@ class NodeBuilder(Generic[_P, _T]):
         connected_nodes: Iterable[Type[Node]] | None = None,
         tool_details: str | None = None,
         tool_params: list[Parameter] | None = None,
-        middleware: MiddlewareSet[[UserInput], StructuredResponse[_TStructured]] | None = None,
-        model_middleware: MiddlewareSet[[MessageHistory, type[BaseModel] | None, list[Tool] | None], Response]  | None = None,
+        middleware: MiddlewareSet[[UserInput], StructuredResponse[_TStructured]]
+        | None = None,
+        model_middleware: MiddlewareSet[
+            [MessageHistory, type[BaseModel] | None, list[Tool] | None], Response
+        ]
+        | None = None,
         context_injection: bool = True,
     ) -> NodeBuilder[[UserInput], StructuredResponse[_TStructured]]: ...
 
@@ -161,7 +165,10 @@ class NodeBuilder(Generic[_P, _T]):
         tool_details: str | None = None,
         tool_params: list[Parameter] | None = None,
         middleware: MiddlewareSet[[UserInput], _R] | None = None,
-        model_middleware: MiddlewareSet[[MessageHistory, type[BaseModel] | None, list[Tool] | None], Response]  | None = None,
+        model_middleware: MiddlewareSet[
+            [MessageHistory, type[BaseModel] | None, list[Tool] | None], Response
+        ]
+        | None = None,
         context_injection: bool = True,
     ) -> NodeBuilder[[UserInput], _R]:
         instance = cls()
@@ -179,7 +186,7 @@ class NodeBuilder(Generic[_P, _T]):
         # context_injection=False, or flow/session-wide via prompt_injection=False.
         if context_injection:
             model_invoker.register_sys_gateway_entry(context_injection_gateway)
-        
+
         model_invoker.register_sys_wrapper(llm_observe)
 
         casted_instance._invoke = llm_invoke_factory(
--- a/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
+++ b/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
@@ -35,7 +35,7 @@ from railtracks.llm.history import MessageHistory
 from railtracks.llm.message import Message
 from railtracks.llm.response import Response
 from railtracks.llm.type_mapping import TypeMapper
-from railtracks.middleware import MiddlewareSet
+from railtracks.middleware import MiddlewareChain
 from railtracks.nodes.nodes import Node
 from railtracks.prompts.prompt import context_injection_gateway
 from railtracks.validation.node_creation.validation import (
@@ -107,7 +107,7 @@ class NodeBuilder(Generic[_P, _T]):
         self._tool_info: Callable[[], Tool] | None = None
         self._prepare_arguments: Callable[..., dict[str, Any]] | None = None
 
-        self._frozen_middleware: MiddlewareSet = MiddlewareSet()
+        self._frozen_middleware: MiddlewareChain = MiddlewareChain()
 
     @overload
     @classmethod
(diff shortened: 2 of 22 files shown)


packages/railtracks/src/railtracks/built_nodes/llm_helpers.py
```diff
@@ -394,3 +434,28 @@ def prepare_string_response(
...
+@wrapper
+async def llm_observe(
+    call: Callable[
+        [MessageHistory, type[BaseModel] | None, list[Tool] | None], Awaitable[Response]
+    ],
+    message_history: MessageHistory,
+    schema: type[BaseModel] | None,
+    tools: list[Tool] | None,
+) -> Response:
+    prev_message_history = deepcopy(message_history)
```
CoronRing, line 448: Not super sure why a deepcopy is required since we are just logging?

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
```diff
@@ -49,7 +49,7 @@ def function_node(
     *,
     name: str | None = None,
     manifest: ToolManifest | None = None,
-    middleware: MiddlewareSet | list | None = None,
```
CoronRing, line 52: `MiddlewareSet.coerce()` still accepts a bare list and the implementation overload still uses the unparameterized `MiddlewareSet | None.

The support for the list was somewhat legacy carried over before middleware combined gateway_entry and gateway_exit togethe. The initial design was that user can pass a list gateway for each. After combined, the logic is that user can pass in gateway_entry or wrapper in a single list for simplicity. I am ok removing it but the docs/code for parsing a list may need to be removed/updated as well


This was listed as a point of discussion, whether we wanted
1. MiddlewareSet (list makes less sense) or 
``` python
rt.function_node(
   middleware=rt.MiddlewareSet( # New type here
        wrappers=[trace_node],
        gateway_entry=[shout_in],
        gateway_exit=[check_out],
    ),
```
2. _entry + _exit + wrapper (list makes total sense).

``` python
rt.function_node(
        wrappers=[trace_node],
        gateway_entry=[shout_in],
        gateway_exit=[check_out],
    ),
```
But it was dropped to get the PR in faster. 

FMP the second seems easier to use/understand, without requiring the definition of a new custom type (similar to the declarator). But MiddlewareSet make the entire block easily transferable (ie attach to different things).

This was why the current approach is a weird middle ground: 

``` python
rt.function_node(
   middleware=[trace_node, shout_in] # but check_out is not supported 
```

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
```diff
@@ -119,8 +122,11 @@ def llm(
         connected_nodes: Iterable[Type[Node]] | None = None,
         tool_details: str | None = None,
         tool_params: list[Parameter] | None = None,
-        middleware: MiddlewareSet | list | None = None,
-        model_middleware: MiddlewareSet | list | None = None,
+        middleware: MiddlewareSet[[UserInput], StringResponse] | None = None,
```
CoronRing: I think the node itself already enforces this. If the middleware is actually a translator (ie from a custom dict->UserInput), this will be violated as wrappers run first.

packages/railtracks/src/railtracks/middleware/primitives.py
```diff
@@ -266,13 +230,28 @@ async def reorder(a, b):
...
+def gateway(fn: Callable[_P, _R]) -> Gateway[_P, _R]: ...
 
-    See :meth:`Gateway.apply_entry` for how an entry gateway's return is interpreted;
-    use :func:`gateway.args` to specify multiple positional/keyword args explicitly."""
+
+@overload
+def gateway(fn: Callable[..., Any]) -> Gateway[Any, Any]: ...
+
+
+def gateway(fn: Callable) -> Gateway:
```
CoronRing: The overloads are correct; however, it was actually on purpose to make the gateway a single entry point to reduce the number of overloads, so IDE typehint handles this better. Also, since async is required + call param is always args/kargs (position is supported, but its more like a convenience thing; report that as args is still correct + middleware is dynamically linked (ie user written once, system ones do not change interface) there just seems no need for a case where the overload actually helps that much in user experience vs the pain of seeing 3 overload when typing `gateway(`

I guess it's always the balance if we are writing for a Static Type Checkers (auto check) or IDE (human experience like pylance). Honestly, both will be replaced by AI, which likes both  =D

packages/railtracks/src/railtracks/middleware/set.py
```diff
@@ -122,36 +117,41 @@ def _coerce_gateways(items: Iterable[Any] | None) -> list[Gateway]:
...
     """
 
     def __init__(
         self,
-        wrappers: Iterable[Wrapper] | None = None,
-        gateway_entry: Iterable[Gateway] | None = None,
-        gateway_exit: Iterable[Gateway] | None = None,
-        inner_wrappers: Iterable[Wrapper] | None = None,
+        wrappers: Iterable[Wrapper[_P, _R]] | None = None,
+        gateway_entry: Iterable[Gateway[_P, tuple[tuple, dict[str, Any]]]]
```
CoronRing: gateway entry vs exit is actually expected to have the same interface, (ie a logger can be attached to entry and exit).

In fact user do not have to return anything (no return) for the system to work; the dynamic layer will capture this and pass the input to output as needed. 

so if we want all types to be 100%, it would have to be Callable[_P, _R | None] for both wrapper and gateway, but as with other types, they only cover corner cases and usecase people care about is fullly covered by the base Callable[_P, _R] (ie no one will directly invoke a gateway and depend on it to not return a none), that its not worth the effort to type as it sometimes adds confusion.

CoronRing (commented):
Great job updating the hints! Just some comments explaining the original idea that maybe consulted. >_<

packages/railtracks/src/railtracks/built_nodes/llm_helpers.py
soulFood5632: Just there for completeness to protect against any downstream rep exposure.

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
soulFood5632: i think I would good with both 1 and 2. I don't really like the straight trace as ordering is not really explainable in code which is not great. I think there is argument for either 1 or 2 that we will need to decide. I will bring it up to the team on thrusday.

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
soulFood5632: The node may or may not enforce but this but the idea is the middlwares can't change types or else you end up breaking the modular contract of a node. (i.e. its return or input types do not match what they actually are). In the approach i have above it does at a static typing level which is nice.

packages/railtracks/src/railtracks/middleware/primitives.py
soulFood5632: These overloads only had to be there becuase of the limiattions of paramspec within python. Basically this approach allows us to pass in generalized *args, **kwarsg methods without breaking things. 

in general if there is a way for us to type it correctly, I am always gonna err on that side.

packages/railtracks/src/railtracks/middleware/set.py
soulFood5632: Well, maybe I didn't understand what you meant but gateway exists operate after the node has completed on its return type and the gateway entries operate on the inputs are returned modified inputs. 

so they will have different types 

that is good point with the None. I don't mind adding that. I think we can put up as a future issues as it is a simplicity thing but it doesn't prohibit our abilities to do anything

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
CoronRing: I think allowing a contract to be broken is the whole point of a  middleware, for instance, if we define a node as input message to llm and output model response, allowing insertable variables (ie {var}) already breaks this soft contract in that the input is passed to a model directly. In other words the whole purpose of gateway is to 1. observe, 2. bend the input somehow.

Same for types, if the user wants to have middleware do some lifting in type conversion (ie, passing in a custom type, have that reformatted to rt supported structure) I don't see a reason of not supporting it. 

A simple example might be a guardrail on tool node that force converts a string to int because LLMs often return the wrong type.

packages/railtracks/src/railtracks/middleware/primitives.py
CoronRing: I think a return type can be correctly typed only when the class itself is generic (Generic[P, R]) (meaning gateway need to be overloaded with awaits etc. 

Also instead of `def gateway(fn: Callable[_P, Awaitable[_R]]) -> Gateway[_P, _R]: ...` I think `def gateway(fn: Callable[_P, Awaitable[_R]]) -> Gateway[_P, Awaitable[_R]]: ...` since gateway does not transform.

In addition, def gateway(fn: Callable[..., Any]) -> Gateway[Any, Any]: ... is actually a dead case that is fully covered by the top two so its nore required. 

Without genric

Image

With:

Image

packages/railtracks/src/railtracks/middleware/set.py
CoronRing:
```
Well, maybe I didn't understand what you meant but gateway exists operate after the node has completed on its return type and the gateway entries operate on the inputs are returned modified inputs.
```
For example, if all I want to do is to strip a secret,
``` python
def filter(in:str)->str
   return in.replace("key", "safe")
```
or 
``` python
def log(in:str)
   print(in)
```
The same gateway can be attached to either the entry or exit. 

so a entry gate do not enforce: tuple[tuple, dict[str, Any]]], they can return anything they want, just like an output do not enforce [_R, _R]

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
soulFood5632: I'm going to disagree. Structurally, middleware as a concept should preserve types.
This matters specifically because of how contracts work in Railtracks: when you call a node, it should return the type you expect. If middleware can change that type arbitrarily, you can no longer guarantee the returned type, and silently breaking the caller's contract isn't an acceptable tradeoff.

The key distinction: middleware should let you modify the value of `{val}`, because that's a value transformation, not a type transformation. Value transforms are flexible and cover the large majority of cases. Type transforms belong in node logic, which is where they can be expressed explicitly.

Type transforms are a reasonable use case. They're just not middleware at their core. So the cost of breaking the type contract isn't worth the benefit of supporting that case here.

packages/railtracks/src/railtracks/middleware/primitives.py
soulFood5632: so yea it may not seem obvious what that overload is doing on first glance but it actually does matter.

basically if you have a generalized gateway which takes in args and kwargs without the (weird param overload) it creates a weird situation where the generic when passed into a middleware list with another more specific gateway it can't reconcile the types and says there is an error. You can see an example of that in the code example I provided. 

By loosening the one overload to Any's it allows the the general gateway to conform to the specific gateways for that node without raising a type error.

packages/railtracks/src/railtracks/middleware/set.py
soulFood5632: As much as I agree with you in that specific case at their core they are structurally different interfaces. The flexibility of my generic approach allows for a much greater sense of type checking that is worth the trade off.

docs/advanced_usage/node_builder.md
Pooria90, line 1: Is this going to be published at some point?

soulFood5632: Yea we will traansforming it into something a bit more user facing soon

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
```diff
@@ -55,6 +57,7 @@ def wrapper(_cls, *args, **kwargs):
 _T = TypeVar("_T")
 _P2 = ParamSpec("_P2")
```
Pooria90, line 57: I'm not a typing expert; but was wondering if we really need to define a T2 and P2; can't we just use T and P for `function` too?

CoronRing: I miss the python 3.14 syntax `def name[P, T](var:P, Var:T)` -_-

soulFood5632: I did a sweep but I'll do another. This problem is the initial _T and _P are bound to the original class and then for the class methods we need to create new bindings to the created class becuase it can't use the generics bound to class. I'll ask best friend if I have overcomplicated it.

PR #1220, opened 2026-06-25
1206 streaming v2 middleware

CoronRing:
## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist


CoronRing:
- There is two interface RT support
  - Flow and Node, both need to support:
    - `invoke` + callback: interface compatible as it returns the same thing as static, non-stream invoke. 
    - `stream` + generator: Easier to navigate and write, many existing systems might expect a generator interface.
    - Call back is a paired logic, it will only match to the invoking node: ie .call(abc, callback=cb) will only trigger when abc itself yield, not any child.
    - If user returns a generator for `node` on `call`, system will not handle as async type is not compatible. (AsyncGenerator, not a coroutine, which gets handled)
  - Although ideally only the last invoke (last node) need to be streamed in pratical use, but we need to support streaming at any level because of compatibility. 
  -  We support both for a bit of flexibilty, but if some one whats to reduce complexity, we can go with callback-only and have a simple wrapper that converts callback to generator.
  - we can esily convert a generator to callback, and wise-versa at any level. But the principle of async is somewhat betrayed, blocking the event loop.

``` python
# example converter
async def consume_generator_with_callback(
    async_gen: AsyncGenerator[Any, None],
    callback: Callable[[Any], Any],
):
    """Consumes an async generator and pushes every chunk to a callback."""
    async for chunk in async_gen:
        # Check if the provided callback is async or sync, and execute accordingly
        if asyncio.iscoroutinefunction(callback):
            await callback(chunk)
        else:
            callback(chunk)
```
``` python
def call_as_sync_generator(agent, user_input):
    q = queue.Queue()

    # 1. Define the worker that will get trapped inside the blocking rt.call
    def worker():
        def wrapped_callback(chunk: str):
            q.put(chunk)  # Push chunk to the queue

        # This blocks the background thread until it's completely done
        rt.call(agent, user_input, callback=wrapped_callback)

        q.put(None)  # Push the "all done" signal

    # 2. Start the background thread
    threading.Thread(target=worker, daemon=True).start()

    # 3. The main thread pulls from the queue and yields
    while True:
        chunk = q.get()  # Blocks the main thread until a chunk arrives
        if chunk is None:
            break
        yield chunk
```

docs/tutorials/notebooks/gateway_middleware.md
soulFood5632, line 1: probaably best in a seperate PR

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
```diff
@@ -113,6 +116,42 @@ def agent_node(
...
+    tool_nodes: Iterable[Type[Node] | RTFunction] | None = None,
+    output_schema: None = None,
+    llm: StreamingModelSource,
+    system_message: SystemMessage | str | None = None,
+    manifest: ToolManifest | None = None,
+    middleware: MiddlewareChain | list | None = None,
+    model_middleware: MiddlewareChain | list | None = None,
+    context_injection: bool = True,
+    stream: Literal[True],
+) -> type[Node[[UserInput], StringResponse]]: ...
```
soulFood5632, line 139: this typing will not work since it returns a generator

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
```diff
@@ -113,6 +116,42 @@ def agent_node(
...
+    *,
+    tool_nodes: Iterable[Type[Node] | RTFunction] | None = None,
+    output_schema: None = None,
+    llm: StreamingModelSource,
+    system_message: SystemMessage | str | None = None,
+    manifest: ToolManifest | None = None,
+    middleware: MiddlewareChain | list | None = None,
+    model_middleware: MiddlewareChain | list | None = None,
+    context_injection: bool = True,
+    stream: Literal[True],
```
soulFood5632, line 138: do we need this stream parameter if the type of the llm changes?

soulFood5632: Looking at the below logic. this is just a weird API to me. It seems weird to me to put in stream flag in 2 different places. I think we need to think this through a bit more.

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
```diff
@@ -149,6 +189,11 @@ def agent_node(
             Defaults to True. Set to False to disable context injection for this specific agent regardless
             of the session-level prompt_injection setting. Can also be controlled at the session level via
             rt.Session(prompt_injection=False) or per-message via message.inject_prompt = False.
+        stream (bool): Enable the streaming execution path. When ``True`` the node still returns a
+            ``StringResponse`` / ``StructuredResponse`` (unchanged from the non-streaming API), but
+            each token chunk is broadcast in real time to the ``broadcast_callback`` registered on
```
soulFood5632, line 194: I don't think this design makese sense to use the broadcast callback. The main principle of broadcast callback is its generic so if we wanted to inject a callback it should be a seperate input.

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
```diff
@@ -128,6 +134,7 @@ def llm(
         ]
         | None = None,
         context_injection: bool = True,
+        stream: Literal[False] = ...,
```
soulFood5632, line 137: again see above, but we now have 2 seperate flags again.

packages/railtracks/src/railtracks/built_nodes/llm_helpers.py
```diff
@@ -80,13 +86,15 @@ class ModelInvoker:
...
         self,
-        model: ModelSource,
+        model: ModelSource | StreamingModelSource,
         middleware: MiddlewareChain[
             [MessageHistory, type[BaseModel] | None, list[Tool] | None], Response
         ]
         | None = None,
     ):
-        self._get_model = model if callable(model) else lambda: model
+        self._get_model: Callable[[], ModelBase[Literal[False]] | ModelBase[Literal[True]]] = (
```
soulFood5632, line 95: this typing is weird,  since you have created a union for the 2 possible generic types.

packages/railtracks/src/railtracks/built_nodes/llm_helpers.py
```diff
@@ -162,6 +214,20 @@ def llm_invoke_factory(
 ) -> StructuredLLMInvoke[_TStructured]: ...
 
 
+class StringLLMStreamInvoke(Protocol):
+    async def __call__(
+        self,
+        user_input: MessageHistory | UserMessage | str | list[Message],
+    ) -> StringResponse: ...
```
soulFood5632, line 221: Is this the correct type here? Shouldn't it be generator

packages/railtracks/src/railtracks/built_nodes/llm_helpers.py
```diff
@@ -141,6 +149,50 @@ async def _core_llm_call(
...
+                "Use invoke() for non-streaming models."
+            )
+
+        async def _core_llm_stream(
+            messages: MessageHistory,
+            schema: type[BaseModel] | None,
+            tools: list[Tool] | None,
+        ) -> AsyncGenerator[str | Response, None]:
+            if tools is not None and len(tools) > 0:
+                gen = await model.achat_with_tools(messages, tools)
```
soulFood5632, line 182: we don't use the async API

packages/railtracks/src/railtracks/built_nodes/llm_helpers.py
```diff
@@ -141,6 +149,50 @@ async def _core_llm_call(
...
+            elif schema is not None:
+                gen = await model.astructured(messages, schema=schema)
```
soulFood5632, line 184: ditto

packages/railtracks/src/railtracks/built_nodes/llm_helpers.py
```diff
@@ -141,6 +149,50 @@ async def _core_llm_call(
...
+            else:
+                gen = await model.achat(messages)
```
soulFood5632, line 186: ditto

soulFood5632 (changes requested):
Hey Guan, thanks for the PR. There are 3 structural issues I want to address before we go further with the review.

**Interface:** right now it requires setting stream=True in two places, on the model and on agent_node. In the dcostring you say "pass a model constructed with stream=True and set stream=True here too."  I want us to agree on what the interface should look like before anything else.

**Deprecated LLM methods:** The implementation builds on achat, astructured, and achat_with_tools. These exist in the codebase but they are not mantained by litellm.  We need to base it on the sync version.

**Backend Typing:** ModelInvoker is not designed well enough. It now accepts ModelSource | StreamingModelSource but exposes both invoke() and invoke_stream(), and the # type: ignore[return-value] in the constructor is no bueno. there is corresponding red squiggles and overirddes everywhere. This needs a proper redesign so we use the generics properly.

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
CoronRing: I agree its a weird api, but I cannot think of a solid way around it. LLM def need its own stream, as user may want to invoke it streamly themselves, the stream at the node level is somewhat not required, more so for supporting of type (ie, when a generator is returned), as well as trigger what wrapper version (async_generator safe or couratine safe, as they cannot co-exist). Let me know if you have a better idea

Issue #1144, opened 2026-06-01
[Feature] Implement Obseravbility of Add Ons


soulFood5632:
### Feature Request

Right now add on's are running amuck with now observability tied to them within the node structure. 
```
    async def wrapped_invoke(self, *args: _P.args, **kwargs: _P.kwargs) -> _TOutput:
        """
        A special method that will track and save the latency of the running of this invoke method.
        """
        invoke_method = self.invoke
        for wrapper in self.wrappers:
            invoke_method = wrapper(invoke_method)

        prelim_args, prelim_kwargs = args, kwargs
        for input_map in self.input_maps:
            prelim_args, prelim_kwargs = await input_map(*prelim_args, **prelim_kwargs)

        result: _TOutput = await invoke_method(*prelim_args, **prelim_kwargs)  # type: ignore

        for output_map in self.output_maps:
            result = await output_map(result)

        return result

```

We should implement observability of the components utilizing the new structure found in #1137. Using this node obseravability store we can track the add ons. 

--- 

Tracking for each type of add on will be different. For input and output maps we can show the changes and their latency (of course if any agents are run within this structure they will also be tracked and shown). For the wrappers we can implement custom observability for sample ones like retries but for the scope of this ticket just lay the groundwork for general tracking which we fill later.  

One key challenge will be how we track agents within these add ons if they do run. In our state object they will currently show as tools called by the parent node (however this is not ideal and we should probably be smarter about this). I think one easy solution could be to set a flag within add on uuid context that way the UI side of things can interpret and handle it accordingly, mapping agent runs to add on ids so it can be fixed later. This compoenent will need to carefully worked through.

PR #1226, opened 2026-07-03
1139 feature migrate guardrails

### LINKED ISSUE #1139: [Feature] Migrate Guardrails to new Add ons API
### Feature Request
In #1132 we made some major updates which removed the previously implemented guardrails functionality.

We should reimplement the logic within the new interface, ensuring that there is no lost functionality.

This issue is designed to be the initial derisk of the migration of the previous tooling. Future work will allow us to complete all the other guardrails. These will include any form of adjuster around tool-calling LLMs.

### Additional Context
I would like to nail down the definition of a guardrail to be of 2 main classes: Mappers and logic adjusters. Both fit into the new architecture at the node and LLM level. The simplest should be handled first (think PII detection, keyword finders, etc.) that just throw exceptions on failures. The next set, which can trigger retries on failure, should be handled after. 

Note there are going to be some reasonable questions about whether some should live at the LLM level or at the Node level. If in doubt, don't be afraid to implement them in both places.

### LINKED ISSUE #1229: [Feature] Wire guardrails into the `agent_node`
### Feature Request

This issue covers the following tasks:

1. Add hooks for guardrails in the LLM node building
2. Make necessary changes in middleware interface to adapt to guardrail requirements
3. Define `gate` adapters to translate `Guard` class and `GuardRunner` logic into the `gate`/`middleware` logic.
4. Bring `guardrails=` back to the `agent_node`

### Additional Context

Future issues include adding observability using guardrail traces, removing orphaned module and classes (like mixins) and updating the full test set and documentations.

Pooria90:
## Summary

Wires guardrails into the new middleware architecture so `agent_node(guardrails=Guard(...))` works again after the NodeBuilder unification. Guardrails are expressed as **middleware gates on the model call** rather than the old `Guarded*` subclasses/mixin: input rails become an entry gate that runs as the *last* check before the model (after context injection and any user middleware), and output rails become an exit gate that has the *last word* on the response.

The existing guardrails core (`GuardRunner`, `Guard`, `GuardrailDecision`, built-in rails) is reused verbatim. Covers tasks 1–4 of the issue; observability (traces), orphaned-module removal, and the full test/doc sweep are intentionally left for follow-up PRs to keep this one reviewable.

closes #1229

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

**What changed**
- New `guardrails/llm/guardrail_gates.py`: `guardrail_input_gate` / `guardrail_output_gate` adapt a `Guard` into middleware gates, delegating to `GuardRunner` and translating its `(value, traces, decision)` result into the gate contract (transform → forward, BLOCK → raise `GuardrailBlockedError`, ALLOW → pass through). Output rails only fire on the final reply.
- `MiddlewareChain.register_sys_entry_gate` gained a `position="before"|"after"` arg so the input guard can run as the innermost entry gate (hugging the model call); also added user-layer `add_wrapper`/`add_entry_gate`/`add_exit_gate`/`add_inner_wrapper`.
- `NodeBuilder.llm` / `agent_node` accept `guardrails=` and register the gates on the `ModelInvoker` as fixed, non-reorderable system gates (mirrors context injection).
- Fix: `llm_invoke` re-raises `NodeInvocationError` before its generic `except → LLMError`, so a `GuardrailBlockedError` raised from a gate surfaces as-is instead of being masked as an `LLMError`.

**Not in this PR (follow-ups):** guardrail-trace observability (routed through a single `_record_guard_traces` seam, currently a debug-log placeholder), removal of the orphaned `Guarded*` classes + `LLMGuardrailsMixin`, and the full test/doc update (incl. rewriting the legacy `..._uses_guarded_base` integration tests).

**Example** :

```python
import railtracks as rt
from railtracks.guardrails import Guard, GuardrailDecision, InputGuard, LLMGuardrailEvent


class BlockPasswords(InputGuard):
    """Block any request that mentions a password before it reaches the model."""

    def __call__(self, event: LLMGuardrailEvent) -> GuardrailDecision:
        text = " ".join(m.content for m in event.messages if isinstance(m.content, str))
        if "password" in text.lower():
            return GuardrailDecision.block(reason="passwords are not allowed")
        return GuardrailDecision.allow(reason="ok")


agent = rt.agent_node(
    name="assistant",
    llm=rt.llm.OpenAILLM("gpt-4o-mini"),
    guardrails=Guard(input=[BlockPasswords()]),
)

flow = rt.Flow(name="demo", entry_point=agent)

print(flow.invoke("What's the weather like?"))  # passes → normal reply
try: 
    flow.invoke("What is the admin password?")  # raises GuardrailBlockedError before the model call
except Exception as e:
    print(repr(e))
```

### DIFF
--- a/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
+++ b/packages/railtracks/src/railtracks/built_nodes/_node_builder.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import functools
 from typing import (
+    TYPE_CHECKING,
     Any,
     Callable,
     Coroutine,
@@ -26,6 +27,10 @@ from railtracks.built_nodes.llm_helpers import (
     llm_observe,
     llm_prepare_called_as_tool_factory,
 )
+from railtracks.guardrails.llm.guardrail_gates import (
+    guardrail_input_gate,
+    guardrail_output_gate,
+)
 from railtracks.llm import (
     Parameter,
     SystemMessage,
@@ -42,6 +47,9 @@ from railtracks.validation.node_creation.validation import (
     _check_duplicate_param_names,
     _check_tool_params_and_details,
 )
+
+if TYPE_CHECKING:
+    from railtracks.guardrails.core import Guard
 
 
 def classmethod_preserving_function_meta(func):
@@ -121,11 +129,13 @@ class NodeBuilder(Generic[_P, _T]):
         connected_nodes: Iterable[Type[Node]] | None = None,
         tool_details: str | None = None,
         tool_params: list[Parameter] | None = None,
-        middleware: MiddlewareChain[[UserInput], StringResponse] | None = None,
+        middleware: MiddlewareChain[[UserInput], StringResponse] | list | None = None,
         model_middleware: MiddlewareChain[
             [MessageHistory, type[BaseModel] | None, list[Tool] | None], Response
         ]
+        | list
         | None = None,
+        guardrails: Guard | None = None,
         context_injection: bool = True,
     ) -> NodeBuilder[[UserInput], StringResponse]: ...
 
@@ -143,11 +153,14 @@ class NodeBuilder(Generic[_P, _T]):
         tool_details: str | None = None,
         tool_params: list[Parameter] | None = None,
         middleware: MiddlewareChain[[UserInput], StructuredResponse[_TStructured]]
+        | list
         | None = None,
         model_middleware: MiddlewareChain[
             [MessageHistory, type[BaseModel] | None, list[Tool] | None], Response
         ]
+        | list
         | None = None,
+        guardrails: Guard | None = None,
         context_injection: bool = True,
     ) -> NodeBuilder[[UserInput], StructuredResponse[_TStructured]]: ...
 
@@ -163,11 +176,13 @@ class NodeBuilder(Generic[_P, _T]):
         connected_nodes: Iterable[Type[Node]] | None = None,
         tool_details: str | None = None,
         tool_params: list[Parameter] | None = None,
-        middleware: MiddlewareChain[[UserInput], _R] | None = None,
+        middleware: MiddlewareChain[[UserInput], _R] | list | None = None,
         model_middleware: MiddlewareChain[
             [MessageHistory, type[BaseModel] | None, list[Tool] | None], Response
         ]
+        | list
         | None = None,
+        guardrails: Guard | None = None,
         context_injection: bool = True,
     ) -> NodeBuilder[[UserInput], _R]:
         instance = cls()
@@ -187,6 +202,15 @@ class NodeBuilder(Generic[_P, _T]):
             model_invoker.register_sys_entry_gate(context_injection_gateway)
 
         model_invoker.register_sys_wrapper(llm_observe)
+
+        # Guardrails: fixed, non-reorderable system gates; input last before the core, output the last word.
+        if guardrails is not None:
+            if guardrails.input:
+                model_invoker.register_sys_entry_gate(
+                    guardrail_input_gate(guardrails), position="after"
+                )
+            if guardrails.output:
+                model_invoker.register_sys_exit_gate(guardrail_output_gate(guardrails))
 
         casted_instance._invoke = llm_invoke_factory(
             model_invoker=model_invoker,
--- a/packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
+++ b/packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
@@ -7,6 +7,7 @@ from railtracks.built_nodes.concrete import (
 )
 from railtracks.built_nodes.concrete.response import StringResponse, StructuredResponse
 from railtracks.built_nodes.llm_helpers import ModelSource
+from railtracks.guardrails.core import Guard
 from railtracks.llm.message import SystemMessage
 from railtracks.llm.tools.parameters._base import Parameter
 from railtracks.middleware import MiddlewareChain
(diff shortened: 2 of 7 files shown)


packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
```diff
@@ -97,6 +101,7 @@ def agent_node(
     manifest: ToolManifest | None = None,
     middleware: MiddlewareChain | list | None = None,
     model_middleware: MiddlewareChain | list | None = None,
+    guardrails: Guard | None = None,
```
soulFood5632, line 104: This is more of discussion point than a comment. I think there are 2 main interface options (and I want to talk through them both) to make sure we aren't making the wrong decision:

1. The way you have it. 
- clear separation of general middleware from guardrails
2. Passed in into middleware for each guardrail item
- no duplication of a similar interface. 
```
middleware: MiddlewareChain(entry_gate[PIIDetect(), ...], exit_gate=[PIIDetect()]
```

My first instinct would be to go with option 2 because largely guardrails are a subset of middleware. However, I can see the ease of use argument as well so I am unsure. What would be the pros and cons to go with your over the the alternative I presented

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
```diff
@@ -121,11 +129,13 @@ def llm(
         connected_nodes: Iterable[Type[Node]] | None = None,
         tool_details: str | None = None,
         tool_params: list[Parameter] | None = None,
-        middleware: MiddlewareChain[[UserInput], StringResponse] | None = None,
+        middleware: MiddlewareChain[[UserInput], StringResponse] | list | None = None,
         model_middleware: MiddlewareChain[
             [MessageHistory, type[BaseModel] | None, list[Tool] | None], Response
         ]
+        | list
```
soulFood5632, line 136: flagging this becuase I missed this abvoe and need to make it so you can't input a list. I will handle it in my own work.

packages/railtracks/src/railtracks/guardrails/llm/guardrail_gates.py
```diff
@@ -0,0 +1,149 @@
...
+logger = get_rt_logger("guardrails")
+
+
+def _node_metadata() -> tuple[str | None, str | None]:
+    """``(node_uuid, run_id)`` for event observability.
+
+    Returns ``(None, None)`` when no run context is active — the gate logic never
+    depends on this metadata, so it must not raise just to populate it.
+    """
+    if not is_context_present():
```
soulFood5632, line 45: @Amir-R25 look another one of us has come to same conclusion of observability using the contextvars.

packages/railtracks/src/railtracks/guardrails/llm/guardrail_gates.py
```diff
@@ -0,0 +1,149 @@
...
+        return None, None
+    return get_parent_id(), get_run_id()
+
+
+def _record_guard_traces(traces: list[GuardrailTrace]) -> None:
+    """The single sink for guardrail traces.
+
+    TODO: Determine the correct home for observability traces. For now, just log them at debug level.
+    """
+    for t in traces:
```
soulFood5632, line 55: we can handle this in #1144

packages/railtracks/src/railtracks/middleware/set.py
```diff
@@ -206,15 +220,49 @@ def _fresh_copy(self) -> "MiddlewareChain":
...
+        self, g: Gate[_P, tuple[tuple, dict[str, Any]]] | Callable[..., Any]
+    ) -> None:
+        """Append a user entry gate. Transforms the input args before the core runs."""
+        self._entry.add_user(_coerce_gate(g))
+
+    def add_exit_gate(self, g: Gate[[_R], _R] | Callable[..., Any]) -> None:
+        """Append a user exit gate. Transforms the return value after the core runs."""
+        self._exit.add_user(_coerce_gate(g))
+
+    def add_inner_wrapper(self, w: Wrapper[_P, _R] | Callable[..., Any]) -> None:
```
soulFood5632, line 242: I added indentical functions in my work. I will need to be careful to merge.

soulFood5632 (changes requested):
Most of my comments are either to myself or general discussion points. The first comment about how you pass in Guardrails is the only critical point to talk about.

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
Pooria90: The obvious pro of option 2 is unification: we can use the same interface for middleware and guardrail.

The key point is that `guardrails=Guard(...)` is not a parallel interface. It compiles down to gates under the hood. We still have option 2's flexibility and anyone can customize their guardrails as middlewares. With option 1, we have both a safe default and a freeform path.

The main reason for option 1 was that guardrails have extra requirements, for example they are fixed-place policy layers; they are non-bypassable: every input guardrail must run right before the core model call and every output guardrail should run right before the final output, so context injection or other middlewares won't be able to alter their outputs.  They also have behaviors other than data transforms (e.g. a BLOCK).

Another reason for me to go with 1 was that current guardrails are not gates. Going with 2 meant writing the entire built-ins in the guardrail module in the middleware literature and breaking the backward compatibility.

soulFood5632: Yep I think this is reason enough to stick with your interface. I think we could run into some trickiness around where we draw the line but we can deal with that when the time comes.

soulFood5632 (approved):
LGTM

Issue #1237, opened 2026-07-10
Define a Clean Interface for the Prebuilt Add ons


soulFood5632:
### Description

We need to have a very clean interface for the exposing prebuilt add ons. 

So in the recent changes to #1221, we have changed the interface for middleware to be very clean. But we have not decided how we are gonna be exposing all of our prebuilt middleware and guardrail objects. 

We need to decide and define a clear interface. 

 

rt.agent_node(
    "Agent",
    tool_nodes=[],
    llm=rt.llm.OpenAILLM(model_name="gpt-4o"),
    middleware=[retry_middleware, middleware.MaxCreations(2)],
    model_middleware=[middleware.Retry(max_retries=10), guards.MaxCost(100), guards.KeywordDetection("password"), guards.PIIDetection()],
)

As part of this issue the following should be completed:
- Define an interface for all of our Add Ons (breaking it up separate sub modules as you seem fit)
- Write overview docs on the approach you described above
- Write a super small internal guide on where you should be creating your prebuilt add ons and put in the parent ticket.

soulFood5632:
By interface I mean, it should be how a user would import the middleware:

so lets say a user wants to use our middleware components we have built. Literally how do they import them and use them. 

Langchain may have some good examples:

https://reference.langchain.com/python/langchain/middleware

soulFood5632:
Great work here @Pooria90 I'm gonna close it

Issue #1145, opened 2026-06-01
[Feature] General Issue for Supported Add Ons


soulFood5632:
### Feature Request
We will need to write out a large number of default add-ons to be used by your agents. An incomplete list follows (each will need to be written up as an issue soon):

| Category | Add-ons |
| -- | -- |
| Pre-Message History Mapping | • PII detection • Keyword finders • General mappers • Prompt injection checks • Stateless compression (summarization, etc.) • Context injection |
| Pre-Invoke Mapping | • General input guardrails (SQL injection prevention) • General observability |
| Post-Message History Mapping | • PII detection • Keyword finders • General mappers • Context deinjection |
| Post-Invoke Mapping | • General input guardrails (PII detection, etc.) • Tool summarizer • General observability |
| Pre-Message History Mapping (Stateful) | • Stateful mirrors of all the above |
| Pre-LLM Calls | • Observability for LLMs • Keyword filters • Context injection |
| Post-LLM Calls | • Observability for LLMs • Keyword filters • Context injection |
| LLM Logic Adjustments | • Retries • Model selection • Model fallback • Token limits • General rate limiting |
| LLM Tools Logic | • Tool emulator • Tool filtering • Retries • Human-in-the-loop • Max tool calls • Tool calling strategy • Tool authorization |
| Tools | • Shell script tools • File system tools • Todo and planning • Memory modules |
| General Wrappers | • Timeouts |
| Human-in-the-loop | • Used as a wrapper |

### Additional Context
No response

soulFood5632:
As we get closer to tackling this I will add issues for specific add ons that we can expose to the community.

Issue #1140, opened 2026-06-01
Parent Issue for Add Ons


soulFood5632:
There are various work items tied to the new Add On structures. They will be documented under this ticket

Amir-R25:
@soulFood5632 good to close? All sub issues seem to have been closed

edit: nvm, I see we have some open ones left

soulFood5632:
Nope they are good to go. I moved them out from the epic as they are no smaller additions to be worked on later.

Issue #1137, opened 2026-06-01
[Feature] Migrate Previous LLM Observability into new Hooked based Approach


soulFood5632:
We need to create events per design in #1264. There is some confusion here but design decisions to be made

> for more details please contact @soulFood5632 

# Below is a previous spec 

### Feature Request
In #1132 there were some major changes to the structure of LLMs. This included the deletion of previous LLM observability tracking pieces. We need to migrate them into the hook-based approach utilizing the ModelGateway type.

In general this should happen without the user even knowing about it. It should be internally added to all LLM nodes at creation time.

----

Previously we just added the history to the details category of the node itself. This was largely a band-aid solution that led to many issues. In this new implementation, we should leverage a storage layer that lives at the session level and tracks all details attached to each node ID in the session. This change is akin to taking our heavily object-oriented (encapsulated) approach and making it more relational (NoSQL → SQL).

I propose a new object called a NodeObservability store, living at the session level, that the observability hooks use to store and save data for each model call. (Make sure to leverage other key context variables like `get_parent_id` in order to access the correct node.)

In the serialization method, we will need to reconstruct the previous data structure using the NodeObservability store.

# Overview of Requirements:
- No lost functionality from the previous approach on main
- Utilize a global "session-scoped" observability instead of the current encapsulated storage
- Make sure all existing tests still pass, and add regression tests for the new behavior

### Additional Context
For reading on the changes please review:
[i-would-like-you-inherited-forest.md](https://github.com/user-attachments/files/28482350/i-would-like-you-inherited-forest.md)

soulFood5632:
Most of this stuff around node observability store is now irelevant and already completed via other work in the parent: #1239

PR #1238, opened 2026-07-10
1221 middleware public api change

### LINKED ISSUE #1146: [Feature] Design a Clean user facing interface for user defined add ons
### Feature Request

Langchain provides some nice helpful wrappers to make your own middleware 

Image

Of course we call ours add ons but we should also provide a nice clean interface for people to use if they want to implement their own. This should be document and designed with user facing care. 

### Additional Context

_No response_

### LINKED ISSUE #1221: [Bug] Remove ability for decorator passage of middleware
### Description

Structurally middleware should never be passed in as part of the decorator. This defeats the purpose of the customizability. This should be removed and instead you should be able to add middleware dynamically to any node.

N/A

### Additional Context

_No response_

### Operating System

_No response_

### railtracks Version

_No response_

### Python Version

_No response_

soulFood5632:
### Public API
I'll start by demonstrating the big points of the new API

We have a set of new wrappers that will allow the public to easily create their own middleware (in addition to us internally creating them). Check out the below code to see it in action.

The main idea is we now have the following:

- `@rt.before_model`  -> runs before model calls
- `@rt.after_model` -> runs after model calls
- `@rt.wrap_model` -> wraps model calls
- `@rt.wrap_node` -> wraps a node
- `@rt.after_node` -> runs after a node

Critically you can insert the `node` level middleware into either `function_nodes` or `agent_nodes`. Additionally node level middleware can be added after the type has been created using `rt.couple(node, *middleware)`. This will create a new type so it doesn't pollute the previous lists.

`agent_nodes` in particular have an additional parameter called model_middleware which will allow you pass in Middleware of the specific type of the model calls. As of right now we are only supporting this addition of these components at type creation time. This could be future work as outlined in the notes section.

```python
import railtracks as rt


@rt.before_model
def print_message(message_history: rt.llm.MessageHistory, schema, tools):
    print(message_history)
    return message_history, schema, tools


@rt.after_model
def print_response(response: rt.llm.Response):
    print(response.message)
    return response

@rt.wrap_model
async def periodic_failure(llm_call, message_history, schema, tools):
    import random
    if random.random() < 0.5:
        raise Exception("Random failure")
    return await llm_call(message_history, schema, tools)

@rt.wrap_node
async def retry_middleware(call, *args, **kwargs):
    import asyncio
    for i in range(3):
        try:
        except Exception as e:
            print(f"Attempt {i+1} failed: {e}")
            await asyncio.sleep(1)
    raise Exception("All attempts failed")

@rt.after_node
    return result

agent = rt.agent_node(
    "Agent",
    tool_nodes=[],
    middleware=[retry_middleware, print_after],
    model_middleware=[periodic_failure],
)

adjusted = rt.couple(agent, print_after)
```

### Backend Structure
Previously we had Gate's and Wrapper's sitting in our backend. We also had this notion of a Layered list that would allow us to insert things into system list keeping them separate from the user ones. I have made a structural design change to this approach.

Now we are only going to be storing Wrappers on the backend. And further the MiddlewareChain object will be made purely an internal construct. What this means is it allows us to encapsulate the system and user split inside the node itself which allows us to handle whatever we need within our codebase without relying on modifying a user facing object.

```python
_exterior_middleware: list[Middleware[_P, _TOutput]] = []
    _user_middleware: list[Middleware[_P, _TOutput]] = []
    _interior_middleware: list[Middleware[_P, _TOutput]] = []

    def __init__(
        self,
    ):
        # each fresh node will have a generated uuid that identifies it.
        self.uuid = str(uuid.uuid4())
        self.middleware = MiddlewareChain[_P, _TOutput](
            middleware=[
                *self._exterior_middleware,
                *self._user_middleware,
                *self._interior_middleware,
            ]
        )
```

In the above snippet you can see at type creation time we can fill the `_exterior` and `_interior` blocks with whatever internal pieces we need. You can see we create the middleware chain object at object creation time which handles the execution of the items. (quick note here: I am finalizing the ordering approach of the `MiddlewareChain` object so bear with me if in the actual code the ordering is flipped).

In a related change all LLM nodes (or `agent_nodes`). will still have the `ModelInvoker` object. This `ModelInvoker` object will take a list of Wrapper objects which it will create a `MiddlewareChain` from the inputted values and automatically adding the observability logic.

```python
# ModelInvoker class
def create_with_observability(
        cls,
        model: ModelSource,
        middleware: list[ModelMiddleware] | None = None,
    ):
        self._get_model = model if callable(model) else lambda: model
        self._middleware = middleware + _llm_observe
# psuedo code ^
```

One key question about the above class is the choice to encapsulate the observability within the `ModelInvoker` class. Similar to the pattern outlined in the notes section about `context_injection` we may want to allow users to input this middleware if they want observability. Since observability is so core, I think it has more of an argument to be kept in here. We will have to address a similar question in #1144 when we think about general middleware observability.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] Breaking change
- [ ] Docs

## Checklist

## Notes

Open questions that remain:
- We do not support `rt.before` which is done intentionally due to the interface struggles around what it means to return a param spec. There are potential remedies but that is left for future work.
- Do we want to add model middleware after the type has been created using an interface similar to `rt.couple`
- Question for @Pooria90: Now that we have simplified the interface for attachment of middleware, does it make a bit more sense to allow guardrails to be passed in directly so the user can be in control of the ordering. I'm curious to hear your thoughts because I do really like the cleanness of the below API:

```python
rt.agent_node(
    "Agent",
    tool_nodes=[],
    middleware=[retry_middleware, middleware.MaxCreations(2)],
    model_middleware=[middleware.Retry(max_retries=10), guards.MaxCost(100), guards.KeywordDetection("password"), guards.PIIDetection()],
)
```
- Question for @CoronRing and everyone, similar to above: should we move `context_injection` from a parameter to something you inject as a middleware. I would lean yes, but as above I want to hear the thoughts of everyone:

```python
# Option 1

rt.agent_node(
    "Agent",
    tool_nodes=[],
    context_injection=True,
)


# Option 2
rt.agent_node(
    "Agent",
    tool_nodes=[],
    model_middleware=[middleware.ContextInjection()],
)
```

- We really need to decide a clean interface for our additional add ons. We're gonna start building all of the add-ons for #1145. This will be addressed in #1237.

- How do we inject observability into the middleware objects (see #1144) with these new changes?

Closes #1146.


soulFood5632:
Not ready for review just yet. Doing some polishing work and some final testing before it is ready for consumption

soulFood5632:
Key foundational work for #1140

soulFood5632:
Note for reviewers:

I will be fixing bad docstings and writing up proper testing later. Please do not bother reviewing those sections.

Aryan-Railtown:
Quick question: How would one know what kind of parameters are going to be passed into the functions that are going to be converted into `middleware`/`model_middleware`? Is it set or flexible?

1. Overall the design looks clean and intuitive. 
2. As for a similar interface as `rt.couple`, I believe it'll be best if we keep it consistent between `model_middleware` as well as `middleware`, if we are giving users the option of adding middleware post agent instantiation, i think it might be good to have a similar thing for `model_middleware`. Then again we need to consider if it'll be extra work for us to maintain this going forward (we had something similar show up prior to the existence of NodeBuilder and it was a little hard to maintain since it was decoupled) 
3. Quick comment: I especially like how intuitive it is to use guardrails via middleware. I believe this will be a core usage for this feature going forward.

soulFood5632:
> Quick question: How would one know what kind of parameters are going to be passed into the functions that are going to be converted into `middleware`/`model_middleware`? Is it set or flexible?

The typing system will basically check that all for you. The idea is it types must either match the type of the node it is acting on. the typing is flexible however that it will assume your `any`'s will still not raise any warnings bells. The idea is when you use the @wrap_model wrapper is it can provide some additional typing help (even if t the pylance hint is not really doing its job)

> 1. Overall the design looks clean and intuitive.

:)

> 2. As for a similar interface as `rt.couple`, I believe it'll be best if we keep it consistent between `model_middleware` as well as `middleware`, if we are giving users the option of adding middleware post agent instantiation, i think it might be good to have a similar thing for `model_middleware`. Then again we need to consider if it'll be extra work for us to maintain this going forward (we had something similar show up prior to the existence of NodeBuilder and it was a little hard to maintain since it was decoupled)

Yea that is a good point. I think one easy way to go about things is to just add it to the method and raise an exception if they provided `model_middleware` for a function node. 



> 3. Quick comment: I especially like how intuitive it is to use guardrails via middleware. I believe this will be a core usage for this feature going forward.

Yea I want to make sure the design works with @Pooria90 but the stackability of things seems nice. It really gives the user full control of things.

packages/railtracks/src/railtracks/middleware/chain.py
```diff
@@ -0,0 +1,49 @@
...
+    def add_middleware(self, m: Middleware[_P, _R]) -> None:
+        """Append a user outer middleware (outermost band). Runs around the whole call."""
+        self._middleware.append(m)
+
+    @property
+    def middleware(self) -> list[Middleware[_P, _R]]:
+        """User-layer outer middleware (excludes system-registered layers)."""
+        return list(self._middleware)
+
+    async def run(
```
Pooria90, line 36: I like how simple this has turned. Good job!

packages/railtracks/src/railtracks/middleware/chain.py
```diff
@@ -0,0 +1,49 @@
...
+from railtracks.middlewares.core import Middleware
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+
+class MiddlewareChain(Generic[_P, _R]):
+    def __init__(
+        self,
+        middleware: Iterable[Middleware[_P, _R]] | None = None,
```
Pooria90, line 21: possibly change to `middlewares` as it's an Iterable?

Amir-R25:
This looks great, putting a few thoughts, some will feel like nits but I do think naming things will matter a bit here especially for adopotion:

- Regarding Context Injection: I vote for it to be moved inside the middelware param for consistency. 
- nit, I suggest renaming `rt.before_/after_/wrap_llm` to be more specific
- nit: `rt.after/(before)` might be worth having a node appended to it. 
- Does `rt.couple(node, *middleware)` modify the inputted node or does it return a new node type?

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
```diff
@@ -20,7 +20,7 @@
 from railtracks.built_nodes._node_builder import NodeBuilder
 from railtracks.built_nodes.concrete.function_base import RTFunction
 from railtracks.exceptions import NodeCreationError
-from railtracks.middleware import MiddlewareChain
+from railtracks.middlewares.core import Middleware
```
Amir-R25: nit: shouldn't the module be called middleware?

packages/railtracks/src/railtracks/built_nodes/llm/middleware/__init__.py
```diff
@@ -0,0 +1,9 @@
+__all__ = [
+    "before_model",
```
Amir-R25: see my nitpicky comment about the names

soulFood5632:
> * Regarding Context Injection: I vote for it to be moved inside the middelware param for consistency.

I am inclined to agree here
> * nit, I suggest renaming `rt.before_/after_/wrap_llm` to be more specific

Yea I agree there. 
> * nit: `rt.after/(before)` might be worth having a node appended to it.

Yea just to match probably a good idea
> * Does `rt.couple(node, *middleware)` modify the inputted node or does it return a new node type?

It creates a new type. no modifications in place

packages/railtracks/src/railtracks/middleware/chain.py
soulFood5632: I've been doing extensive naming research and the plural of middleware is middleware.

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
```diff
@@ -191,26 +185,23 @@ def llm(
...
 
-        model_invoker = ModelInvoker(model, middleware=model_middleware)
+        unwrapped_model_middleware: list[ModelMiddleware] = (
+            list(deepcopy(model_middleware)) if model_middleware is not None else []
+        )
+        unwrapped_middleware = (
+            list(deepcopy(middleware)) if middleware is not None else []
+        )
+
+        if guardrails is not None and guardrails.output:
```
Pooria90, line 198: My main concern for guardrail was fix location. I think those should come exactly before the model call and right before the final output. 
We can give the full control of the order to the user; the fact that they can control the order means that can put something like the context injection after the guardrails.
But I don't have a strong opinion on it.  We can remove the guardrail parameter for the sake of simplicity.

soulFood5632: I'm gonna leave it in here for this PR and make the adjustments after to keep us up to date.

soulFood5632:
> > 2. As for a similar interface as `rt.couple`, I believe it'll be best if we keep it consistent between `model_middleware` as well as `middleware`, if we are giving users the option of adding middleware post agent instantiation, i think it might be good to have a similar thing for `model_middleware`. Then again we need to consider if it'll be extra work for us to maintain this going forward (we had something similar show up prior to the existence of NodeBuilder and it was a little hard to maintain since it was decoupled)
> 
> Yea that is a good point. I think one easy way to go about things is to just add it to the method and raise an exception if they provided `model_middleware` for a function node.
> 

Following up on this comment. I have created a new issue in #1245 to handle it. Your point about it making this more complex is very real. We have moved to even more stateless approach for the nodes so it makes it even more difficult to inject things like `model_middleware` without extreme care.

packages/railtracks/src/railtracks/middleware/chain.py
soulFood5632: Thanks :)

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
soulFood5632: In general I am so much more of a fan of letting the user shoot themselves in the foot rather than imposing something behind the scenes. That being said if we are super definite about what is a guardrail (vs. some general mapping middleware maybe it would be worth it)

CoronRing:
Not sure if result of a change, it appears 

The proposed interface is updated.
```python
adjusted = rt.couple(agent, print_after)
```

But the implementation takes an iterable as the second positional arg:

```python
def couple(node, middleware: Iterable[Middleware[_P, _R]]) -> ...
```

So docstring etc do not match, I like varargs ver, as the description says, it reads better and matches the `extend_middleware(*middleware)` classmethod it delegates to

packages/railtracks/src/railtracks/interaction/couple.py
```diff
@@ -0,0 +1,40 @@
...
+        unwrapped_node = node.node_type
+        is_rt_function = True
+    else:
+        unwrapped_node = node
+        is_rt_function = False
+    
+    new_klass = unwrapped_node.extend_middleware(*middleware)
+
+    if is_rt_function:
+        node.node_type = new_klass
```
CoronRing: This seems to be modifying the input node, I think it's fine if we document it. But the PR message says we want to preserve the original identity. 

I know it does not directly apply to node, but it seems similar logic. Since `function_node` results are typically shared across agents as tools, this could be a landmine.

We can return a fresh `RTFunction` or document really well.

packages/railtracks/src/railtracks/built_nodes/llm/middleware/before_llm.py
```diff
@@ -0,0 +1,32 @@
...
+    ],
+):
+    @wrap_model
+    async def wrapper(
+        llm_call: LLM_CALL,
+        message_history: MessageHistory,
+        schema: type[BaseModel] | None,
+        tools: list[Tool] | None,
+    ):
+        message_history, schema, tools = await unpack_async_sync(
```
CoronRing, line 39: I tried this locally, it appears Today we must accept all three params and faithfully return all three. I wonder if it can be updated to a similar logic as before for gates, where if nothing is returned, we simply return the original set for this user.

1. This allows for a pure observation task to easily build.
2. This allows modification of the original list (object mutation) to easily pass so user do not even need to return anything, simplifying the interface. This is user code so I think it's a safe build.

CoronRing: I also most wonder if `ModelRequest` dataclass (`.messages`, `.schema`, `.tools`) or alike (similar to langchain) is more convenient, in case we add item (ie, stream tag) so this list does not get too long and the interface do not get breaking changes. 

ie right now we require 3, if user passes 4 in future, this will not work, 4th one does not even get ignored, hence a breaking change. 

This was my fault in the previous design with the args stuff, but args themselves are containers as well so the problem can be avoided.

CoronRing:
Maybe I am dreaming but where did before_node go? Was it not a major piece? I remember someone else mentioned symmetry in the PR, not sure if the changes were already made.

packages/railtracks/src/railtracks/built_nodes/_node_builder.py
```diff
@@ -17,19 +17,18 @@
     overload,
 )
 
+from litellm import deepcopy
```
CoronRing: I don't think we need litellm's deepcopy? I want to avoid litellm import because init is slow as hell (5+ second) we will migrate to remote litellm stuff other than model provider page

packages/railtracks/src/railtracks/built_nodes/middlewares/after_llm.py
```diff
@@ -0,0 +1,24 @@
+from typing import Awaitable, Callable
+
+from litellm import BaseModel
```
CoronRing: Same here, litellm import I believe pydantic import BaseModel is more than enough? I do not see feature depending on specialized version of it.

packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+
+    def __init__(
+        self,
+        model: ModelSource,
+        middleware: list[ModelMiddleware] | None = None,
+    ):
+        self._get_model = model if callable(model) else lambda: model
+        unwrapped_middleware = deepcopy(middleware) if middleware is not None else []
+        self._middleware = MiddlewareChain(
```
CoronRing: `__init__` always prepends `_llm_observe` to the chain. `extend_middleware` deep-copies the chain (which already contains `_llm_observe`) and feeds it back into the constructor. which prepends `_llm_observe` again:

```python
return ModelInvoker(self._get_model, new_middleware_chain._middleware)
# chain is now [_llm_observe, _llm_observe, ...original..., ...new...]
```

Maybe we scan if its there and only add otherwise.

CoronRing:
I do not like the fact there is two major block that are commonly used both name middlewares; it can lead to some confusion. Also `before_llm.py` defines `before_model`, `after_llm.py` defines `after_model`, maybe a file update will help =D.

packages/railtracks/src/railtracks/interaction/couple.py
soulFood5632: Yea good catch. Fixed

packages/railtracks/src/railtracks/built_nodes/llm/middleware/before_llm.py
soulFood5632: Yea that is a fair comment about the data class. I will create an issue and handle that as my next item.

packages/railtracks/src/railtracks/middlewares/core.py
```diff
@@ -0,0 +1,74 @@
...
+        """Compose this middleware onto ``inner``, returning a new callable with the same signature."""
+        fn = self._fn
+
+        async def wrapped(*args: _P.args, **kwargs: _P.kwargs) -> _R:
+            return await fn(inner, *args, **kwargs)
+
+        return wrapped
+
+    def __repr__(self) -> str:
+        return f"Middleware({getattr(self._fn, '__name__', self._fn)!r})"
```
CoronRing: I noticed this when running my code, 
`repr(mw)` is `Middleware('wrapper')` for *every* middleware, a list of five middlewares prints as five identical `wrapper`s, so we can't tell which is which in logger.

Maybe `functools.wraps(fn)(wrapper)` in each decorator, and having `Middleware.__repr__` prefer `self._fn.__qualname__` will help.

packages/railtracks/src/railtracks/built_nodes/llm/middleware/before_llm.py
soulFood5632: > I tried this locally, it appears Today we must accept all three params and faithfully return all three. I wonder if it can be updated to a similar logic as before for gates, where if nothing is returned, we simply return the original set for this user.
> 
> 1. This allows for a pure observation task to easily build.

While I can see your point what I have implemented is what we are gonna deliver for now. We can add that functionality later. I will create an issue for it. 

> 2. This allows modification of the original list (object mutation) to easily pass so user do not even need to return anything, simplifying the interface. This is user code so I think it's a safe build.

Honestly, this is case I'd prefer to avoid modifying lists in place although possible is just such a bad idea in practice and so hard to maintain, so I wouldn't really want to encourage people do it.

soulFood5632:
> I do not like the fact there is two major block that are commonly used both name middlewares; it can lead to some confusion. Also `before_llm.py` defines `before_model`, `after_llm.py` defines `after_model`, maybe a file update will help =D.

Hey Guan, I am currently addresses comments so its a work in progress. I will be updating things once I am done.

soulFood5632:
> Maybe I am dreaming but where did before_node go? Was it not a major piece? I remember someone else mentioned symmetry in the PR, not sure if the changes were already made.

I deleted it a while ago, before I made the PR due to the interface challenges. We can worry about it later.

CoronRing:
i am wondering how `couple()` is future-planned; will it support non-middleware stuff in the future? Currently, the term is somewhat generic and accepts a really specialized object` (middleware)`. Is there a reason `with_middleware` was not a good fit?

soulFood5632:
@CoronRing I appreciate the review but please just hold off. I am still working through updates and the code is not ready for review. The main goal of preparing this PR was to get specific feedback on the design and DX.

soulFood5632:
> i am wondering how `couple()` is future-planned; will it support non-middleware stuff in the future? Currently, the term is somewhat generic and accepts a really specialized object` (middleware)`. Is there a reason `with_middleware` was not a good fit?

There is not much thought on it. To me I think I prefer `couple` over `with_middleware` so I will stick with that for now.

CoronRing:
> @CoronRing I appreciate the review but please just hold off. I am still working through updates and the code is not ready for review. The main goal of preparing this PR was to get specific feedback on the design and DX.

Sorry, github have this thing where I write comment in a batch and send, no more comments

packages/railtracks/src/railtracks/built_nodes/concrete/function_base.py
```diff
@@ -21,3 +21,9 @@ class RTFunction(Protocol, Generic[_P, _TOutput]):
     """
 
     node_type: type[Node[_P, _TOutput]]
+
+    def with_node_type(
```
Amir-R25, line 25: nice 👍

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
Amir-R25: we talked about this, all good if there's no way to avoid

packages/railtracks/src/railtracks/middleware/core.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(
+        self,
+        fn: Callable[Concatenate[Callable[_P, Awaitable[_R]], _P], Awaitable[_R]],
+    ) -> None:
+        _require_async(fn, "Middleware function")
+        self._fn = fn
+
+    def wrap(self, inner: Callable[_P, Awaitable[_R]]) -> Callable[_P, Awaitable[_R]]:
+        """Compose this middleware onto ``inner``, returning a new callable with the same signature."""
+        fn = self._fn
```
Amir-R25, line 57: a little bit strange here we go `self._fn = fn` and then `fn=self._fn` all while these are all just variable reassignments so they also are not offering any protection

packages/railtracks/src/railtracks/middleware/chain.py
```diff
@@ -0,0 +1,49 @@
...
+
+    async def run(
+        self,
+        core: Callable[_P, Awaitable[_R]],
+        *args: _P.args,
+        **kwargs: _P.kwargs,
+    ) -> _R:
+        func = core
+        for m in reversed(self._middleware):
+            func = m.wrap(func)
```
Amir-R25, line 44: the onion?

packages/railtracks/src/railtracks/nodes/nodes.py
```diff
@@ -124,10 +130,20 @@ async def invoke(self, *args: _P.args, **kwargs: _P.kwargs) -> _TOutput:
 
     async def wrapped_invoke(self, *args: _P.args, **kwargs: _P.kwargs) -> _TOutput:
         """
-        Runs ``invoke`` through the node-level middleware (wrappers + gateways).
+        Runs ``invoke`` through the node-level middleware.
         """
         return await self.middleware.run(self.invoke, *args, **kwargs)
 
+    @classmethod
```
Amir-R25, line 137: is the use case for this where a user already has a list of middleware and on a node and then they want to modify the node somewhere else in the code? So they're injecting middleware?

packages/railtracks/src/railtracks/utils/unpack.py
```diff
@@ -0,0 +1,10 @@
+from typing import Awaitable, TypeVar
+
+_T = TypeVar("_T")
+
+
+async def unpack_async_sync(value: Awaitable[_T] | _T, /) -> _T:
```
Amir-R25, line 6: :D

packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
```diff
@@ -0,0 +1,116 @@
...
+    schema: type[BaseModel] | None,
+    tools: list[Tool] | None,
+) -> Response:
+    prev_message_history = deepcopy(message_history)
+    response: Response = await call(message_history, schema, tools)
+    _ = RequestDetails(
+        message_input=prev_message_history,
+        output=response.message,
+        model_name=response.message_info.model_name,
+        model_provider=None,  # TODO: implement parsing logic here
```
Amir-R25, line 34: Are we missing something with the `model_provider` here?

packages/railtracks/src/railtracks/built_nodes/llm/middleware/after_llm.py
```diff
@@ -0,0 +1,28 @@
+from typing import Awaitable, Callable
```
Amir-R25, line 1: I'm noticing the logic and maybe this is getting obsessive so feel free to ignore is that the code for `after_model` and `before_model` so I'm wondering would it be easier (from a maintainability perspective) to unify them and call them `llm_middleware` that accepts a literal parameter called `mode` with either `after` or `before`.

Just a thought, again feel free to resolve this comment

packages/railtracks/src/railtracks/interaction/couple.py
```diff
@@ -0,0 +1,56 @@
...
+def couple(
+    node: CallableAsyncRTFunction[_P, _R], *middleware: Middleware[_P, _R]
+) -> CallableAsyncRTFunction[_P, _R]: ...
+
+
+def couple(
+    node: type[Node[_P, _R]] | RTFunction[_P, _R], *middleware: Middleware[_P, _R]
+) -> type[Node[_P, _R]] | RTFunction[_P, _R]:
+    """
+    Attaches middleware to a Node or RTFunction. Returns a new object; the one passed in is never modified.
```
Amir-R25: nit: technically it "Returns a new type` not a `new object` right?

packages/railtracks/src/railtracks/interaction/couple.py
```diff
@@ -0,0 +1,56 @@
...
+    Ordering:
+    - Middleware = [A, B, C] -> A wraps B, B wraps C, C wraps the node. A -> B -> C -> Node -> C -> B -> A
+    - Existing middleware on `node` is preserved and wraps around the newly added middleware
+      (the new middleware ends up innermost, closest to the node).
+    """
+    if isinstance(node, RTFunction):
+        new_node_type = node.node_type.extend_middleware(*middleware)
+        return node.with_node_type(new_node_type)
+
+    new_klass = node.extend_middleware(*middleware)
```
Amir-R25, line 54: klass? :D

packages/railtracks/src/railtracks/built_nodes/llm/middleware/after_llm.py
```diff
@@ -0,0 +1,28 @@
...
+
+from railtracks.built_nodes._types import LLM_CALL
+from railtracks.llm.history import MessageHistory
+from railtracks.llm.response import Response
+from railtracks.llm.tools.tool import Tool
+from railtracks.middlewares.core import wrap_node
+from railtracks.utils.unpack import unpack_async_sync
+
+
+def after_model(fn: Callable[[Response], Response | Awaitable[Response]]):
```
Amir-R25: putting a note here regarding naming -> `after_llm`

Amir-R25 (changes requested):
Thanks, complicated but powerful work that simplifies the framework quite a bit

packages/railtracks/src/railtracks/built_nodes/llm/middleware/__init__.py
soulFood5632: Bah, my brain got very confused...

packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/function.py
soulFood5632: Ok so I think we should label it middleware. I'll make that change.

packages/railtracks/src/railtracks/built_nodes/llm/middleware/__init__.py
soulFood5632: Done

packages/railtracks/src/railtracks/middleware/core.py
soulFood5632: A lot of this was from when it was extenral module. I think I'll leave it in there even though it doesn;t do much

packages/railtracks/src/railtracks/nodes/nodes.py
soulFood5632: Hmm, didn't actually catch this. The idea is its an internal function but it can't be protected since it has to be called by other callers. This is a bit of an issue.

packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
soulFood5632: Yes we are see #1137

packages/railtracks/src/railtracks/built_nodes/llm/middleware/after_llm.py
soulFood5632: Honeslty, I think this splitting at the public APi perspective is ok. The reason I say that is that internally we still only have the wrapper types. So it follows a good fan in fan out principle to allow for user ease of use without sacrificing too much backend complexity.

packages/railtracks/src/railtracks/interaction/couple.py
soulFood5632: This was a convention I discovered a bit ago haha

packages/railtracks/src/railtracks/built_nodes/llm/middleware/after_llm.py
soulFood5632: Done

packages/railtracks/src/railtracks/interaction/couple.py
soulFood5632: complicated becuase for Node's it is a nbew type and for functions its a new object. Regardless I will update it

packages/railtracks/src/railtracks/middleware/chain.py
soulFood5632: yes you have to invert the order of the onion to make it work the way we want. 

[A, B, C] should be A -> B -> C -> Node -> C -> B -> A

in order to wrap things we need to start with `A(B(C(Node)))`

Amir-R25 (approved):
lgtm-looks-good-to-me

PR #1253, opened 2026-07-15
extras and unused and broken links

Amir-R25:
Thanks to @CoronRing's note, the current unused and outdated `system_internals` folder in our docs was causing confusion for AI coding assistants so they are removed in this PR.


soulFood5632 (approved):
LGTM

PR #1255, opened 2026-07-15
major structure updates to the guardrails api

### LINKED ISSUE #1246: Determine the Whether Guardrails should be user ordered or whether we should hide that
In this issue we should decide once and for all whether the interface for adding `guards` should a param based approach or if it should be user controller through the `middleware` params. This builds on the recent work completed by @Pooria90  and @soulFood5632  in #1238 and #1139. 

## Option 1 
```python 
rt.agent_node(
    "Agent",
    tool_nodes=[],
    middleware=[retry_middleware, middleware.MaxCreations(2)],
    model_middleware=[middleware.Retry(max_retries=10), ],
    guards=Guards(input= guards.MaxCost(100), guards.KeywordDetection("password"), guards.PIIDetection())
)
```

## Option 2
```python
rt.agent_node(
    "Agent",
    tool_nodes=[],
    middleware=[retry_middleware, middleware.MaxCreations(2)],
    model_middleware=[middleware.Retry(max_retries=10), guards.MaxCost(100), guards.KeywordDetection("password"), guards.PIIDetection()],
)
```

- Decide on the best interface 
- Polish and complete the public API

soulFood5632:
## Summary

In this PR I have removed the ability to pass in guard's using the Guard object and instead you now pass them in as individual middelware components. This allows users to directly control ordering so we don't do anything unexpected on the backend. 

### Before 
```python 
rt.agent_node(
   model_middleware=[...]
   guards=Guard(input_guards=..., output_guards=...)
)
```

### After
```python 
rt.agent_node(
   model_middleware=[guard.PIIDetectInput()]
)
```

The major backend change will be around the idea instead of having a GuardrailRunner object which takes in a list of guards and runs them all at once. Instead, we encapsulate the running running operation within the Guard itself. This encapsulation allows us to make them conform to the middleware type so they can be run within the context of middleware. 

## Guide to Review 
1. Understand the new inheritance structure of Guard classes (see how they all are subtypes of Middleware) (note also that the only exposed method is the __call__ method at the bottom for custom implementations)
2. Inspect deeply the refactor of all the logic in the GuardRunner to the new encapsulated version in BaseLLMGuardrails
3. Check out the changes to the creation of an agent through agent node
4. (for @Pooria90) please make sure i haven't deleted things from the public API that I shouldn't have

## Questions for Pooria
@Pooria90 Can you please ensure that there wasn't behavior that I missed that was a part of GuardRunner during the move. Is there any ordering pitfalls that I missed etc. Let me know if I have overlooked anything critical. 

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Breaking change
- [ ] Docs

## Checklist

## Notes

The new classes will work well with the coming observability. Look forward to that. 

The observaility work may require some addition changes to the current logic but most of that is left for future work.


soulFood5632:
@Pooria90 can you walk me through what the fail_open flag does. I think it is irrelvant now that we are running individual guards, but I could be wrong.

soulFood5632:
tests have yet to be fixed. I don't want to finalize until @Pooria90 gives me the green light.

Pooria90:
> @Pooria90 can you walk me through what the fail_open flag does. I think it is irrelvant now that we are running individual guards, but I could be wrong.

It determines guard behavior when something unusual happens (like other than GuardrailBlockedError or other decisions); like when a weird exception comes up. If it's true, the guardrail that raised the exception will be skipped, but the trace will be recorded. If it's false, the unusual exception will stop the whole workflow with GuardrailBlockedError.

We can still have it and change the scope to single guardrails.

packages/railtracks/src/railtracks/guardrails/core/interfaces.py
Pooria90: flagging the `BaseLLMGuardrail` since it's removed

packages/railtracks/src/railtracks/guardrails/core/interfaces.py
```diff
@@ -1,14 +1,22 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Any, Protocol
+from typing import Any, Awaitable, Callable, Generic, ParamSpec, Protocol, TypeVar
 
+from pydantic import BaseModel
+from railtracks.context.central import get_parent_id, get_run_id, is_context_present
```
Pooria90: Are these going to be needed here? 🤔

soulFood5632:
> > @Pooria90 can you walk me through what the fail_open flag does. I think it is irrelvant now that we are running individual guards, but I could be wrong.
> 
> It determines guard behavior when something unusual happens (like other than GuardrailBlockedError or other decisions); like when a weird exception comes up. If it's true, the guardrail that raised the exception will be skipped, but the trace will be recorded. If it's false, the unusual exception will stop the whole workflow with GuardrailBlockedError.
> 
> We can still have it and change the scope to single guardrails.

Ok makes sense. I can move it to be a optional flag on the creation of the guardrail. Thanks for the explanation.

packages/railtracks/src/railtracks/guardrails/core/interfaces.py
soulFood5632: No they are not but its a long story. I forgot to `ruff` my last PR, and I didn't want to clutter the diff with a big clean. Once we feel good about this PR, I will `ruff` the entire thing and introduce all the formatting changes. I just wanted to keep the review as clean as possible for now.

packages/railtracks/src/railtracks/guardrails/core/interfaces.py
soulFood5632: Yep good catch thanks

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
```diff
@@ -0,0 +1,292 @@
...
+    def __call__(self, event: LLMGuardrailEvent) -> GuardrailDecision:
+        """Evaluate the event and return a decision. Implemented by each concrete guard."""
+        pass
+
+    @abstractmethod
+    def convert(self, value: str | Any | MessageHistory | LLMGuardrailEvent, /) -> LLMGuardrailEvent:
+        """Build an event from a raw value. Implemented per phase by InputGuard and OutputGuard."""
+        pass
+
+    def decide(self, value: str | Any | MessageHistory | LLMGuardrailEvent, /) -> GuardrailDecision:
```
Pooria90: This typing here is my previous work, blame me for this! 
I think we should replace `Any` with `Message`

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
```diff
@@ -0,0 +1,292 @@
...
+    ) -> MessageHistory:
+        """Convert str, Message, or MessageHistory to a MessageHistory."""
+        if isinstance(input, MessageHistory):
+            return input
+        if isinstance(input, str):
+            return MessageHistory([UserMessage(input)])
+        if isinstance(input, Message):
+            return MessageHistory([input])
+        raise TypeError(
+            f"Expected str, Message, MessageHistory, or LLMGuardrailEvent, "
```
Pooria90: Also this one message doesn't need to mention Event

soulFood5632:
closes #1246

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
```diff
@@ -0,0 +1,292 @@
...
+
+    def _handle_rail_exception(
+        self,
+        *,
+        exc: Exception,
+        traces: list[GuardrailTrace],
+        value: _TValue,
+        reason_prefix: str,
+    ) -> (
+        tuple[Literal["continue"], _TValue] | tuple[Literal["stop"], _TValue, GuardrailDecision]
```
Pooria90: "continue" only happens when `fail_open=True` which currently is removed from the code.

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
```diff
@@ -0,0 +1,292 @@
...
+                    f"Guardrail {self._rail_name()} returned {type(decision).__name__}, expected GuardrailDecision."
+                )
+        except Exception as e:
+            outcome = self._handle_rail_exception(
+                exc=e,
+                traces=traces,
+                value=value,
+                reason_prefix="Guardrail raised exception",
+            )
+            if outcome[0] == "continue":
```
Pooria90, line 194: This case does not happen now without `fail_open`

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
```diff
@@ -0,0 +1,292 @@
...
+            },
+        )
+        return ("stop", value, block)
+
+    def run(
+        self,
+        *,
+        event: LLMGuardrailEvent,
+        value: _TValue,
+    ) -> tuple[_TValue, list[GuardrailTrace], GuardrailDecision | None]:
```
Pooria90, line 151: This was `list[GuardrailTrace]` because run used to execute a chain of guardrails. For now, it should be just one trace.

packages/railtracks/tests/unit_tests/guardrails/test_runner.py
```diff
@@ -8,12 +8,12 @@
     Guard,
     GuardrailAction,
     GuardrailDecision,
-    GuardRunner,
     InputGuard,
     LLMGuardrailEvent,
     LLMGuardrailPhase,
     OutputGuard,
 )
+from railtracks.guardrails.core.GuardRunner import GuardRunner
```
Pooria90: flagging this: removed already

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
soulFood5632: Yep I had that exact comment when working through it :)

packages/railtracks/tests/unit_tests/guardrails/test_runner.py
soulFood5632: Yep my test fixes are coming in full force soon

todelete.py
Pooria90, line 1: oops

packages/railtracks/src/railtracks/guardrails/llm/concrete.py
```diff
@@ -0,0 +1,247 @@
...
+from pydantic import BaseModel
+
+from railtracks.llm.history import MessageHistory
+from railtracks.llm.message import AssistantMessage, Message
+from railtracks.llm.response import Response
+from railtracks.llm.tools.tool import Tool
+from railtracks.middleware.core import Middleware
+
+from ..core.decision import GuardrailDecision
+from ..core.event import LLMGuardrailEvent, LLMGuardrailPhase
```
Pooria90, line 20: flagging repeated imports

Pooria90 (changes requested):

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
soulFood5632: I moved it back in :)

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
soulFood5632: It does now :)

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
soulFood5632: tech debt to solve later. I'll link an issue

packages/railtracks/src/railtracks/guardrails/llm/concrete.py
soulFood5632: beautiful

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
soulFood5632: done

packages/railtracks/src/railtracks/guardrails/llm/concrete.py
```diff
@@ -0,0 +1,247 @@
...
+
+from typing import Awaitable, Callable
+
+from pydantic import BaseModel
+from railtracks.guardrails.core.decision import GuardrailDecision
+from railtracks.guardrails.core.event import LLMGuardrailEvent, LLMGuardrailPhase
+from railtracks.llm.history import MessageHistory
+from railtracks.llm.response import Response
+from railtracks.llm.tools.tool import Tool
+from railtracks.utils.logging.create import get_rt_logger
```
Pooria90: This is making a circular import with InputGuard in llm_guard.py

packages/railtracks/src/railtracks/guardrails/llm/concrete.py
```diff
@@ -0,0 +1,224 @@
...
+
+        new_messages, traces, decision = self.run(event=event, value=message_history)
+        self._record_guard_traces(traces)
+        self._raise_if_blocked(decision, traces)
+
+        return new_messages, schema, tools
+
+    def convert(
+        self,
+        value: str | Any | MessageHistory | LLMGuardrailEvent,
```
Pooria90, line 73: `Any` should be `Message` here

packages/railtracks/src/railtracks/guardrails/llm/concrete.py
```diff
@@ -0,0 +1,224 @@
...
+        new_message, traces, decision = self.run(event=event, value=result.message)
+        self._record_guard_traces(traces)
+        self._raise_if_blocked(decision, traces)
+
+        if new_message is result.message:
+            return result
+
+        return Response(message=new_message, message_info=result.message_info)
+
+    def convert(self, output: str | Any | MessageHistory | LLMGuardrailEvent, /):
```
Pooria90, line 165: same here

packages/railtracks/src/railtracks/guardrails/llm/llm_guard.py
```diff
@@ -0,0 +1,310 @@
...
+    def decide(
+        self, value: str | Message | MessageHistory | LLMGuardrailEvent, /
+    ) -> GuardrailDecision:
+        """Convert value to an event and run this guard on it directly, outside a chain."""
+        converted_event = self.convert(value)
+
+        return self(converted_event)
+
+    @staticmethod
+    def _coerce_to_message_history(
```
Pooria90, line 70: same here

Pooria90 (approved):
Good job Logan. I just added minor comments for typing. Looks good to me.

Issue #1239, opened 2026-07-10
Rethinking observability into an events stream


Amir-R25:
### Description & Design

Parent ticket to track a new event streaming observability module to replace the current "in-memory state object json dump upon completion"

PR #1256, opened 2026-07-15
Observability Feature 1: Event Model, Observer pub/sub class, and Writer protocol

### LINKED ISSUE #1240: Required pieces: `Event` data model, `Observer` Submodule, and `jsonl` Writer
### Description
In this work we would like to add:
- `observability` submodule
- `Event` datatype:
    - Data model:
      ```python
      event_id: str
      event_type: str # "llm.call" | "retrieval.query" | "evaluator.metric_result" | "session.start" | "evaluator.aggregate_result" ...
      timestamp: datetime.timestamp
      scope_type: str # "session" | "retrieval" | "evaluation"
      scope_id: str  # "session_id" | "evaluation_id"
      parent_scope_id: str | None
      payload: dict
      ```
- `Observer` class that handles the event stream queue.
-  Writer Protocol 
    - `jsonl` Writer
    - Writer exceptions are caught and logged inside the consumer
- Out of scope: `publish_event` api, `contextvar` reads, emission sites in scopes, and viz integration (DuckDB)

### Requirements
- This module should be independent of the rest of the framework
- Observer should be able to separate queues for different writers.

Amir-R25:
## Summary

`railtracks.observability` submodule:
  - Feature 1 of the observability epic #1239 . This work adds an independent streaming-event pipeline: 
  - `Event` model, `Observer` with per-writer queues 
  - `Writer` protocol, and a `JsonlWriter` that appends per-scope `.jsonl` files 
- Closes #1240.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

This work is purely additive, no existing code paths change so it's safe to merge to main directly.

## Demo
  1. Fan-out from a single publish() to multiple writers, with per-scope
     routing landing events in .../{scope}.jsonl files.
  2. Drop-oldest policy under overload and a WARNING logs per drop, so a
     stalled writer becomes visible instead of silently growing memory.

```python
import asyncio
import logging
import shutil
from pathlib import Path

from railtracks.observability import (
    SCOPE_EVALUATION,
    SCOPE_RETRIEVAL,
    SCOPE_SESSION,
    Event,
    JsonlWriter,
    Observer,
    QueuePolicy,
    Stamp,
)

logging.basicConfig(level=logging.WARNING, format="%(levelname)s %(name)s: %(message)s")

DATA_DIR = Path("/tmp/rt-obs-demo")


class PrintWriter:
    """A tiny custom writer — 4 lines is all the Writer protocol asks for."""

    async def start(self) -> None: ...

    async def write(self, event: Event) -> None:
        print(f"  [stdout]  {event.scope_type:<10} {event.event_type:<22} scope_id={event.scope_id}")

    async def shutdown(self) -> None: ...


class SlowWriter:
    """Sleeps on every write so the publisher outpaces the queue."""

    async def start(self) -> None: ...

    async def write(self, event: Event) -> None:
        await asyncio.sleep(0.1)

    async def shutdown(self) -> None: ...


def make_event(
    scope: str,
    event_type: str,
    scope_id: str,
    parent_scope_id: str | None = None,
    **payload,
) -> Event:
    return Event(
        event_type=event_type,
        stamp=Stamp.now(),
        scope_type=scope,
        scope_id=scope_id,
        parent_scope_id=parent_scope_id,
        payload=payload,
    )


async def demo_fan_out_and_routing() -> None:
    print("\n=== 1. Fan-out to two writers + per-scope routing ===\n")
    shutil.rmtree(DATA_DIR, ignore_errors=True)

    async with Observer() as obs:
        await obs.register(JsonlWriter(DATA_DIR), "jsonl")
        await obs.register(PrintWriter(), "stdout")

        s = "session-42"
        await obs.publish(make_event(SCOPE_SESSION, "session.start", s, hello="world"))
        await obs.publish(make_event(SCOPE_RETRIEVAL, "retrieval.query", "ret-1", s, q="what is x"))
        await obs.publish(make_event(SCOPE_RETRIEVAL, "retrieval.query", "ret-2", s, q="what is y"))
        await obs.publish(make_event(SCOPE_EVALUATION, "evaluator.metric_result", "eval-1", s, metric="accuracy", score=0.87))
        await obs.publish(make_event(SCOPE_SESSION, "session.end", s))

    print(f"\n  Files written under {DATA_DIR}:")
    for path in sorted(DATA_DIR.glob("*.jsonl")):
        print(f"    {path.name}: {len(path.read_text().splitlines())} events")


async def demo_drop_oldest_policy() -> None:
    print("\n=== 2. Drop-oldest policy: publisher outpaces a slow writer ===\n")
    async with Observer() as obs:
        await obs.register(
            SlowWriter(),
            "slow",
            maxsize=3,
            policy=QueuePolicy.DROP_OLDEST,
        )
        for i in range(10):
            await obs.publish(make_event(SCOPE_SESSION, "session.tick", f"tick-{i}"))
        print("\n  (shutdown will drain the surviving events before returning...)")


async def main() -> None:
    await demo_fan_out_and_routing()
    await demo_drop_oldest_policy()


if __name__ == "__main__":
    asyncio.run(main())

```


packages/railtracks/src/railtracks/observability/models.py
```diff
@@ -0,0 +1,30 @@
...
+from typing import Any
+
+from pydantic import BaseModel, Field
+
+SCOPE_SESSION = "session"
+SCOPE_RETRIEVAL = "retrieval"
+SCOPE_EVALUATION = "evaluation"
+
+
+class Stamp:
```
soulFood5632: this is gonna be a namespace disaster so we either will want to combine the types or choose a new name.

packages/railtracks/src/railtracks/observability/models.py
```diff
@@ -0,0 +1,30 @@
...
+    @staticmethod
+    def now() -> datetime:
+        return datetime.now(timezone.utc)
+
+
+class Event(BaseModel):
+    event_id: str = Field(default_factory=lambda: str(uuid.uuid4()))
+    event_type: str
+    stamp: datetime
+    scope_type: str
```
soulFood5632, line 25: I like the idea of flexability here but we could probably also throw a `Literal` here as well.

packages/railtracks/src/railtracks/observability/observer.py
```diff
@@ -0,0 +1,156 @@
...
+        if name not in self._writers:
+            raise KeyError(f"No writer registered as {name!r}.")
+        await self._teardown(name)
+
+    async def publish(self, event: Event) -> None:
+        if not self._running:
+            raise RuntimeError("Observer is not running.")
+        for name, entry in self._writers.items():
+            try:
+                entry.queue.put_nowait(event)
```
soulFood5632, line 100: why do we require the `no_wait` condition on this?

packages/railtracks/src/railtracks/observability/observer.py
```diff
@@ -0,0 +1,156 @@
...
+    """Marker pushed onto a writer's queue to tell its consumer task to drain and exit."""
+
+
+_END = _EndOfStream()
+
+
+@dataclass
+class _Entry:
+    writer: Writer
+    queue: asyncio.Queue[Any]
```
soulFood5632: Pin it down with Generics :). We can make Observer, _Entry and share a generic type so when you access the elements from queue we know they are of type Event

packages/railtracks/src/railtracks/observability/observer.py
```diff
@@ -0,0 +1,156 @@
...
+logger = get_rt_logger(__name__)  # for now, we can make another logger later
+
+
+class QueuePolicy(Enum):
+    """How a writer's queue behaves when it's full at publish time."""
+
+    DROP_OLDEST = "drop_oldest"
+
+
+class _EndOfStream:
```
soulFood5632, line 20: Noting this now but we will send a final confirmation of exiting a session as an event as well.

packages/railtracks/src/railtracks/observability/observer.py
```diff
@@ -0,0 +1,156 @@
...
+
+    async def _teardown(self, name: str) -> None:
+        entry = self._writers.pop(name)
+        self._drops.pop(name, None)
+        _enqueue_end(entry.queue)
+        await entry.task
+        await entry.writer.shutdown()
+
+    async def _consumer_loop(
+        self, name: str, writer: Writer, queue: asyncio.Queue[Any]
```
soulFood5632: Generics are best friend see above

soulFood5632 (dismissed):
Some comments. But looks good to me

packages/railtracks/src/railtracks/observability/models.py
Amir-R25: changed it to `Timestamp`

packages/railtracks/src/railtracks/observability/observer.py
Amir-R25: From emission side, we should not wait for the queues at all

packages/railtracks/src/railtracks/observability/observer.py
Amir-R25: yeah this one is mainly just to kill the queue

soulFood5632: Yep an internal type

packages/railtracks/src/railtracks/observability/observer.py
Amir-R25: pinning this down, not going with Generics but using `_QueueItem` instead of any which is tied to `Event | _EndOfStream`

packages/railtracks/src/railtracks/observability/models.py
Amir-R25: Let's keep it general so `observability` module is fully decoupled from its user

soulFood5632 (approved):
LGTM

Issue #1206, opened 2026-06-16
[Feature] [Middleware] Enhanced support for streaming


CoronRing:
### Feature Request

# Streaming
As a part of the middleware update, we need to fully support streaming action, includeing guards, wrappers, loggers, session file generation and last but not least the successful result (ie toolcall + response)

## Background
Streaming is when an LLM returns responses in chunks instead of in a single response (ModelStream vs ModelResponse). This behaviour is controlled by the provider, and a specialized request format is used to trigger this response behaviour (namely, `streaming` arg = True. ). 

## Current system
We only support LLM level streaming (`Openai(streaming=true)`), nodes welcome the return of the streaming object, but there is no detailed handling and requires manual surface path.

## CoS
1. Support Middleware's wrapper and gate objecy, which should have ability to act on each chunk and return a chunk.
2. Maintain LLM level streaming (`Openai(streaming=true)` or `openai().stream`). 
3. Should support stream at Node level (either stream = true or .stream)
4. Should support stream at flow level
5. Should support callback system (Push)
6. Should support direct return of a generator (Pull)
7. A high level strema command (ie top node) should not make all low-level streams on defaut (slower, more compute). There should be a way for the user to specify which node streams, yet prevent double trigger

## V3 Design
1. *No `stream` flag on the model or on `agent_node`.** Streaming is requested only at the call site — `astream` (pull) or an attached push consumer.
  - Simplify interface, stream is also treated less of a behaviour modification, but just return type modification (actually not true, stream changes agent response in some cases, but can commonly thought this way)
2. **`ModelBase`: no generic, separate methods.
  - ** `chat` / `structured` / `chat_with_tools` → `Response`.
  -  `stream_chat` / `stream_structured` / `stream_chat_with_tools` → `Generator[str | Response, None, Response]`.
  - Clean no overload interface
3. **`ModelInvoker`: one `ModelSource`** (`ModelBase | Callable[[], ModelBase]`),
  - Clean no overload interface
4. **Streaming built on the sync path**
  - No complex threading loop and concurrency.

## Other framework analysis

| Framework | Where the stream switch lives | Mechanism | Does it change the return type? | Sync/async | Caller model |
|---|---|---|---|---|---|
| **LiteLLM** | Per **call** | `completion(..., stream=True)` | **Yes** — `ModelResponse` → `CustomStreamWrapper` iterator | `completion` / `acompletion` | pull |
[...]

PR #1257, opened 2026-07-16
1206: model-level per-call streaming API (astream_*, Response.text, litellm bridge)

### LINKED ISSUE #1259: Migrate to sync api from litellm.
### Description

litellm's synchronous `completion(stream=True)` is better maintained than its async
counterpart (the async path is prone to leaking noisy logs/warnings), so railtracks
ideally should migrate to depend on sync api for all tasks.

CoronRing:
Work item: #1206 #1259

## Summary

First of two stacked PRs splitting the call-scoped streaming work (#1206) for easier review. This PR adds the **model-level per-call streaming primitives** — the self-contained bottom layer that the framework PR builds on. It imports nothing from `pubsub`/`context`/`interaction`, so it can be reviewed and merged in isolation.

This workitem also attemp a full migration from the async api litellm poorly maintains.

The follow-up PR (**call-scoped streaming — `rt.astream`, channels, `stream_callback`**) targets this branch and adds the framework-level API, node integration, and docs.

## What's included

- **`ModelBase.astream_chat` / `astream_chat_with_tools` / `astream_structured`** — per-call streaming: async generators that yield `str` chunks then a final `Response`. Accept `MessageHistory` or a plain `Sequence[Message]`. Independent of the deprecated constructor `stream` flag.
- **`LiteLLMWrapper` streaming bridge** — streaming rides litellm's synchronous `completion(stream=True)` API, bridged onto the event loop via a sync→async queue; a per-call `stream=` override on `_invoke`/`_ainvoke`. The `self: LiteLLMWrapper[Literal[...]]` overloads that made internal calls unbindable were removed.
- **`Response.text`** — typed accessor for text content (raises `TypeError` on tool-call/structured content), so downstream code stays pylance-clean.
- **`TOOL_CALLING_STREAMING_BLACKLIST`** (`llm/providers.py`) — providers that fall back to buffered tool-calling at call time (with a warning) instead of failing at node construction.
- **Deprecate `ModelBase(stream=True)`** — emits a `DeprecationWarning`; buffered paths drain any legacy generator to a complete `Response`.
- **Fix:** `check_message_history` mistyped `SystemMessage` (`llm/message.py`).

## Testing

- `tests/unit_tests/llm/models/test_litellm_wrapper.py`, `test_retry_wrapper.py`, `models/conftest.py` updated for the new async surface (async chat now runs the sync retry loop on a worker thread).
- **39 passed locally** on this branch in isolation (`pytest tests/unit_tests/llm/models/test_litellm_wrapper.py tests/unit_tests/llm/test_retry_wrapper.py`).

## Review order

Merge **this PR first**, then the framework PR (which targets this branch and will retarget to `feature-branch-node-add-on` once this lands).


CoronRing:
## Self-review notes 

1. **async → thread-pool concurrency change (main one to watch).** Consolidating onto the sync `_invoke` means the whole async surface (`_achat` / `_astructured` / `_achat_with_tools`) now runs via `asyncio.to_thread`, i.e. the default `ThreadPoolExecutor` (`min(32, cpu+4)` workers) — not just streaming. This is a deliberate tradeoff (the sync litellm path is better maintained and doesn't leak the noisy logs/warnings the async path does), but it effectively caps concurrent model calls at the pool size, where we previously had native non-blocking `acompletion`. Follow-up: benchmark high-fan-out node runs and, if needed, give model calls a dedicated/larger executor.

2. **Blacklist not wired at the wrapper layer.** `TOOL_CALLING_STREAMING_BLACKLIST` is consumed at the framework level (model_invoker / _tool_call_base in the follow-up PR), but the public `astream_chat_with_tools` / `_astream_chat_with_tools` don't consult it. A direct `model.astream_chat_with_tools(...)` against a blacklisted provider (Anthropic/Gemini/etc.) will attempt to stream tool calls with no buffered fallback. Consider honoring the blacklist (or at least documenting the limitation) at the wrapper layer since these are public methods. This is the same as `streaming_blacklist` object which will be resolved in a separate ticket, this is kept for correctness of this very PR for structure/pass of test. It will be removed later.

3. **Unbounded bridge queue / no backpressure.** `_bridge_sync_stream` uses an unbounded `asyncio.Queue` and the worker never blocks (`put_nowait`). Fine for typical responses, but a slow/paused consumer that doesn't close the generator lets the worker buffer the full response in memory. Could add a bounded queue with threadsafe backpressure later.

4. **Early-break latency.** On early break the worker only observes `stop` after the next chunk arrives (it's blocked in `for item in gen` on network I/O), so the worker thread + network stream can linger until one more token lands. Correct, just not immediate.

5. **Streaming retry scope.** With `stream=True`, `call_with_retry` wraps only the call that returns the `CustomStreamWrapper` — mid-stream (token iteration) failures aren't retried, same as before, informational.

### Test coverage gaps (covered next PR)
- Mid-stream error propagation (worker raises *during* iteration, not just at open).
- `astream_chat_with_tools` streaming path.
- The deprecated `stream=True` → `_achat` bridge path.
- Minor: the deprecation warning now fires from test fixtures that construct with `stream=True` — a `filterwarnings` marker would keep CI logs clean.

packages/railtracks/src/railtracks/llm/model.py
soulFood5632: If you are getting rid of the streaming flag we should just get rid of generic type as it now has no value. Further lets just get rid of the flag set here anyway as it will make typing impossible

packages/railtracks/src/railtracks/llm/model.py
```diff
@@ -63,6 +66,16 @@ def __init__(
         else:
             exception_hooks = __exception_hooks
 
+        if stream:
```
soulFood5632: To my above comment, we can say goodbye to backwards compatability so don't worry about deprecation here.

packages/railtracks/src/railtracks/llm/model.py
```diff
@@ -338,11 +357,134 @@ async def achat_with_tools(self, messages: MessageHistory, tools: List[Tool]):
...
+    def _coerce_message_history(
+        messages: MessageHistory | Sequence[Message],
+    ) -> MessageHistory:
+        """Normalizes a plain sequence of messages into a MessageHistory."""
+        if isinstance(messages, MessageHistory):
+            return messages
+        return MessageHistory(list(messages))
+
+    async def astream_chat(
+        self, messages: MessageHistory | Sequence[Message]
```
soulFood5632: We should not allow anything but MessageHistory types. No need to add extra flexability for just this method unless we are gonna add it to all methods

packages/railtracks/src/railtracks/llm/model.py
```diff
@@ -338,11 +357,134 @@ async def achat_with_tools(self, messages: MessageHistory, tools: List[Tool]):
...
+
+        ```python
+        async for item in model.astream_chat([UserMessage("hi")]):
+            if isinstance(item, str):
+                ...  # token chunk
+            else:
+                final = item  # the terminal Response
+        ```
+
+        Tip: inside a node, prefer `await rt.broadcast_stream(model.astream_chat(...))`, which
```
soulFood5632: We shouldn't put referecnes of how to work things in rt in our llm module.

packages/railtracks/src/railtracks/llm/model.py
```diff
@@ -338,11 +357,134 @@ async def achat_with_tools(self, messages: MessageHistory, tools: List[Tool]):
...
         return response
 
+    # ================ START Streaming (per-call) LLM calls ===============
+    # These methods request a streamed response for a single call, regardless of how the model
+    # was constructed. They are the model-level building blocks of railtracks streaming (see
+    # `rt.astream` and `rt.broadcast_stream`).
+
+    @staticmethod
+    def _coerce_message_history(
+        messages: MessageHistory | Sequence[Message],
```
soulFood5632: see my other comment but if we are to implement this should be true for all methods

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
soulFood5632: see above comment but there is no need for the generic now so we can get rid of it

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
```diff
@@ -167,38 +180,36 @@ def __init__(
...
-    ) -> Tuple[CustomStreamWrapper, float]:
-        pass
-
     def _invoke(
         self,
         messages: MessageHistory,
         *,
         response_format: Optional[Any] = None,
         tools: Optional[list[Tool]] = None,
+        stream: Optional[bool] = None,
```
soulFood5632: with no generic it should probably not be optional

soulFood5632: although see my above comment about a suggested overload comment, it may be helpful to keep optional to help with that typing

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
```diff
@@ -167,38 +180,36 @@ def __init__(
...
-    def _invoke(
-        self: LiteLLMWrapper[Literal[True]],
-        messages: MessageHistory,
-        *,
-        response_format: Optional[Any] = None,
-        tools: Optional[list[Tool]] = None,
-    ) -> Tuple[CustomStreamWrapper, float]:
-        pass
-
     def _invoke(
```
soulFood5632, line 179: We should be able to type this properly using overloads so we get rid of the union return type.

soulFood5632 (changes requested):
Looks generally good, couple of nits around typing structure

packages/railtracks/src/railtracks/llm/model.py
CoronRing: `ModelBase` is no longer `Generic[_TStream]`, the `stream` constructor param and
`self.stream` are gone, and all the `@overload`.

removed like 10 overload, makes me happt

packages/railtracks/src/railtracks/llm/model.py
CoronRing: Removed

CoronRing:
@soulFood5632 ifx according to request note Removed: `StreamingTerminalLLM`, `StreamingStructuredLLM`,
  `StreamingToolCallLLM`, `StreamingOutputLessToolCallLLM`, the three `GuardedStreaming*` as they all depend on streaming types, I do not feel like patching it, they will retire next PR anyways.

docs/scripts/streaming.py
```diff
@@ -1,44 +1,24 @@
...
-
-agent = rt.agent_node(
-    llm=rt.llm.OpenAILLM(model_name="gpt-4o", stream=True),
-)
+import railtracks.llm as llm
 
-# --8<-- [end: streaming_with_agents]
 
+async def main():
+    model = llm.OpenAILLM(model_name="gpt-4o")
```
soulFood5632: This isn't equaivelent code since the deleted stuff uses our actual agent_node api to do it whereas the stuff you added is at the llm level. I know we are implementing this in a future but I'd rather just keep the mark the docs as TODO instead of impleenting that could make it through our reviews.

soulFood5632 (changes requested):
In general looks good to me. Just be careful to not change docs that out of scope. We can handle proper docs updates when the user facing behavior gets changed

soulFood5632:
> @soulFood5632 ifx according to request note Removed: `StreamingTerminalLLM`, `StreamingStructuredLLM`, `StreamingToolCallLLM`, `StreamingOutputLessToolCallLLM`, the three `GuardedStreaming*` as they all depend on streaming types, I do not feel like patching it, they will retire next PR anyways.

Yea totally, I have actually deleted on the brancgh anyway so you don't need to worry about it anymore

soulFood5632 (approved):
LGTM

Issue #15, opened 2025-01-28
Build a Internal Typing for LLM interaction


soulFood5632:
As described in the parent ticket there is a need for internal module which handles interactions with llms. Without this internal structure any hope at building helpful tools will have to be custom built for each individual API. This layer is essential to the next phase of the parent ticket

PR #1262, opened 2026-07-17
1206 V4 streaming (1/n): rt.astream — call-scoped streaming for agents

CoronRing:
## Summary

First slice of the V4 streaming rollout — the public surface is **one feature**: `rt.astream(agent, ...)`. Strict subset of `1206/v4_streaming`; remaining surfaces ship in follow-up PRs so each is reviewable on its own.

> **Rebased onto the updated base.** The base now carries the model-level streaming plumbing from #1257 (`astream_*` API, `Response.text`, the litellm sync-stream bridge, `TOOL_CALLING_STREAMING_BLACKLIST`). This PR previously included that plumbing; it now contains **only the astream wiring on top of #1257** — 15 files, down from 24.

```python
stream = rt.astream(agent, user_input="Write a short poem about rain.")
async for chunk in stream:
    print(chunk, end="", flush=True)   # str token chunks
final = stream.result                  # the complete (gated) response

# or, when you only want the final result of a streamed run:
final = await rt.astream(agent, user_input="...")
```

**The rule this PR establishes: streaming is requested at the *call*, not baked into the agent.** The same agent object serves both modes — `rt.call` runs buffered (no streaming cost), `rt.astream` streams. Streaming is frame-local: nested `rt.call` children run buffered.

Works for terminal, tool-calling, and structured agents. Token streaming with tool calling is OpenAI-only for now; blacklisted providers fall back to a buffered model call with a warning (final result unaffected).

## Semantics

- `Stream` yields **only** `str` chunks; the node's return value is exposed separately via `.result` (no ambiguity between a chunk and the final value).
- `.result` is authoritative: output guardrails run on the complete buffered response, so it may differ from the concatenated chunks.
- Errors raise out of the `async for` / `await`, exactly like `rt.call`.
- Early `break` does not cancel the run — it completes in the background; `await stream` collects the result.
- The session `timeout` applies as a wall-clock limit on the whole streamed run.

## What this PR wires on top of #1257

- **pubsub**: `Streaming` gains `channel`/`stream_id`; `RequestCreation` gains `stream`/`current_stream_id`.
- **context**: `InternalContext.stream_enabled` (frame-local, never inherited) and `stream_id` (inherited scope id); `is_streaming_enabled()` / `get_stream_id()` accessors.
- **execution/state**: the flag rides `RequestCreation → RTState → Task → update_parent_id`.
- **ModelInvoker**: streaming branch consumes `model.astream_*` (already on base) and broadcasts each chunk, handing the complete `Response` back through the middleware chain — guardrails/observability operate on the buffered final.
- **`broadcast_stream`** (internal bridge, not exported): forwards model-stream chunks to the bus and returns the final `Response`; when nobody listens the chunks are dropped, so the same node works buffered and streamed.
- **`stream_callback`** (Session / Flow / `rt.set_config`): passive session-wide listener for the **stream lane** — the callback form of consuming tokens, alongside the pull form (`rt.astream`). Split from `broadcast_callback` (events) by a `kind` tag on `Streaming` messages, so neither lane is ever handed the other's traffic; neither enables streaming.
- **exports** `rt.astream`; `_call.py` threads `current_stream_id`. (`Stream` is the return type only and is intentionally not exposed at the package root.)

## Explicitly NOT in this PR (follow-ups)

`Stream.route()` (push), channels (`on_channel`, public `broadcast_stream`, `rt.broadcast(channel=)`), `Flow.astream`, `agent_node(stream_channel=...)`, `rt.context.get_stream` + `StreamEnd` markers, `stream=True` deprecation and `Streaming*` class retirement. Existing `stream=True` behavior is unchanged.

## Verification

- Unit suite green (the pre-existing `test_structured_tool_call_llm` failures were resolved on the new base; `test_nodes.py` collection error remains pre-existing on base).
- Integration (run + library): failure set **identical** to the base tip — no regressions (the 3 remaining failures are tool-calling/function tests untouched by this PR).
- Ruff clean on `src`; pyright `standard` clean on touched files (`state.py` / `_call.py` carry pre-existing base errors — `_astream.py` mirrors base `_call.py`'s own accepted `node = node_` pattern).
- Runtime smokes (mock models): pull iteration, awaitable stream, buffered `rt.call` unaffected, frame-locality, mid-stream error propagation, and absent-surface checks for all follow-up APIs.

## Note on typing (`agent_node` input)

`agent_node` keeps its `type[Node[[UserInput], …]]` return, so the **positional** call form
type-checks: `rt.astream(agent, "prompt")` / `rt.call(agent, "prompt")`. The **keyword** form
`rt.astream(agent, user_input="prompt")` runs fine but does not statically type-check, because
`[UserInput]` is a positional-only ParamSpec. Making the keyword form check cleanly (a Protocol
with a named parameter, or a more general signature) is deferred to future work rather than
changing the shared `Node` generic here.


docs/advanced_usage/config.md
```diff
@@ -20,7 +20,8 @@ Configuration parameters follow a specific precedence order, allowing you to ove
 ### Advanced Settings
 
 - **`context`** (`Dict[str, Any]`): Global context variables for execution
-- **`broadcast_callback`** (`Callable`): Callback function for broadcast messages
+- **`broadcast_callback`** (`Callable`): Passive listener for one-off `rt.broadcast` **events**
```
soulFood5632: These files have been deleted from main at this point. can you just not change them so we don't have conflicts later

docs/integrations/llms/streaming.md
```diff
@@ -4,6 +4,48 @@
...
+```
+
+A `Stream` is also awaitable — `await stream` consumes it to completion and returns the final result:
+
+```python
+--8<-- "docs/scripts/streaming.py:astream_await"
+```
+
+!!! Note "The final result is authoritative"
+    `stream.result` may differ from the concatenation of the streamed chunks — for example when output guardrails gate or correct the buffered response after the raw tokens were streamed. Treat the chunks as live progress and `.result` as the answer.
```
soulFood5632: em dash should be removed

docs/integrations/llms/streaming.md
```diff
@@ -4,6 +4,48 @@
 
 Streaming is a way to make your agent feel more responsive. Instead of waiting for the complete response, you can stream intermediate results as they arrive.
 
+## Streaming an Agent Run — `rt.astream`
+
+Streaming is requested at the **call**, not baked into the agent. The same agent object serves streaming and non-streaming runs:
+
+- `rt.call(agent, ...)` runs buffered — no streaming overhead, no chunks.
```
soulFood5632: em dash

docs/integrations/llms/streaming.md
soulFood5632, line 1: These docs are generally ok, but they need more of a human touch. Lost of em dashes and stacatto language so it doesn't read very naturally. If you could go through and write them with more conciseness that would be great.

docs/integrations/llms/streaming.md
```diff
@@ -4,6 +4,48 @@
...
+- **`stream_callback`** — a passive session-wide listener for **stream chunks** (`rt.broadcast_stream` productions, which include all LLM token streaming). This is the callback form of consuming tokens; `rt.astream` is the pull form.
+- **`broadcast_callback`** — a passive session-wide listener for one-off **events** published with `rt.broadcast` (progress notes, tool events). It is *never* flooded with tokens, even while a run streams.
+
+Both are set on the `Flow` (or globally via `rt.set_config`), and neither one *enables* streaming — only `rt.astream` does.
+
+```python
+--8<-- "docs/scripts/streaming.py:stream_callback"
+```
 
 ## Streaming Support
```
soulFood5632: this is out of date now

docs/scripts/streaming.py
```diff
@@ -1,3 +1,49 @@
...
+
+
+# --8<-- [start: stream_callback]
+# Passive session-wide listeners — neither enables streaming (only rt.astream does):
+#   stream_callback    -> chunks from rt.broadcast_stream (LLM tokens included)
+#   broadcast_callback -> one-off events from rt.broadcast (progress notes, ...)
+observed_flow = rt.Flow(
+    name="poet_observed",
+    entry_point=agent,
+    stream_callback=lambda chunk: print(chunk, end="", flush=True),
```
soulFood5632, line 35: there is leakage of the a future module here. we haven't yet implemented the stream callback

docs/scripts/streaming.py
```diff
@@ -1,3 +1,49 @@
...
+    broadcast_callback=lambda event: print(f"[event] {event}"),
+)
+
+
+async def main_observed():
+    # tokens flow to stream_callback while the run streams; the broadcast_callback
+    # only ever sees explicit rt.broadcast events (never the token chunks)
```
soulFood5632, line 42: comments here reference the unbuilt module

packages/railtracks/src/railtracks/built_nodes/llm/node.py
```diff
@@ -95,7 +95,10 @@ def agent_node(
     manifest: ToolManifest | None = None,
     middleware: list[Middleware[[UserInput], StringResponse]] | None = None,
     model_middleware: list[ModelMiddleware] | None = None,
-) -> type[Node[[UserInput], StringResponse]]: ...
+    # `...` (not `[UserInput]`) so the canonical keyword style — rt.call / rt.astream(
```
soulFood5632: While I am ok with a more general solution removing the generic is not a good solution. This change is also adverse to what was said in your PR descriuption. Regadrless. lets address this in future work

packages/railtracks/src/railtracks/built_nodes/llm/node.py
```diff
@@ -110,7 +113,10 @@ def agent_node(
     middleware: list[Middleware[[UserInput], StructuredResponse[_TBaseModel]]]
     | None = None,
     model_middleware: list[ModelMiddleware] | None = None,
-) -> type[Node[[UserInput], StructuredResponse[_TBaseModel]]]: ...
+    # `...` (not `[UserInput]`) so the canonical keyword style — rt.call / rt.astream(
+    # agent, user_input=...) — type-checks alongside the positional style; the node
+    # accepts its input either way at runtime.
+) -> type[Node[..., StructuredResponse[_TBaseModel]]]: ...
```
soulFood5632: same up here. 

Also please as a style comment. Comments like this will be rejected. This comment is explaining a change that the codebase no longer has so it has no value to a reader of the code base and is just another headache to maintain.

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -199,11 +246,12 @@ async def shutdown_publisher():
     This function should be called to stop the publisher and clean up resources.
     """
     context = safe_get_runner_context()
-    context = context.internal_context
-    assert context is not None
+    internal_context = context.internal_context
```
soulFood5632, line 249: Can you please manage scope creep a bit better. We can leave in this his unrelated change for this PR, but please just watch for this.

The main thing when working on a feature branch is your work should be mainly focussed on additions and only touch core modules unless absolutely needed otherwise it can induce massive conflicts down the line.

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -240,7 +288,8 @@ def set_local_config(
     """
     context = safe_get_runner_context()
 
-    context.executor_config = executor_config
+    # the config lives on the internal context (RunnerContextVars has no such attribute)
```
soulFood5632, line 291: again ditto here. Good catch on this error but this needs to be merged into main into a feature branch

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -360,6 +425,9 @@ def set_config(
     broadcast_callback: (
         Callable[[str], None] | Callable[[str], Coroutine[None, None, None]] | None
     ) = None,
+    stream_callback: (
```
soulFood5632, line 428: scope creep here

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -369,7 +437,12 @@ def set_config(
     - If you call this function after the runner has been created, it will not affect the current runner.
     - This function will only overwrite the values that are provided, leaving the rest unchanged.
 
-
+    Args:
+        broadcast_callback: A passive listener for one-off events published with `rt.broadcast`.
+            Stream chunks go to `stream_callback` instead.
+        stream_callback: A passive listener for stream chunks published through
```
soulFood5632, line 443: ditto

packages/railtracks/src/railtracks/interaction/__init__.py
```diff
@@ -8,6 +9,8 @@
 __all__ = [
     "call",
     "call_batch",
+    "astream",
+    "Stream",
```
soulFood5632: do we need to expose Stream on the public API?

packages/railtracks/src/railtracks/interaction/_astream.py
```diff
@@ -0,0 +1,338 @@
...
+        self._queue: asyncio.Queue[tuple[str, Any]] | None = None
+        self._started = False
+        self._finished = False
+        self._result: Any = _UNSET
+        self._error: BaseException | None = None
+        self._sub_id: str | None = None
+        self._deadline: float | None = None
+        self._timeout: float | None = None
+        # a session this stream is responsible for closing once the run completes
+        self._owned_session: Session | None = owned_session
```
soulFood5632: This logic really scares me from a design perective. the idea that a child can close out a parent is a bit worrisome to me. The logic for closing out a session should really be in the State class and not here

packages/railtracks/src/railtracks/interaction/_astream.py
```diff
@@ -0,0 +1,338 @@
...
+@overload
+def astream(
+    node_: RTFunction[_P, _TOutput],
+    *args: _P.args,
+    **kwargs: _P.kwargs,
+) -> Stream[_TOutput]: ...
+
+
+def astream(
+    node_: type[Node[_P, _TOutput]] | RTFunction[_P, _TOutput],
```
soulFood5632, line 283: You cannot stream a function node so no need to include that as a possible input

packages/railtracks/src/railtracks/interaction/_astream.py
```diff
@@ -0,0 +1,338 @@
...
+        return self.result
+
+    def __await__(self) -> Generator[Any, None, _TOutput]:
+        """Awaiting a Stream consumes it to completion and returns the final result."""
+        return self._drain().__await__()
+
+
+@overload
+def astream(
+    node_: type[Node[_P, _TOutput]],
```
soulFood5632: Are these overloads not pointless?

packages/railtracks/src/railtracks/interaction/_astream.py
```diff
@@ -0,0 +1,338 @@
...
+            elif isinstance(message, RequestFinishedBase):
+                if message.request_id == request_id:
+                    queue.put_nowait(("done", message))
+            elif isinstance(message, FatalFailure):
+                queue.put_nowait(("done", message))
+
+        # subscribe BEFORE publishing the request so no chunk can be missed
+        self._sub_id = publisher.subscribe(_subscriber, name="astream subscriber")
+
+        self._timeout = get_local_config().timeout
```
soulFood5632, line 165: This is sligthly flawed logic. Timeout is a global object and you are applying it it to a single node. Probably just ignore for now

packages/railtracks/src/railtracks/interaction/_astream.py
soulFood5632, line 1: I am bit confused by some of the purpose of the logic being here why do we have manually manage the startupo and closedown logic here. All of this should be housed in a session object and not here. I left some comments but most of the confusion is around that

soulFood5632: as a bit of an aside, this was kinda the reason I wanted to just focus on handling thing within a session first (using just `astream`. Once your cross the session barrier we need to think about what you are doing here where you have manage the closing out a of a session.

packages/railtracks/src/railtracks/interaction/broadcast_.py
```diff
@@ -1,16 +1,94 @@
...
+
 
 async def broadcast(item: str):
     """
-    Streams the given message
+    Broadcasts a one-off **event** to the session bus.
+
+    This triggers the `broadcast_callback` you have already provided. Events ride a lane
+    separate from stream chunks (`rt.broadcast_stream` / LLM token streaming): a
+    `broadcast_callback` receives events only and is never flooded with tokens.
```
soulFood5632: Most of the dscotring reads more as an explanation of what you did and why instead of what it used for.

packages/railtracks/src/railtracks/interaction/broadcast_.py
```diff
@@ -1,16 +1,94 @@
...
+            if isinstance(item, str):
+                parts.append(item)
+                await _publish_chunk(item)
+            else:
+                final = item
+
+    if final is not None:
+        return final
+
+    return "".join(parts)
```
soulFood5632: Why not fail fast here and raise an exception. The callers of this function in your PR expect it to be a Response anyway, so why allow it be a string anyway

packages/railtracks/src/railtracks/orchestration/flow.py
```diff
@@ -45,6 +46,9 @@ def __init__(
         broadcast_callback: (
             Callable[[str], None] | Callable[[str], Coroutine[None, None, None]] | None
         ) = None,
+        stream_callback: (
```
soulFood5632, line 49: scope leak

packages/railtracks/src/railtracks/pubsub/messages.py
```diff
@@ -175,14 +196,50 @@ def __repr__(self):
         return f"{self.__class__.__name__}(error={self.error})"
 
 
+# The two kinds of broadcast traffic, kept on separate lanes so consumers never mix them up.
```
soulFood5632: Most of this comment follows bad abstractions. The message type should really now about the subscribers that use it so it shouldn't be in the comment here. Please remove since it doesn't add much

packages/railtracks/src/railtracks/pubsub/messages.py
```diff
@@ -175,14 +196,50 @@ def __repr__(self):
...
+    Args:
+        streamed_object: The item being streamed (typically a `str` chunk).
+        node_id: The id of the node that emitted the item.
+        channel: The named channel this item was emitted on. Consumers can filter/route on
+            this name. Defaults to `"default"`.
+        stream_id: The stream scope this item belongs to. This is the request id of the entry
+            frame that was invoked with streaming enabled (see `rt.astream`), or None if the
+            item was broadcast outside any streaming scope.
+        kind: What produced this item: `"event"` for a one-off `rt.broadcast`, `"stream"` for
+            a chunk of an `rt.broadcast_stream` production (LLM token streams included). The
```
soulFood5632: ditto here. the docstring shouldn't reference other parts of code base. You are essentially coupling this code to the code calls it by including it in docstring

packages/railtracks/src/railtracks/pubsub/messages.py
```diff
@@ -175,14 +196,50 @@ def __repr__(self):
...
 
+# The two kinds of broadcast traffic, kept on separate lanes so consumers never mix them up.
+# "event": a one-off item published with `rt.broadcast` (progress notes, tool events, ...) —
+# this is what `broadcast_callback` observes. "stream": one chunk of a continuous production
+# published through `rt.broadcast_stream`, which includes all LLM token streaming — this is
+# what `rt.astream` consumes and, deliberately, what `broadcast_callback` ignores.
+StreamingKind = Literal["event", "stream"]
+
+
 class Streaming(RequestCompletionMessage):
```
soulFood5632, line 204: From a design perspective why did you not just create a new type called TokenStereaming so the new logic could be contained within that new object? It also would allow for more flexibility in the future event streaming and token Streaming diverge. I'm curious to hear your take

packages/railtracks/src/railtracks/pubsub/messages.py
```diff
@@ -175,14 +196,50 @@ def __repr__(self):
...
-    A message that indicates a streaming operation in the request completion system.
+    A message carrying a single streamed item (e.g. an LLM token chunk or a user broadcast).
+
+    Args:
+        streamed_object: The item being streamed (typically a `str` chunk).
+        node_id: The id of the node that emitted the item.
+        channel: The named channel this item was emitted on. Consumers can filter/route on
+            this name. Defaults to `"default"`.
+        stream_id: The stream scope this item belongs to. This is the request id of the entry
+            frame that was invoked with streaming enabled (see `rt.astream`), or None if the
```
soulFood5632: again here the docstring references some other module that uses it. Keep the those types of dependencies out of the docstring

packages/railtracks/src/railtracks/utils/config.py
```diff
@@ -13,6 +13,9 @@ def __init__(
         broadcast_callback: (
             Callable[[str], None] | Callable[[str], Coroutine[None, None, None]] | None
         ) = None,
+        stream_callback: (
```
soulFood5632, line 16: scope leak

packages/railtracks/src/railtracks/__init__.py
```diff
@@ -71,7 +73,7 @@
 from .built_nodes.llm.middleware import after_llm, before_llm, wrap_llm
 from .context.central import session_id, set_config
 from .guardrails import input_guard, output_guard
-from .interaction import broadcast, call, call_batch, couple
+from .interaction import Stream, astream, broadcast, call, call_batch, couple
```
soulFood5632: see above comment about exposing the Stream type publically

packages/railtracks/src/railtracks/_session.py
```diff
@@ -215,13 +225,21 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
 
         if self.executor_config.subscriber is not None:
             self.publisher.subscribe(
-                stream_subscriber(self.executor_config.subscriber),
-                name="Streaming Subscriber",
+                stream_subscriber(self.executor_config.subscriber, kind="event"),
+                name="Broadcast Callback Subscriber",
+            )
+
+        if self.executor_config.stream_callback is not None:
```
soulFood5632, line 239: scope creep

soulFood5632 (changes requested):
Thanks for the great work Guan. I have left a couple of comments. From a high level I like the approach. There is a couple concerns I had around where logic was located. 

I would love to talk about the decision to put so much handling of the session object in the interaction.astream function. We descoped this addition so the astream must exist within a session so I don't see the need for such complexity. 

Similarly, the items you added to the `Streaming` object type could have utilized new objects instead of adding a to previous one. Maybe a subclass could work

I left a bunch of comments flagging scope creep. There are active changes elsewhere in the codebase that these additions will complicate, so it's fine to leave in for this PR, but worth watching for going forward.

packages/railtracks/src/railtracks/interaction/_astream.py
CoronRing: This was somewhat of a compatibility thing for astream, as astream can be wrapped as a function and gets handled; hence, it is technically possible to use a streamer as an RTFunction; it will just be buffered (as the return is largely the same).

packages/railtracks/src/railtracks/interaction/_astream.py
CoronRing: These are actually thrown by pyright, ie error:

```
"astream" is marked as overload, but additional overloads are missingPylance[reportInconsistentOverload](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportInconsistentOverload.md)
```

Unless I mask/multi-type, which seems beyond this ticket. Same as call interface

packages/railtracks/src/railtracks/interaction/_astream.py
CoronRing: image

Is the logic not the same as _call? It was attached to object because we need to calculate total time for entire call rather than each chunk.

packages/railtracks/src/railtracks/interaction/_astream.py
soulFood5632: If you get rid of both overloads wouldn't that error not exist?

packages/railtracks/src/railtracks/interaction/_astream.py
soulFood5632: That branch that uses is for when it is a top level request:

```python 
    # if the context is not active then we know this is the top level request
    if not is_context_active():
        result = await _start(node, args=args, kwargs=kwargs)
        return result

    # if the context is active then we can just run the node
    result = await _run(node, args=args, kwargs=kwargs)
    return result
```

packages/railtracks/src/railtracks/interaction/_astream.py
CoronRing: Some of these logic are just for stuff that was removed, like flow, which is heavy on session management and detect session in session etc. I can do a full redesign, but honestly it will be very painful. I was hoping to just "turn off" some of the public api stuff and leave internals largely the same so its easier. 

Floating code that does not get used is bad, but it's also pain having to write 2-3 versions of the same file that I know will be replaced

packages/railtracks/src/railtracks/pubsub/messages.py
CoronRing: A single `isinstance(msg, Streaming)` check and filter on one field keeps that path really simple for this review, This was actually a custom object before, but it contained a lot of custom logic like termination type etc. I removed it hoping it would be another PR.

packages/railtracks/src/railtracks/pubsub/messages.py
CoronRing: I honestly think stuff like this is reasonable, its not speic reference but just tells one wheer the spec is. I added it because stream_id cannot colide, and writing that multiple times is really not DRY

docs/scripts/streaming.py
CoronRing: It is, in some sense, as right as everything is still streamed to callback with `kind`, I thought having stream_callback will be helpful for observability stuff, ie seeing each chunk. Otherwise, there is no default way to provide visibility unless we intercept streams, just final response, which is enough for now I guess.
 

Let me know if you think if this should be removed entirely or just for interface like flow

packages/railtracks/src/railtracks/built_nodes/llm/node.py
CoronRing: I reverse the change, but again pyright is rejecting the base version: 

```
Overloaded implementation is not consistent with signature of overload 1
  Type "(name: str | None = None, *, tool_nodes: Iterable[type[Node[..., Unknown]] | RTFunction[..., Unknown]] | None = None, output_schema: type[_TBaseModel@agent_node] | None = None, llm: ModelSource, system_message: SystemMessage | str | None = None, manifest: ToolManifest | None = None, middleware: list[Middleware[(UserInput), StructuredResponse[_TBaseModel@agent_node] | StringResponse]] | None = None, model_middleware: list[ModelMiddleware] | None = None) -> (type[Node[(UserInput), StringResponse]] | Unknown)" is not assignable to type "(name: str | None = None, *, tool_nodes: Iterable[type[Node[..., Unknown]] | RTFunction[..., Unknown]] | None = None, output_schema: None = None, llm: ModelSource, system_message: SystemMessage | str | None = None, manifest: ToolManifest | None = None, middleware: list[Middleware[(UserInput), StringResponse]] | None = None, model_middleware: list[ModelMiddleware] | None = None) -> type[Node[(UserInput), StringResponse]]"
    Keyword parameter "middleware" of type "list[Middleware[(UserInput), StringResponse]] | None" is incompatible with type "list[Middleware[(UserInput), StructuredResponse[_TBaseModel@agent_node] | StringResponse]] | None"
      Type "list[Middleware[(UserInput), StringResponse]] | None" is not assignable to type "list[Middleware[(UserInput), StructuredResponse[_TBaseModel@agent_node] | StringResponse]] | None"
        Type "list[Middleware[(UserInput), StringResponse]]" is not assignable to type "list[Middleware[(UserInput), StructuredResponse[_TBaseModel@agent_node] | StringResponse]] | None"
          "list[Middleware[(UserInput), StringResponse]]" is not assignable to "list[Middleware[(UserInput), StructuredResponse[_TBaseModel@agent_node] | StringResponse]]"
          "list[Middleware[(UserInput), StringResponse]]" is not assignable to "None"Pylance[reportInconsistentOverload](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportInconsistentOverload.md)
node.py(88, 5): Overload signature is defined here
``

packages/railtracks/src/railtracks/built_nodes/llm/node.py
CoronRing: Reversed

packages/railtracks/src/railtracks/__init__.py
CoronRing: I was looking at session and other Param type objects I though we wanted they for type exposure, ie response: Stream = xyz, but I guess StringResponse is not returned so we are keeping the interface minimal.

packages/railtracks/src/railtracks/interaction/_astream.py
CoronRing: The reason why it has to apply to both flow level and this single node is dirty:
asyncio.wait_for on its completion future and astream doesn't use that path at all (it pulls a queue), so that timeout never fires for a streamed run (ie infinitely returns empty and do not terminate a connection (Ie that company that starts with R). The __anext__ deadline re-creates the same caller-side give-up timer for the queue loop; without it, a hung run waits forever.

CoronRing:
Summary of task completed:


| # | Location | Comment (short) | Status | Notes |
|---|----------|------------------|--------|-------|
| 1 | config.md:23 | file deleted on main  | **Done** | `config.md` reverted to base (no edits). |
| 2 | streaming.md:27 | Remove em dash  | **Done** | `streaming.md` rewritten. |
| 3 | streaming.md:11 | Em dash  | **Done** | `streaming.md` rewritten. |
| 4 | streaming.md:1 | Needs human touch, fewer em dashes  | **Done** | Prose rewritten, em dashes removed. |
| 5 | streaming.md:50 | "Out of date now"  | **Done** | Stale `stream=True` sections removed from `streaming.md`; orphaned snippets removed from `streaming.py` (the flag no longer exists on base). |
| 6 | streaming.py:35 | Leakage of a future module (stream_callback)  | **Skipped** | `stream_callback` kept (your call); the snippet is intended. |
| 7 | streaming.py:42 | Comments reference unbuilt module  | **Skipped** | Ditto. |
| 8 | node.py:98 | Removing the generic isn't a good solution; adverse to PR description  | **Done** | Reverted to `type[Node[[UserInput], …]]`; keyword typing deferred to future work. Positional call form type-checks; keyword runs but shows a Pylance note. Will resolve in future PR |
| 9 | node.py:119 | Comment-style: comments explaining a removed change  | **Done** | No diff-narrating comments remain; the stray autocomplete artifact on the impl line was removed. |
| 10 | central.py:249 | Scope creep (`set_local_config` fix)  | **Skipped** | Fix kept in-PR (your call); not extracted. |
| 11 | central.py:291 | Good catch, but merge via main  | **Skipped** | Fix kept in-PR (your call). |
| 12 | central.py:428 | Scope creep (`set_config` stream_callback)  | **Skipped** | `stream_callback` kept. |
| 13 | central.py:443 | Ditto  | **Skipped** | `stream_callback` kept. |
| 14 | interaction/__init__.py:13 | Do we need `Stream` on the public API?  | **Done** | `Stream` removed from the interaction package exports. |
| 15 | _astream.py:121 | Child closing a parent session is worrisome  | **Done** | Session lifecycle moved out of `Stream` into `astream` via `on_start`/`on_close` hooks; `Stream` no longer touches `Session`. |
| 16 | _astream.py:291 | Can't stream a function node, drop it as input  | **Skipped** | `RTFunction` input kept (your call); function nodes can broadcast_stream. |
| 17 | _astream.py:276 | Are these overloads pointless?  | **Done** | Overloads removed; single signature. |
| 18 | _astream.py:158 | Timeout is global, applied to a single node  | **Skipped** | Kept intentionally — it is the only wall-clock bound on the `astream` path (`rt.call`'s `wait_for` timeout never runs for a streamed run). See detail above. |
| 19 | _astream.py:1 (×2) | Manual startup/closedown should live in a session  | **Done** | Addressed with #15 (creation/termination moved to `astream`). |
| 20 | broadcast_.py:18 | Docstring explains what you did, not what it's for  | **Done** | `broadcast` docstring rewritten usage-first. |
| 21 | broadcast_.py:94 | Fail fast / raise instead of returning `str`  | **Done** | `str` return kept (your call); reintroduce narrowing when the public `broadcast_stream` lands. Signature narrowed Union[Response, str] → Response. Issue: no long support warning + single chunk failure, if we notice this in future, we can have another ticket to expand |
| 22 | flow.py:49 | Scope leak (stream_callback)  | **Skipped** | `stream_callback` kept. |
| 23 | messages.py:199 | Comment couples message to its subscribers  | **Done** | `StreamingKind` comment trimmed to describe the values only. |
| 24 | messages.py:220 | Docstring references other code  | **Done** | `Streaming` docstring made self-contained. |
| 25 | messages.py:207 | Why not a `TokenStreaming` subclass? | **Skipped** | `kind` tag kept; revisit as a subclass if the lanes diverge (future). |
| 26 | messages.py:217 | Docstring references a module that uses it  | **Done** | Docstring self-contained (no `rt.astream` reference). |
| 27 | messages.py (dup :217/:496) | Ditto  | **Done** | Docstring self-contained. |
| 28 | config.py:16 | Scope leak (stream_callback / kind alias)  | **Skipped** | `stream_callback` kept. |
| 29 | __init__.py:76 | Exposing `Stream` publicly  | **Done** | `Stream` removed from the package-root exports (`rt.Stream` gone). |
| 30 | _session.py:239 | Scope creep (stream_callback subscriber)  | **Skipped** | `stream_callback` kept. |

soulFood5632 (approved):
LGTM

PR #1270, opened 2026-07-17
debug/output-rail-roundtrip

Pooria90:
## Summary

Restores a guardrail behavior lost in the `GuardRunner` → per-guard middleware migration (#1255): output guards must skip intermediate tool-call turns and fire only on the final reply. The old `guardrail_output_middleware` gate checked `_is_intermediate_tool_call` before running output rails; the new `OutputGuard._middleware_fn` ran on every model round-trip inside the tool-calling loop. This reinstates the check in `OutputGuard`: responses that request tools now pass through unguarded, exactly as before #1255.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

**Why the old behavior matters** — without the skip, two failure modes appear in tool-calling agents:

1. Text-oriented output guards (`BlockTextOutputGuard`, `PIIRedactOutputGuard`, …) evaluate intermediate tool-call messages whose `content` is a tool-call list rather than text — they can misfire or raise, and with the default `fail_open=False` a raised exception kills the whole request mid-loop.
2. A `TRANSFORM` rebuilds the `Response` wholesale from `decision.output_message`, which doesn't carry the original `tool_calls` — so a transforming guard on a tool-call turn silently amputates the tool calls, `process_message` reclassifies the turn as `"Content"`, and the agent terminates early with an internal step's text as its final answer.

**Changes**:
- `guardrails/llm/concrete.py`: added `_is_intermediate_tool_call` (same implementation as the deleted `guardrail_gates.py` helper, mirroring `process_message`) and an early return in `OutputGuard._middleware_fn`; class and method docstrings now state the guarantee.
- Two existing tests had codified the regression as intended behavior (asserting output guards fire twice per tool loop): `test_agent_node_guardrails.py` and `TestGuardrailsEndToEnd` in `test_middleware_integration.py`. Both renamed to `..._fires_only_on_final_reply` and now assert a single firing.
- New unit regression test `test_output_guard_skips_intermediate_tool_call_turns` in `test_guard_middleware_wiring.py`: a blocking guard never evaluates a tool-requesting response, and the response passes through by identity with tool calls intact.
- Remaining test-file churn is import-sort autofixes from `ruff check` (project config runs with fix enabled).

### DIFF
--- a/packages/railtracks/src/railtracks/guardrails/llm/concrete.py
+++ b/packages/railtracks/src/railtracks/guardrails/llm/concrete.py
@@ -114,11 +114,22 @@ class InputGuard(BaseLLMGuardrail[MessageHistory]):
         return event.model_copy(update={"messages": value})
 
 
+def _is_intermediate_tool_call(response: Response) -> bool:
+    """A model round-trip is *intermediate* (not the final reply) when it requests tools.
+
+    Mirrors ``process_message`` in ``llm_helpers`` (tool calls present => "Tool").
+    """
+    return len(response.message.tool_calls) > 0
+
+
 class OutputGuard(BaseLLMGuardrail[Message]):
     """Base for guardrails that run on LLM output (e.g. model response).
 
     Inspect ``event.output_message`` for the assistant message produced this turn.
     ``event.messages`` is conversation context and may not yet include that reply.
+
+    Intermediate tool-call turns pass through untouched, so output rails fire only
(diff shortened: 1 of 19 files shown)


soulFood5632 (approved):
LGTM, My bad for dropping it in my PR

PR #1280, opened 2026-07-20
first stab at updated stateful modelinvoker creation

### LINKED ISSUE #1245: Allow for `rt.couple` to also Modify the Model Middleware
With the changes in #1238 we created a new public method called `rt.couple` which created a new node type with the middleware now injected. 

I would like this `rt.couple` to also allow you to inject new params into the model middleware. not just the node level middleware. 

# Some Notes
- This is kinda a bit of pain becuase of how stateless our llm nodes are. In order to this properly you will need to leverage some intellegent abstraction principles since the `ModelInvoker` type is not stored anywhere in the class. This will be tricky

soulFood5632:
## Summary

Previously the `rt.couple` interface only allowed modifications to the node level middlware. In this PR I add the functionality to inject `model_middleware` as well. As part of this PR I completed the following:

- modified the llm invoke factory function so it creates a `ModelInvoker` type at invoke time instead at type creation time
- Added a property called `_user_model_middleware` to node so we can manage the middlewares at the node level. I also included a new method in the `Node` class to modify it
- Modified the typing logic for couple to allow for the injection of model middleware to llm types

#### Some code to play with

```python 
import railtracks as rt

from railtracks.guardrails.llm.input import PIIRedactInputGuard

guard = PIIRedactInputGuard()



@rt.before_llm
def print_message(message_history: rt.llm.MessageHistory, schema, tools):
    print(message_history)
    return message_history, schema, tools


@rt.after_llm
def print_response(response: rt.llm.Response):
    print(response.message)
    return response

@rt.wrap_llm
async def periodic_failure(llm_call, message_history, schema, tools):
    import random
    if random.random() < 0.5:
        raise Exception("Random failure")
    return await llm_call(message_history, schema, tools)

@rt.wrap_node
async def retry_middleware(call, *args, **kwargs):
    import asyncio
    for i in range(3):
        try:
        except Exception as e:
            print(f"Attempt {i+1} failed: {e}")
            await asyncio.sleep(1)
    raise Exception("All attempts failed")

@rt.after_node
    return result

agent = rt.agent_node(
    "Agent",
    tool_nodes=[],
    middleware=[retry_middleware],
    model_middleware=[retry_middleware, periodic_failure],
)


adjusted = rt.couple(agent, middleware=[print_after], model_middleware=[print_message, guard, print_response])

flow = rt.Flow("puppy", agent)

flow.invoke("puppy")
```

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] Breaking change
- [x] Docs

## Checklist


packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
```diff
@@ -33,21 +34,21 @@
...
-        self,
+        self: Node,
         user_input: MessageHistory | UserMessage | str | list[Message],
     ) -> StringResponse: ...
 
 
 class StructuredLLMInvoke(Protocol[_TStructured]):
     async def __call__(
-        self,
+        self: Node,
```
Amir-R25, line 44: :D

Amir-R25 (approved):
LGTM

PR #1286, opened 2026-07-22
feat(evaluations): add runtime evaluator

### LINKED ISSUE #1163: [Feature] [Evaluation] `RuntimeEvaluator`
### Feature Request

We would like to add an Evaluator that tracks a total time across an agent invocation/flow. 

This way users can benchmark the wall clock time it takes for different models/tools etc on a bigger picture for their respective flows.

### Proposed Solution
- extend https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/evaluations/evaluators/evaluator.py
- Note: https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/evaluations/evaluators/llm_inference_evaluator.py achieves similar results but on a per LLM call basis but it can be a good starting point
- Update visualizer to handle this

cc: @mbrailtown @jbueza-railtownai 

### Additional Context

_No response_

jstar0:
## Summary

Add a `RuntimeEvaluator` for comparing end-to-end wall-clock time across complete agent invocations. Run boundaries are retained on root agent data points, while open or incomplete runs are omitted instead of being reported as zero.

The evaluator reuses the existing numerical aggregate result, so serialized evaluation output includes minimum, maximum, mean, median, mode, and standard deviation without introducing a visualizer-specific result type. Legacy session payloads without per-run boundaries remain supported.

Closes #1163

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

## Verification

- `pytest -q packages/railtracks/tests/unit_tests/evaluations packages/railtracks/tests/integration_tests/evaluations`
- `ruff check . --no-fix`
- `ruff format --check .`
- `mkdocs build --strict`

## Notes

Runtime is recorded once for each run's root agent so nested agents do not duplicate a flow-level measurement. Failed runs with complete timing boundaries remain measurable; open and malformed runs are skipped.

### DIFF
--- a/docs/evaluations/evaluators/runtime_evaluator.md
+++ b/docs/evaluations/evaluators/runtime_evaluator.md
@@ -1,0 +1,15 @@
+The **`RuntimeEvaluator`** reports the wall-clock duration of complete agent invocations. Use it to compare end-to-end flow performance across models, tools, and agent configurations instead of measuring individual LLM or tool calls.
+
+## Usage
+
+```python
+--8<-- "docs/scripts/evaluations/runtime_evaluator.py"
+```
+
+## Metric Tracked
+
+| Metric | Description |
+|---|---|
+| `Runtime` | Wall-clock execution time per agent invocation (seconds). |
+
+Runs without a recorded runtime are omitted from the metric instead of being reported as zero. The evaluation payload includes minimum, maximum, mean, median, mode, and standard deviation across measured invocations for visualization.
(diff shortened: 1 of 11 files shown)


soulFood5632:
@jstar0 Thanks so much for contribution. My sincerest apologies for the slow response in getting back to you. We are currently in the middle of a shift in our roadmap, and we will get back to as soon as possible with a review. I will task @Amir-R25 with giving some preliminary comments.

Amir-R25:
Hey, thanks for the PR! Quick heads up, the `feature-branch-observability-update` branch is reworking how new evaluators plug into the pipeline, and the shape of that surface is changing enough that landing this one first would just create rebase pain on both sides. I'm going to mark the issue as blocked on that branch releasing and pick this back up once it's in. Really appreciate the contribution, will loop back here as soon as we're unblocked.

PR #1288, opened 2026-07-22
Context Managers for Id management 

### LINKED ISSUE #1272: Create Railtracks SessionContext object to encapsulate all the context_vars flowing in a system
### Description

As part of this issue we need to complete the following. 

- A thorough review of where certain `context_vars` are set (talking specifically about ids, and other elements specific to a Session). 
- A decision around whether a new `SessionContext` object would be relevant and helpful 
- Introduce new id -> (middleware id)
- Implementation of the new object
- Connect the new object to all locations in the code that previously used ad hoc methods like `get_parent_id` etc. 
- Implement some safety checks on the setting of variables (think assertation to maintain good state in the system)

soulFood5632:
## Summary

New id structure
image

This PR restructures how we track ids stored in context vars. I'll cover the new primitives first, then how they compose into the public API.

The core new type is a stack-like mechanism that lets us push middleware and node ids:

```python
@dataclass(frozen=True)
class ScopeLink(Generic[T]):
    """Immutable linked chain; each link carries a reference to its parent."""

    value: T
    parent: ScopeLink[T] | None = None

    def pushed(self, value: T) -> "ScopeLink[T]":
        return ScopeLink(value=value, parent=self)

    def find(self, predicate: Callable[[T], bool]) -> T | None:
        link: ScopeLink[T] | None = self
        while link is not None:
            if predicate(link.value):
                return link.value
            link = link.parent
        return None
```

> We push middleware ids, node ids, and a marker for "the node has started executing" onto this chain (more detail in the Notes section below).

This is stored on our `SessionContext` object, which keeps the `ScopeLink` itself private. Rather than exposing it directly, `SessionContext` provides read-only properties (`current_node_id`, `current_middleware_id`, `scope`, ...) plus a small API for pushing new entries. The callers never touch the chain structure directly.

The key method for pushing a new entry:

```python
def with_scope_pushed(
        self, entry: ScopeEntry, *, run_id: str | None = None
    ) -> SessionContext:
        """Returns a new SessionContext with `entry` pushed onto the scope chain."""
        new_scope = ScopeLink(value=entry, parent=self._scope)
        resolved_run_id = self._run_id if run_id is None else run_id

        return SessionContext(
            session_id=self._session_id,
            run_id=resolved_run_id,
            publisher=self._publisher,
            scope=new_scope,
            executor_config=self._executor_config,
        )
```

Rather than making `with_scope_pushed` the only entry point, I wanted a clearer abstraction per scope type (new node, new middleware, node execution starting) so callers don't need to construct `ScopeEntry`s themselves. That abstraction is `ScopeManager`:

```python
class ScopeManager(Protocol):
    """Interface Middleware/Node use to signal scope changes."""

    def enter_node(self, node_id: str) -> ContextManager[None]: ...

    def enter_node_body(self) -> ContextManager[None]: ...

    def enter_middleware(self, name: str) -> ContextManager[str | None]: ...
```

### Concrete implementation

```python
class ContextVarScopeManager:
    """ScopeManager backed by the `runner_context` ContextVar."""

    @contextmanager
    def enter_node(self, node_id: str):
        ctx = safe_get_runner_context()
        established_run_id = (
            ctx.session_context.run_id
            if ctx.session_context.run_id is not None
            else node_id
        )
        token = _push_scope(
            ScopeEntry(ScopeKind.NODE, node_id), run_id=established_run_id
        )
        try:
            yield
        finally:
            runner_context.reset(token)

    @contextmanager
    def enter_node_body(self):
        ctx = safe_get_runner_context()
        node_id = ctx.session_context.current_node_id
        if node_id is None:
            raise RuntimeError(
                "Cannot enter a node-body scope outside of an active node scope"
            )
        token = _push_scope(ScopeEntry(ScopeKind.NODE_BODY, node_id.id))
        try:
            yield
        finally:
            runner_context.reset(token)

    @contextmanager
    def enter_middleware(self, name: str):
        middleware_id = str(uuid.uuid4())
        token = _push_scope(ScopeEntry(ScopeKind.MIDDLEWARE, middleware_id, name=name))
        try:
            yield middleware_id
        finally:
            runner_context.reset(token)
```

These context managers are what callers use to enter and exit scopes; they manage the underlying context var automatically. For example:

```python
        """
        Runs ``invoke`` through the node-level middleware.
        """

        async def body(*a: _P.args, **kw: _P.kwargs) -> _TOutput:
            with self._scope_manager.enter_node_body():
                return await self.invoke(*a, **kw)

        # reassigned per-call since Node.safe_copy() means __init__'s closure can go stale
        self.middleware.get_scope_manager = lambda: self._scope_manager
        return await self.middleware.run(body, *args, **kwargs)
```

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

- I opted for dependency injection of the scope manager. This keeps `middleware/` and `nodes/` decoupled from the concrete context implementation. Neither imports `railtracks.context` directly; only the composition root (`_session.py`) and `context/central.py` know about each other.

### Why a separate tag for "node started executing"

We need to track node ids and middleware ids on the stack, but it's less obvious why a *third* tag marking "the node has started executing" is needed

For observability, we need to tell apart a node created from within a middleware versus one created during the node's own execution. We also need each middleware invocation attributed to the node it's wrapping. Both of those depend on being able to answer, at any point in time, "is the top of the stack still inside middleware, or are we past it and into the node's own code?" That requires two distinct events, entering the node (before middleware runs) and the node body actually starting (after middleware runs), so they need two distinct tags.

### Logging updates

We now add the middleware id to the log record's extra fields:

```python
class RTContextLoggingAdapter(logging.LoggerAdapter):
    def process(self, msg, kwargs):
        try:
            parent_id = get_parent_id()
            run_id = get_run_id()
            session_id = get_session_id()
            middleware_id = get_middleware_id()

        except ContextError:
            parent_id = None
            run_id = None
            session_id = None
            middleware_id = None

        new_variables = {
            "node_id": parent_id,
            "run_id": run_id,
            "session_id": session_id,
            "middleware_id": middleware_id,
        }

        kwargs["extra"] = {**kwargs.get("extra", {}), **new_variables}

        return msg, kwargs
```


packages/railtracks/src/railtracks/context/central.py
Pooria90, line 62: Is this note needed as we don't expose session context manager?

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -90,7 +79,7 @@ def is_context_active():
         bool: True if the global variables are active, False otherwise.
     """
     context = runner_context.get()
-    return context is not None and context.internal_context.is_active
+    return context is not None and context.session_context.is_active
 
 
 def get_publisher() -> RTPublisher:
```
Pooria90, line 85: Typing should be `RTPublisher | None`

packages/railtracks/src/railtracks/execution/execution_strategy.py
Amir-R25, line 82: my type checker is complaining that this is returning `None` and the override expects a coroutine

Amir-R25, line 109: commented code?

Amir-R25 (approved):
LGTM

packages/railtracks/src/railtracks/nodes/nodes.py
```diff
@@ -137,7 +143,14 @@ async def wrapped_invoke(self, *args: _P.args, **kwargs: _P.kwargs) -> _TOutput:
         """
         Runs ``invoke`` through the node-level middleware.
         """
-        return await self.middleware.run(self.invoke, *args, **kwargs)
+
+        async def body(*a: _P.args, **kw: _P.kwargs) -> _TOutput:
```
Pooria90, line 148: SMART!

packages/railtracks/src/railtracks/nodes/nodes.py
Pooria90: I get a deprecation message on this:

> The function "iscoroutinefunction" is deprecated
  Deprecated since Python 3.14. Use `inspect.iscoroutinefunction()` instead.

soulFood5632: 3.14 user goes crazy

packages/railtracks/src/railtracks/execution/execution_strategy.py
soulFood5632: yes an aritifact and some dead code. we should delete it

packages/railtracks/src/railtracks/context/central.py
soulFood5632: Future PR. I will create the relevant issue

packages/railtracks/src/railtracks/context/central.py
soulFood5632: i'm gonna put this off a for a furture pr

packages/railtracks/src/railtracks/execution/execution_strategy.py
soulFood5632: dead code to delete later

Pooria90 (approved):
LGTM

PR #1295, opened 2026-07-23
1206 V4 streaming (2/n): named channels — rt.broadcast(channel=) + Stream.on_channel

CoronRing:
## Summary

Second slice of the V4 streaming rollout: **named channels**.

Every streamed item already travels on a named channel (`Streaming.channel`, from mini_1). This PR exposes that to users on both ends:

- **`rt.broadcast(item, channel="default")`** — the one-off event producer now takes a channel. It also stamps the current stream scope (`get_stream_id()`), so an event emitted inside a streamed run now reaches `rt.astream` as well as `broadcast_callback`. Events stay on the event lane (`kind="event"`); a `stream_callback` is never flooded.
- **`Stream.on_channel(name)`** — restrict a pull stream to a single channel. Chain it before iterating:

```python
async for token in rt.astream(agent, user_input="...").on_channel("default"):
    ...
```

Agent tokens ride `"default"`, so `on_channel("default")` selects just the tokens while a custom node emits progress events on another channel in the same run. By default (no `on_channel`) a stream still yields every channel.

## Why `on_channel` is a method, not `astream(..., channel=...)`

`astream` forwards its keyword arguments to the node (like `rt.call`), so a `channel=` keyword could collide with a node's own parameter. Selecting the channel on the returned handle keeps the node's argument list unambiguous. It must be called before iteration starts (raises `RuntimeError` otherwise).

### DIFF
--- a/docs/integrations/llms/streaming.md
+++ b/docs/integrations/llms/streaming.md
@@ -34,6 +34,23 @@ A few details worth knowing:
 - **Timeouts.** The session `timeout` applies to the whole streamed run as a wall-clock limit.
 - **Tool calling.** Token streaming with tool calling is currently supported on OpenAI models only. On other providers a streamed tool-calling run falls back to a buffered model call (with a logged warning), and the final result is unaffected.
 
+### Named channels
+
+Every streamed item travels on a named channel. An agent's tokens ride the `"default"` channel, and you can send your own one-off events on any channel with `rt.broadcast(item, channel=...)` from inside a run:
+
+```python
+--8<-- "docs/scripts/streaming.py:channels"
+```
+
+By default a `Stream` yields chunks from every channel. When a run produces on more than one channel (say, the agent's tokens on `"default"` and progress notes on `"status"`), chain `on_channel` to consume just one:
+
+```python
+async for token in rt.astream(agent, user_input="...").on_channel("default"):
+    ...
+```
+
+`on_channel` is a method rather than an `astream(..., channel=...)` keyword because `astream` forwards its keyword arguments to the node, where a `channel` keyword could collide with the node's own parameters.
+
 ### Two callback lanes: `broadcast_callback` and `stream_callback`
 
 Streaming shares the same pubsub bus as `rt.broadcast`, but the two kinds of traffic travel on separate lanes so a listener never receives the wrong kind:
--- a/docs/integrations/llms/streaming.md
+++ b/docs/integrations/llms/streaming.md
@@ -49,8 +49,6 @@ async for token in rt.astream(agent, user_input="...").on_channel("default"):
     ...
 ```
 
-`on_channel` is a method rather than an `astream(..., channel=...)` keyword because `astream` forwards its keyword arguments to the node, where a `channel` keyword could collide with the node's own parameters.
-
 ### Two callback lanes: `broadcast_callback` and `stream_callback`
 
 Streaming shares the same pubsub bus as `rt.broadcast`, but the two kinds of traffic travel on separate lanes so a listener never receives the wrong kind:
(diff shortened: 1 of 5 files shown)


docs/integrations/llms/streaming.md
```diff
@@ -34,6 +34,21 @@ A few details worth knowing:
 - **Timeouts.** The session `timeout` applies to the whole streamed run as a wall-clock limit.
 - **Tool calling.** Token streaming with tool calling is currently supported on OpenAI models only. On other providers a streamed tool-calling run falls back to a buffered model call (with a logged warning), and the final result is unaffected.
 
+### Named channels
+
+Every streamed item travels on a named channel. An agent's tokens ride the `"default"` channel, and you can send your own one-off events on any channel with `rt.broadcast(item, channel=...)` from inside a run:
```
soulFood5632, line 39: Why did you choose to use the default channel for the streaming of llm calls. Why not keep the previous behavior of `rt.broacast`. Not saying one is better, just curious your reasoning

docs/integrations/llms/streaming.md
```diff
@@ -34,6 +34,21 @@ A few details worth knowing:
...
+
+```python
+--8<-- "docs/scripts/streaming.py:channels"
+```
+
+By default a `Stream` yields chunks from every channel. When a run produces on more than one channel (say, the agent's tokens on `"default"` and progress notes on `"status"`), chain `on_channel` to consume just one:
```
soulFood5632, line 45: ngl, this text is really hard to read. I have no idea what it is trying to say.

packages/railtracks/src/railtracks/interaction/_astream.py
```diff
@@ -128,6 +131,40 @@ def __init__(
...
+        `on_channel` before iterating narrows it to one, which is useful when a run produces
+        on several channels (e.g. the agent's tokens on `"default"` and progress notes on
+        `"status"`) and you only want one of them:
+
+        ```python
+        async for token in rt.astream(agent, user_input="...").on_channel("default"):
+            ...
+        ```
+
+        It is a method rather than an `astream(..., channel=...)` keyword because `astream`
```
soulFood5632, line 150: This shound't be as part of the public api

packages/railtracks/src/railtracks/interaction/broadcast_.py
```diff
@@ -8,20 +8,29 @@
...
 
 
-async def broadcast(item: str):
+async def broadcast(item: str, channel: str = "default"):
     """
     Broadcasts a one-off **event** to the session bus.
 
-    This triggers the `broadcast_callback` you have already provided. Events are 
-    separate from stream callbacks.
+    The event is delivered to a `broadcast_callback` and, when emitted inside a streamed run,
```
soulFood5632, line 15: I think the complexity of this explanation is really difficult for me to follow. Remember the bar of understanding is not to explain to us, but so that user can use it with ease.

packages/railtracks/src/railtracks/interaction/_astream.py
soulFood5632: As in we don't need to explain why we chose to expose it the way we did in the docstring

packages/railtracks/src/railtracks/interaction/broadcast_.py
soulFood5632: Like to me I don't think I understand how to access the streams I want and how they are kept separate.

soulFood5632 (changes requested):
Super small questions. I think we should walk through DX because I don't think i follow it

soulFood5632:
I think it would be good to write out the DX for the following:

1. Streaming a single agent and consuming the generator
2. Streaming a agent and publishing rt.broadcast at the same time and consuming both seperately
3. Streaming 2 agents and consuming each of them indepdentely
4. Streaming 2 agent and publishing rt.broadcast at the same time and consuming both separately

soulFood5632:
Also one other question I had that may be addressed in a future PR. Lets say I had a parent node which called 2 separate nodes which both streamed to their own channels. From the top level flow, how could consume those streams? Generally what would that look like in code.

PR #1298, opened 2026-07-24
1296 add llm call

### LINKED ISSUE #1296: Add `llm_call_id` to `SessionContext`
### Description

We need to create and link a `llm_call_id` to the `SessionContext` object. This shoould be accessible and setting through a similar to API to the changes in #1272

soulFood5632:
## Summary

In this short PR I make a couple of adjustments to the ground work laid in #1288. In this PR i will be adding an additional id param for llm calls and adding that to the `SessionContext` object. Additionally I added a new context manager to be used to inject the id param. 

One important design consideration i would like to pay careful attention to:
- The location in which I decided to inject the `llm call id`. I chose to put it in the model invoker instead of wrapping it invoke body. This is done because `ModelInvoker`'s purpose is to inject observability and middleware into an llm. This feels like the natural location to put it in.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

closes #1296

### DIFF
--- a/packages/railtracks/src/railtracks/context/session_context.py
+++ b/packages/railtracks/src/railtracks/context/session_context.py
@@ -141,4 +141,5 @@ class SessionContext:
             publisher=self._publisher,
             scope=new_scope,
             executor_config=self._executor_config,
+            llm_call_id=self._llm_call_id,
         )
--- a/packages/railtracks/src/railtracks/context/session_context.py
+++ b/packages/railtracks/src/railtracks/context/session_context.py
@@ -39,8 +39,10 @@ class SessionContext:
         publisher: RTPublisher | None = None,
         scope: ScopeLink[ScopeEntry] | None = None,
         executor_config: ExecutorConfig,
+        llm_call_id: str | None = None,
     ):
         self._scope: ScopeLink[ScopeEntry] | None = scope
+        self._llm_call_id: str | None = llm_call_id
         self._publisher: RTPublisher | None = publisher
         self._session_id: str = session_id
         self._run_id: str | None = run_id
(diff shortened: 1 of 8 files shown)


packages/railtracks/src/railtracks/context/session_context.py
```diff
@@ -131,4 +141,5 @@ def with_scope_pushed(
             publisher=self._publisher,
             scope=new_scope,
             executor_config=self._executor_config,
+            llm_call_id=self._llm_call_id,
```
Amir-R25, line 144: nit: not sure what best practices are for accessing a param when you have a @property for it, just thought to point it out

packages/railtracks/src/railtracks/scope_manager.py
```diff
@@ -13,6 +13,8 @@ def enter_node_body(self) -> ContextManager[None]: ...
 
     def enter_middleware(self, name: str) -> ContextManager[str | None]: ...
 
+    def enter_llm_call(self) -> ContextManager[None]: ...
```
Amir-R25, line 16: this is nice extension. Also I'm wondering, do you foresee a "enter_tool_call" coming as well?

Amir-R25 (approved):
LGTM. Two comments for thoughts, no action needed

packages/railtracks/src/railtracks/scope_manager.py
Amir-R25: or that's just "enter_node" right?

packages/railtracks/src/railtracks/context/session_context.py
soulFood5632: Honestly, PEP recommends the opposite of what I have done but that can be a future issue. I didn't want to mess with convention.

packages/railtracks/src/railtracks/scope_manager.py
soulFood5632: that is just `enter_node`

PR #1301, opened 2026-07-24
fix: clean up central context typing

### LINKED ISSUE #1300: Clean up Typing in `context/central.py`
### Description

There is a total mess in this file mainly to with unpacking of `None`'s. This should be completely fixed so there is no type errors in the module.

ColumbusLabs:
## Summary

- add complete return and adapter annotations to `context/central.py`
- narrow publisher access without changing the Python 3.10 runtime surface
- remove unreachable `None` handling after `safe_get_runner_context()`
- fix `set_local_config()` to update the internal context that `get_local_config()` reads
- strengthen focused tests for the nested config update and synchronous publisher state check

Closes #1300.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [x] Lint & format pass (`ruff check . --no-fix && ruff format --check .`)
- [x] Docs updated if user-facing behavior changed — not applicable; no user-facing API change
- [x] Breaking changes include migration notes — not applicable; no breaking change

## Notes

Verification:

- strict mypy for `central.py`: 25 errors before, 0 after
- configured mypy for `central.py`: pass
- focused context tests: 15 passed, no warnings
- full unit suite: 1,858 passed, 5 skipped
- integration suite: 205 passed, 2 skipped; the two MCP fixture cases also passed after adding `pip` to the temporary uv environment, matching CI
- repository-wide Ruff lint and format: pass
- dependency ordering: pass
- strict MkDocs build: pass
- current function-node API smoke test: pass

### DIFF
--- a/packages/railtracks/src/railtracks/context/central.py
+++ b/packages/railtracks/src/railtracks/context/central.py
@@ -3,17 +3,38 @@ from __future__ import annotations
 import contextvars
 import logging
 import warnings
-from typing import TYPE_CHECKING, Any, Callable, Coroutine, KeysView
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Coroutine,
+    KeysView,
+    MutableMapping,
+    Protocol,
+    cast,
+)
 
 from railtracks.exceptions import ContextError
 
 if TYPE_CHECKING:
     from railtracks.pubsub.publisher import RTPublisher
 
+    _LoggerAdapter = logging.LoggerAdapter[logging.Logger]
+else:
+    _LoggerAdapter = logging.LoggerAdapter
+
 from railtracks.utils.config import ExecutorConfig
 
 from .external import ExternalContext, MutableExternalContext
 from .internal import InternalContext
+
+
+class _PublisherProtocol(Protocol):
+    async def start(self) -> None: ...
+
+    async def shutdown(self) -> None: ...
+
+    def is_running(self) -> bool: ...
 
 
 class RunnerContextVars:
@@ -26,11 +47,13 @@ class RunnerContextVars:
         *,
         internal_context: InternalContext,
         external_context: ExternalContext,
-    ):
+    ) -> None:
         self.internal_context = internal_context
         self.external_context = external_context
 
-    def prepare_new(self, new_parent_id: str, new_run_id: str | None = None):
+    def prepare_new(
+        self, new_parent_id: str, new_run_id: str | None = None
+    ) -> RunnerContextVars:
         """
         Update the parent ID of the internal context.
         """
@@ -76,13 +99,13 @@ def safe_get_runner_context() -> RunnerContextVars:
     return context
 
 
-def is_context_present():
+def is_context_present() -> bool:
     """Returns true if a context exists."""
     t_c = runner_context.get()
     return t_c is not None
 
 
-def is_context_active():
+def is_context_active() -> bool:
     """
     Check if the global variables for the current thread are active.
 
@@ -104,7 +127,9 @@ def get_publisher() -> RTPublisher:
         RuntimeError: If the global variables have not been registered.
     """
     context = safe_get_runner_context()
-    return context.internal_context.publisher
+    publisher = cast("RTPublisher | None", context.internal_context.publisher)
+    assert publisher is not None
+    return publisher
 
 
 def get_session_id() -> str | None:
@@ -132,7 +157,7 @@ def get_parent_id() -> str | None:
         ContextError: If the global variables have not been registered.
     """
     context = safe_get_runner_context()
-    return context.internal_context.parent_id
+    return cast("str | None", context.internal_context.parent_id)
 
 
 def get_run_id() -> str | None:
@@ -157,7 +182,7 @@ def register_globals(
     parent_id: str | None,
     executor_config: ExecutorConfig,
     global_context_vars: dict[str, Any],
-):
+) -> None:
     """
     Register the global variables for the current thread.
     """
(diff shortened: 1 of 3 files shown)


packages/railtracks/src/railtracks/context/central.py
```diff
@@ -3,19 +3,40 @@
...
+else:
+    _LoggerAdapter = logging.LoggerAdapter
+
 from railtracks.utils.config import ExecutorConfig
 
 from .external import ExternalContext, MutableExternalContext
 from .internal import InternalContext
 
 
+class _PublisherProtocol(Protocol):
```
soulFood5632: What is the purpose of this protocol. The RTFunction type should handle the typing you'd need for it

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -177,33 +202,29 @@ def register_globals(
...
 
     This function should be called to stop the publisher and clean up resources.
     """
     context = safe_get_runner_context()
-    context = context.internal_context
-    assert context is not None
-
-    assert context.publisher.is_running()
-    await context.publisher.shutdown()
+    publisher = cast("_PublisherProtocol | None", context.internal_context.publisher)
```
soulFood5632: see above comment

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -177,33 +202,29 @@ def register_globals(
...
     This function should be called to ensure that the publisher is running and can be used to publish messages.
     """
     r_c = safe_get_runner_context()
-    internal_context = r_c.internal_context
-    assert internal_context is not None
-
-    assert internal_context.publisher is not None
-
-    await internal_context.publisher.start()
+    publisher = cast("_PublisherProtocol | None", r_c.internal_context.publisher)
```
soulFood5632: ditto

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -104,7 +127,9 @@ def get_publisher() -> RTPublisher:
         RuntimeError: If the global variables have not been registered.
     """
     context = safe_get_runner_context()
-    return context.internal_context.publisher
+    publisher = cast("RTPublisher | None", context.internal_context.publisher)
```
soulFood5632: ditto

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -3,19 +3,40 @@
...
+    Protocol,
+    cast,
+)
 
 from railtracks.exceptions import ContextError
 
 if TYPE_CHECKING:
     from railtracks.pubsub.publisher import RTPublisher
 
+    _LoggerAdapter = logging.LoggerAdapter[logging.Logger]
```
soulFood5632, line 20: I don't think this needed for proper typing. Could you explain the logic behind that

soulFood5632 (changes requested):
Thanks for the PR. Have a couple of approach comments I would like to go through with you. Just curious to hear your reasoning

packages/railtracks/src/railtracks/context/central.py
ColumbusLabs: Kept intentionally: Python 3.10 evaluates a parameterized LoggerAdapter alias at runtime, while the logger is parameterized only for static checking. The alias preserves the 3.10-safe runtime base without weakening the typed annotation.

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -104,7 +118,9 @@ def get_publisher() -> RTPublisher:
         RuntimeError: If the global variables have not been registered.
     """
     context = safe_get_runner_context()
-    return context.internal_context.publisher
+    publisher = cast("RTPublisher | None", context.internal_context.publisher)
```
soulFood5632: The whole point of not using that protocol was so you didn't have to cast. Can you please review the requirements of casting in this new context without the protocol

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -177,33 +193,29 @@ def register_globals(
...
     Activate the publisher for the current thread's global variables.
 
     This function should be called to ensure that the publisher is running and can be used to publish messages.
     """
     r_c = safe_get_runner_context()
-    internal_context = r_c.internal_context
-    assert internal_context is not None
-
-    assert internal_context.publisher is not None
+    publisher = cast("RTPublisher | None", r_c.internal_context.publisher)
```
soulFood5632: ditto

packages/railtracks/src/railtracks/context/central.py
```diff
@@ -177,33 +193,29 @@ def register_globals(
...
 
     This function should be called to stop the publisher and clean up resources.
     """
     context = safe_get_runner_context()
-    context = context.internal_context
-    assert context is not None
-
-    assert context.publisher.is_running()
-    await context.publisher.shutdown()
+    publisher = cast("RTPublisher | None", context.internal_context.publisher)
```
soulFood5632: ditto

soulFood5632 (changes requested):
Great work, thanks for your comments. I still think there is a couple of uneeded casts that we would like to avoid.

packages/railtracks/src/railtracks/context/central.py
ColumbusLabs: Addressed in 0d4579e: I removed this cast. InternalContext.publisher now explicitly returns RTPublisher | None, so the existing assertion narrows it without a protocol or cast; the focused Python 3.10/mypy checks pass.

packages/railtracks/src/railtracks/context/central.py
ColumbusLabs: Addressed by the same property type change in 0d4579e: this cast is removed and the assertion now narrows RTPublisher | None directly.

packages/railtracks/src/railtracks/context/central.py
ColumbusLabs: Addressed by the same property type change in 0d4579e: this cast is removed and the assertion now narrows RTPublisher | None directly.

packages/railtracks/src/railtracks/context/internal.py
```diff
@@ -64,11 +64,11 @@ def parent_id(self, value: str):
         self._parent_id = value
 
     @property
-    def publisher(self):
+    def publisher(self) -> RTPublisher | None:
         return self._publisher
 
     @publisher.setter
-    def publisher(self, value: RTPublisher):
+    def publisher(self, value: RTPublisher | None) -> None:
```
soulFood5632: You should not be able set the Publisher with None. We should keep it as taking in just a publisher. Please avoid changing the functionality and just focus on typing, thanks :)

soulFood5632 (changes requested):
One last comment from your previous changes.

packages/railtracks/src/railtracks/context/internal.py
ColumbusLabs: Updated in 5b3cd47 after rebasing onto current main: publisher remains typed as RTPublisher, not optional, and the unnecessary casts are removed. Focused context tests, mypy, ruff, and the full suite pass.

soulFood5632 (approved):
LGTM

Issue #1264, opened 2026-07-17
Railtracks Event + Publishing Parent


soulFood5632:
### Description

A parent ticket to house the Event creation and publishing points within Railtracks. The scope of each of these children tickets will be to create both the `Event` types and the locations in which they are sent. Each child issue will refer to a general class of event types. Note each of these general classes should have different types of events (start, end, failure etc.) 

Please ensure that you are working with @Amir-R25 to ensure that we are introducing information loss as we move to a more event driven storage model so there are fields we can drop as compared to the previous functionality (see NodeDetails for an example of what I mean). The overall design of the system is below:

Lots of pseudocode design here. should be ready for implementation soon

## Key Comments

- Where we should do the id injection is an open question to be answered. Here are our approach:
    - expose function to create each type of `SessionEventBase` with custom injected parent logic
        - Pros:
            - imports of context vars stays very contained within the “bridge file”
            - Better testability and separation
        - Cons:
            - Parent logic is far away from the types
            - Lots of extra boilerplate functions
    - Note no matter what we do this id collection will be our first line of defence for bas state. it should error out so we catch bugs early

## General Base

```python
ParentT = TypeVar(bound=Parent)

class SessionEventBase(ABC, BaseModel, Generic[ParentT]):
	timestamp: datetime = default_factory(lambda: time.time)
	parent: ParentT | _UNSET = _UNSET
	
	@abstractmethod
	@classmethod
	def event_type(cls) -> str:
		pass
```

## Helper Classes

```python
# we will need to add a type descrimator here
class Parent(BaseModel):
	pass
	
class MiddlewareParent(Parent):
	middleware_id: str
 
class NodeParent(Parent):
  node_id: str
  middleware_id: str | None 
  
class LLMParent(Parent):
  llm_call_id: str
  middleware_id: str | None
  
class NoParent(Parent):
  pass
  
  
```

## Node Invocations

```python
class NodeEventBase(SessionEventBase[NodeOnlyParent | NoParent | MiddlewareOnlyParent]):
  name: str
  node_id: str
  node_type: str

# initial instantation
class NodeCreation(NodeEventBase):
	pass
	
# entering the node body
class NodeInvocation(NodeEventBase):
  args: tuple[Any, ...]
  kwargs: dict[str, Any]
  
  
# failures inside a node body
class NodeFailure(NodeEventBase):
	failure: Exception

# response to the node itself
class NodeResponse(NodeEventBase):
  response: Any
  
[...]

Issue #1138, opened 2026-06-01
[Feature] Migrate General Node Observability Details (e.g. latency) to new Node Observability Store


soulFood5632:
### Feature Request

Note you are blocked until the Parent types are implemented: see #1294  for more details

The task is to implement: 
- The below types (see `events.node.py`)
- Implement a function to access the correct parent ids for node event types(see `events.send.py`)
- Publish the events from the correct locations (find good locations to send each of the following events)

```python
class NodeEventBase(SessionEventBase[NodeOnlyParent | NoParent | MiddlewareOnlyParent]):
  name: str
  node_id: str
  node_type: str

# initial instantation
class NodeCreation(NodeEventBase):
	pass
	
# entering the node body
class NodeInvocation(NodeEventBase):
  args: tuple[Any, ...]
  kwargs: dict[str, Any]
  
  
# failures inside a node body
class NodeFailure(NodeEventBase):
	failure: Exception

# response to the node itself
class NodeResponse(NodeEventBase):
  response: Any
  
# final response outside of middleware
class NodeDestruction(NodeEventBase):
  response: Any
```

PR #1302, opened 2026-07-27
pooria/llm scope

Pooria90:
## Summary

Move the LLM-call identifier off the ad-hoc flat `SessionContext.llm_call_id`
field and onto the scope chain as a new `ScopeKind.LLM` entry, so all four scope
kinds (node, node-body, middleware, llm) are tracked by the **one** immutable
`ScopeLink` mechanism.

`enter_llm_call()` now pushes/pops a `ScopeEntry` exactly
like `enter_middleware()`, and `get_llm_call_id()` resolves it with the same
chain walk as the other id accessors. 

`get_llm_call_id()` keeps its signature. This unblocks the upcoming event parent-resolution work: a single chain walk can
now resolve every parent kind uniformly instead of special-casing LLM.

Part of the observability event work (#1138, design #1264).

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

- **Behavior nuance (intended):** because a node's tool calls are dispatched
  *after* the LLM call returns, the LLM scope entry is already popped at
  child-dispatch time, so children still do not inherit an `llm_call_id`. This
  invariant is now covered by `test_captured_scope_after_llm_call_carries_no_llm_id`
  (replaces the old `test_restore_scope_does_not_carry_llm_call_id`, whose
  "flat field is always reset" premise no longer applies).
- The base branch has **pre-existing, unrelated** red tests and one file that
  fails collection (`built_nodes/function/test_function.py`), so a full
  `pytest tests` / `ruff check .` is not clean repo-wide independent of this PR.
  This PR introduces no new failures (verified: the failure set for
  execution + middlewares is identical with and without these changes).

### DIFF
--- a/packages/railtracks/src/railtracks/context/central.py
+++ b/packages/railtracks/src/railtracks/context/central.py
@@ -140,7 +140,11 @@ def get_llm_call_id() -> str | None:
         ContextError: If the global variables have not been registered.
     """
     context = safe_get_runner_context()
-    return context.session_context.llm_call_id
+    return (
+        context.session_context.current_llm_call_id.id
+        if context.session_context.current_llm_call_id is not None
+        else None
+    )
 
 
 def get_middleware_id() -> str | None:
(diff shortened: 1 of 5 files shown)


soulFood5632 (approved):
LGTM

PR #1303, opened 2026-07-27
LLM Events

soulFood5632:
## Summary
This PR contains the following changes:

1. Refactoring of `_resolver.py` logic. For simple tree traversals I deemed that less abstraction was easier to follow. @Pooria90 to review.
2. Implementation of helpers to access the llm and middleware ids and attaching them to logs accordingly
3. Reimplementation of the `ScopeEntry` object that allows an optional param for the `type id` in addition to the `invoke id`
4. Optimized object structure of Event types. (Base -> Parent, CreationEvents -> concrete types or other smaller base classes) [this includes some new abstract method structure]
5. Placed serialization requirements on the observer side of the contract instead of in the `pipe` method 
6. Migrated the node observability into a new node observability injected when any node is created via `node_builder`

## Notes
- These changes do not properly handle the full use cases of when nodes are called within middleware, see #1304 
- Tests were not constructured until the schema of all the events is decided
- Some print statements remain in the body of this PR. This will stay until the second stack.

PR #1304, opened 2026-07-28
1144 middleware observability

### LINKED ISSUE #1144: [Feature] Implement Obseravbility of Add Ons

soulFood5632:
## Summary

This PR builds on the work in #1303. It completes the following additions:

1. Adds emission sites inside the user created middleware. (all guards, after_llm, before_llm etc.)
2. Middleware Events
3. Implemented logic for resolving spatial parents for nodes including middleware
4. Implemented logic for resolving spatial parents for all middleware events

## Notes
I don't love ad-hoc the decision on where to place emission sites was. Looking through each decision is ok but there isn't a consistent theme which is bothersome tbh

### DIFF
--- a/packages/railtracks/src/railtracks/built_nodes/llm/middleware/wrap_llm.py
+++ b/packages/railtracks/src/railtracks/built_nodes/llm/middleware/wrap_llm.py
@@ -2,6 +2,12 @@ from typing import Awaitable, Callable
 
 from pydantic import BaseModel
 
+from railtracks.events.middleware import (
+    MiddlewareModelFailureEvent,
+    MiddlewareModelInvocationEvent,
+    MiddlewareModelResponseEvent,
+)
+from railtracks.events.send import pipe
 from railtracks.llm.history import MessageHistory
 from railtracks.llm.response import Response
 from railtracks.llm.tools.tool import Tool
@@ -29,6 +35,36 @@ def wrap_llm(
         return response
     ```
     """
-    wrapped = wrap_node(fn)
 
-    return wrapped
+    @wrap_node
+    async def _pipe_wrapped(
+        llm_call: LLM_CALL,
+        message_history: MessageHistory,
+        schema: type[BaseModel] | None,
+        tools: list[Tool] | None,
+    ):
+        input_event = MiddlewareModelInvocationEvent(
+            message_history=message_history,
+            schema=schema,
+            tools=tools,
+        )
+        await pipe(input_event)
+
+        try:
+            response = await fn(llm_call, message_history, schema, tools)
+        except Exception as e:
+            event = MiddlewareModelFailureEvent(
+                exception=e,
+            )
+            await pipe(event)
+            raise e
+
+        output_event = MiddlewareModelResponseEvent(
+            response=response,
+        )
+
+        await pipe(output_event)
+
+        return response
+
+    return _pipe_wrapped
--- a/packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
+++ b/packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
@@ -121,7 +121,7 @@ class ModelInvoker:
 
         await pipe(
             LLMCreationEvent(
-                model_id=model.id,
+                llm_model_id=model.id,
                 model_name=model.model_name(),
                 model_provider=model.model_provider(),
             )
--- a/packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
+++ b/packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
@@ -120,7 +120,7 @@ class ModelInvoker:
 
         await emit(
             LLMCreationEvent(
-                llm_model_id=model.id,
+                llm_id=model.id,
                 model_name=model.model_name(),
                 model_provider=model.model_provider(),
             )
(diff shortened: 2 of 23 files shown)


packages/railtracks/src/railtracks/built_nodes/llm/middleware/wrap_llm.py
```diff
@@ -2,6 +2,12 @@
```
Amir-R25, line 2: just looking at this code, if I understood it correctly, it's a combination of an `before_llm` + `after_llm`. Out of curiosity, is there a way that this logic can just reuse those two in sequence rather than introducing new code?

packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
```diff
@@ -120,7 +120,7 @@ async def invoke(
 
         await emit(
             LLMCreationEvent(
-                model_id=model.id,
+                llm_model_id=model.id,
```
Amir-R25: MRI Imaging Centre

packages/railtracks/src/railtracks/events/_resolve.py
```diff
@@ -18,6 +22,28 @@ def node_spatial_parent(scope: ScopeLink[ScopeEntry] | None):
...
+    # first we need to traverse up to find the parent
+    node_link = scope.find_link(lambda e: e.kind is ScopeKind.NODE)
+
+    assert node_link is not None, (
+        "Expected a node scope entry for node spatial parent resolution"
+    )
+
+    parent = node_link.parent
+
+    # there are 3 cases here
```
Amir-R25: comments 😱

packages/railtracks/src/railtracks/events/llm.py
```diff
@@ -28,7 +28,7 @@ def _get_parent(self, scope: ScopeLink[ScopeEntry] | None):
 
 @dataclass(kw_only=True)
 class LLMCreationEvent(CreationEventBase):
-    model_id: str
+    llm_model_id: str
```
Amir-R25: ☹️

packages/railtracks/src/railtracks/middleware/core.py
```diff
@@ -27,6 +32,9 @@ def _require_async(fn: Callable, role: str) -> None:
         )
 
 
+logger = get_rt_logger(__name__)
```
Amir-R25, line 35: putting a comment if we want to rethink logging at some point

Amir-R25 (approved):
Great work, put a comment on some logic that looked repeated to me, not a big deal

packages/railtracks/src/railtracks/built_nodes/llm/middleware/wrap_llm.py
soulFood5632: That is a decent idea except there is isn't a super easy way to connect the pair. I'm gonna keep it in this PR but your comment is reasonable.

packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
soulFood5632: on it

packages/railtracks/src/railtracks/events/llm.py
soulFood5632: fixed :(

packages/railtracks/src/railtracks/middleware/core.py
soulFood5632: agreed

packages/railtracks/src/railtracks/middleware/core.py
```diff
@@ -53,6 +61,22 @@ def __init__(
...
+    async def start_creation_task(self):
+        if self._has_registered:
+            return
+
+        self._has_registered = True
+        event = MiddlewareCreationEvent(
+            middleware_type_id=self.type_id,
+            middleware_name=self.name,
+        )
+        return await pipe(event)
```
Pooria90, line 79: This should be `emit`

packages/railtracks/src/railtracks/middleware/chain.py
Pooria90, line 1: `pipe`s should be replaced with `emit`

PR #1309, opened 2026-07-28
pooria/parent event design

Pooria90:
# Node observability events + parent resolution

> Draft PR note, not the final PR. Implements [#1138](https://github.com/RailtownAI/railtracks/issues/1138)
> under parent design [#1264](https://github.com/RailtownAI/railtracks/issues/1264).
>

## What this adds

Node lifecycle events now flow end-to-end: every node emits `creation → invocation →
response → destruction` (or `→ failure`) as it runs, each tagged with its **self-id** (which
invocation it's about) and its **spatial parent** (who it's nested inside).

Core idea: **the scope chain is already the source of truth for nesting**, so we don't add new
context state — we (a) read that one chain with a resolver, and (b) hook it into the existing
`Observer`/`Event` pipeline.

Self vs. parent is the whole subtlety: it falls straight out of "self is your own entry; parent is the entry(ies) below it".

## Design choices (short)

1. **Self-id is passed by the emitter, not injected from context.** Every emission site holds
   the node object, so `name`/`node_id`/`node_type` are passed directly (`Node._event_identity()`);
   the bridge resolves *only* `parent`. This collapses what was going to be a "NodeCreation is
   the explicit-self exception" into simply *which parent resolver runs*.
2. **A node's parent is always `NodeParent`** (+ `NoParent` at the root), never a bare
   `MiddlewareParent`. A node reached from inside a middleware collapses to
   `NodeParent(node_id, middleware_id=<that mw>)`.
3. **`middleware_id` is single-hop** — the *first* middleware below self. Deeper nesting is
   reconstructed by chaining each event's own parent.
4. **LLM is transparent to node parents, by construction.** If an LLM entry ever appears in a
   node's ancestry (a model-middleware dispatching `rt.call`), the resolver skips it — a node
   is never given an `LLMParent`. LLM is an anchor *only* for model-middleware events.
5. **Observability never crashes a node.** All emits go through one defensive `emit()` that
   swallows + debug-logs any failure (no session/observer, resolve error, publish error).
6. **Payload serialization is a placeholder** (`_json_safe`: datetime → ISO, non-native →
   `repr`). The real field-drop / info-loss pass is a follow-up (with Amir, per #1264).

## Resolve logic

One algorithm, parameterized per event family by `(self_kind, anchor_kinds)`
(`events/_resolve.py`):

1. Locate **self** on the chain (topmost entry of `self_kind`) and start strictly **below** it.
   Missing self = bad emission site → raises (first line of defense).
2. Walk down: capture the **first MIDDLEWARE** as the intervening `middleware_id`; stop at the
   **first anchor** (`NODE`/`NODE_BODY` for node/mw/llm events; `LLM` for model-mw events) and
   build the `Parent`. Everything else — including LLM when it isn't an anchor — is transparent.
3. No anchor before the chain ends → `NoParent` (root).

`NodeCreation` is the one variant: self isn't on the chain yet, so it walks from the current
top with **no self-skip** (`resolve_parent_at_creation`). Same walk otherwise.

Family resolvers: `node_parent`, `node_creation_parent`, `llm_parent`, `middleware_parent`,
`model_middleware_parent`. Each event class picks its resolver via a `resolve_parent(scope)`
hook on `SessionEventBase` — dispatch is on the event, not a type-switch in the bridge.

## Emission sites

All five go through `events/send.py::emit`; identity via `Node._event_identity()`.

| Event | Site | Placement |
|-------|------|-----------|
| `NodeCreation` | `state.py::call_nodes` | after `_create_node_and_request`, under the caller's **restored** scope |
| `NodeInvocation` | `nodes.py::wrapped_invoke` → `body()` | inside `enter_node_body`, args/kwargs in hand |
| `NodeResponse` | `nodes.py::wrapped_invoke` → `body()` | right after `invoke()` (inside middleware) |
| `NodeFailure` | `execution_strategy.py::execute` | inner `try` around `task.invoke()`, inside `enter_node`, then re-raise |
| `NodeDestruction` | `execution_strategy.py::execute` | after `task.invoke()` returns (middleware unwound), inside `enter_node` |

Two placement subtleties worth a look:
- **`NodeCreation`** is emitted in async `call_nodes` (which has the instance — I changed
  `_create_node_and_request` to return `(request_id, instance)`), not in the sync creation
  helper. It resolves correctly because `handle` already wraps `call_nodes` in
  `restore_scope(item.current_scope, …)`, so the caller's chain is live and the not-yet-on-chain
  node walks from the top → `NodeParent(caller)` / `NoParent`.
- **`NodeFailure`/`NodeDestruction`** live in `execute` *inside* the `enter_node` block, because
  the outer `except` runs after that block has popped the NODE entry (which `node_parent`
  needs). Putting `NodeFailure` here also catches middleware failures, not just body failures.


packages/railtracks/src/railtracks/events/_resolve.py
```diff
@@ -0,0 +1,126 @@
+"""Parent resolution: project the scope chain onto an event's spatial `Parent`.
+
+The scope chain (`ScopeLink[ScopeEntry]`) is the single source of truth for "who is
+nested inside whom". This module reads that chain and produces the event-space `Parent`
+for an event:
+
+- **self vs. parent** — an event's parent is the enclosing scope entry(ies) *below* its own
+  entry. The runtime resolvers locate self's entry and walk strictly below it.
```
soulFood5632: do me a favour and just kill these long lines of text. They are such a pain to mantain.

packages/railtracks/src/railtracks/events/send.py
```diff
@@ -1,43 +1,67 @@
-from dataclasses import asdict
+"""Emission entry point: resolve an event's parent from the ambient scope, then publish it.
 
+The full resolved `Parent` is carried in the event payload (with a `type` discriminator);
```
soulFood5632: same here

packages/railtracks/src/railtracks/events/send.py
```diff
@@ -1,43 +1,67 @@
...
-# this should modify the session event base object in place
-def _resolve_parent(event: SessionEventBase):
+    Observability must never break execution: if no session/observer is active, or
+    resolution/publishing raises, we log at debug and carry on.
     """
-    Resolves the parent of the event to a string representation.
+    try:
+        await pipe(event)
+    except Exception:  # noqa: BLE001 - observability must not crash a node
+        logger.debug(
```
soulFood5632: lets do logger.exception. These should never be happening so it is ok to have it att that level

packages/railtracks/src/railtracks/events/send.py
```diff
@@ -1,43 +1,67 @@
...
-            f"Event {event} has a parent set, but this is not supported in the current implementation."
-        )
-    # TODO: finish this impelmentation here
-    event.parent = _get_node_parent(event)
+async def pipe(event: SessionEventBase) -> None:
+    """Resolve the event's parent from the current scope chain, then publish it."""
+    if event.parent is UNSET:
+        event.parent = event.resolve_parent(get_current_scope())
 
+    payload = _json_safe(asdict(event))
```
soulFood5632: I think lets defer json safety to the writers. Lets just pass the objects for now.

packages/railtracks/src/railtracks/events/send.py
```diff
@@ -1,43 +1,67 @@
...
+    fields = asdict(parent) if is_dataclass(parent) else {}
+    return {"type": type(parent).__name__, **fields}
 
-def _get_node_parent(event: SessionEventBase) -> Parent:
-    """
-    Resolves the parent of the event to a string representation.
 
-    Args:
-        event (SessionEventBase): The event whose parent is to be resolved.
+def _json_safe(value: Any) -> Any:
```
soulFood5632: see above comment

soulFood5632 (changes requested):
LGTM, super minor comments around docs and json serialization

PR #1310, opened 2026-07-28
Add common LLM hyperparameters (temperature, top_p, reasoning_effort, etc.)

### LINKED ISSUE #1276: Support for Common Parameters like `thinking_level`, `service_tier`, `temperature` etc.
### Description

Right now we do not support any customization of a LLM. We should introduce new parameters in our llm classes to support different modes:

- temperature
- thinking_level
- service_tier

Note that some models will not support these parameters. Make sure that providers that don't support these flags are clearly indicated by the option not appearing in their inits.

By the end of this issue I should be able to do the following:
```python
rt.llm.OpenAILLM("gpt-4o", service_tier="FAST", thinking_level="HIGH")
```

Aryan-Railtown:
## Summary

Adds common LLM hyperparameters (`temperature`, `top_p`, `max_tokens`, `frequency_penalty`, `presence_penalty`, `reasoning_effort`, `service_tier`, `verbosity`) to `rt.llm.*` model wrappers. Per-model support is checked at construction time, so unsupported or invalid hyperparameter combinations raise immediately instead of silently dropping or failing deep into a run.

Closes #1276.

## Type of change

- [x] Feature
- [x] Docs

## Design decisions

- Fail fast at construction (`UnsupportedHyperparameterError` / `MutuallyExclusiveHyperparametersError`) — never silently drop-and-warn.
- litellm's `get_supported_openai_params` is the default source of truth for per-model support, patched with a small manual denylist for a few confirmed litellm staleness gaps (see caveats).
- No client-side validation of hyperparameter *values* (ranges/types/enums) — provider-native errors are already clear and per-model-accurate.

## Caveats / known gotchas

- Anthropic Opus 4.7/4.8 reject `temperature`/`top_p` despite litellm reporting them as supported ([litellm#26444](https://github.com/BerriAI/litellm/issues/26444), still open upstream).
- Anthropic rejects `temperature`+`top_p` **together** on every model tested, not just Opus 4.7+ — a separate `MutuallyExclusiveHyperparametersError`.
- `gpt-5-codex`/`gpt-5.1-codex` don't support `verbosity` despite litellm's schema listing it.
- Gemini (`gemini-2.5-*`) rejects `frequency_penalty`/`presence_penalty` despite litellm reporting them as supported.
- OpenAI silently accepts an invalid `verbosity` value with no error — documented, not coded around.

## Checklist

- [x] Docs updated
- [ ] Breaking changes include migration notes — n/a, additive only


docs/documentation/agent_design/llms/common_hyperparams.md
```diff
@@ -0,0 +1,66 @@
...
+this **when you construct the model** — before making any network call — and raises
+immediately if a hyperparameter (or combination of hyperparameters) isn't supported:
+
+```python
+--8<-- "docs/scripts/documentation/common_hyperparams.py:fail_fast"
+```
+
+!!! tip "Railtracks Recommendation"
+    Treat `UnsupportedHyperparameterError` / `MutuallyExclusiveHyperparametersError` at
+    construction as an actionable signal, not a bug to work around — it's telling you
```
soulFood5632: kill the em dash for me :)

docs/documentation/agent_design/llms/common_hyperparams.md
```diff
@@ -0,0 +1,66 @@
...
+## Things to know
+
+- **Anthropic**: Opus 4.7 and later reject non-default `temperature`/`top_p`. Separately,
+  specifying `temperature` **and** `top_p` together is rejected on Anthropic models in
+  general — pass at most one of the two.
+- **OpenAI**: `verbosity` is only supported on the GPT-5 series, and not on the Codex
+  variants (`gpt-5-codex`, `gpt-5.1-codex`, etc.).
+- **Gemini**: `frequency_penalty` and `presence_penalty` are not currently supported.
+
+Railtracks keeps this list current as providers change their behavior — if a
```
soulFood5632: ditto

docs/scripts/documentation/common_hyperparams.py
```diff
@@ -0,0 +1,30 @@
...
+
+# --8<-- [start: reasoning_effort]
+import railtracks as rt
+
+reasoning_llm = rt.llm.OpenAILLM("gpt-5-mini", reasoning_effort=rt.llm.ReasoningEffort.LOW)
+# --8<-- [end: reasoning_effort]
+
+# --8<-- [start: fail_fast]
+import railtracks as rt
+from railtracks.llm.models._model_exception_base import UnsupportedHyperparameterError
```
soulFood5632: We shouldn't be showing on a non PEP recommended import path. We should either expose using __init__'s in a better way or just not import it

packages/railtracks/src/railtracks/llm/providers.py
```diff
@@ -33,3 +33,16 @@ class ModelProvider(str, Enum):
     TELUS = "Telus"
     PORTKEY = "PortKey"
     UNKNOWN = "Unknown"
+
+
+class ReasoningEffort(str, Enum):
```
soulFood5632: Let's slot this reasoning effort enum in a better location. I think there are 2 good options:
- change the name of this file to `config` or something like that. 
- create a new file for for this reasoning effort named whatever you think would be good. (Overall it also may make sense to move both of these classes closer to the model creation module, I dunno your thoughts?)

packages/railtracks/src/railtracks/llm/models/api_providers/_provider_wrapper.py
```diff
@@ -23,15 +23,77 @@ def __init__(
         api_base: str | None = None,
         api_key: str | None = None,
         temperature: float | None = None,
+        top_p: float | None = None,
+        max_tokens: int | None = None,
+        frequency_penalty: float | None = None,
+        presence_penalty: float | None = None,
+        reasoning_effort: ReasoningEffort | str | None = None,
```
soulFood5632: I think we would be better off with a straight Literal type here, that way users don't have to import the enum class. It also might be good to remove the ability to add a general string. That way it is more fail fast from the user.

packages/railtracks/src/railtracks/llm/models/api_providers/_provider_wrapper.py
```diff
@@ -23,15 +23,77 @@ def __init__(
...
+        service_tier: str | None = None,
+        verbosity: Literal["low", "medium", "high"] | str | None = None,
```
soulFood5632: same here, lets just do the literal

packages/railtracks/src/railtracks/llm/models/cloud/azureai.py
```diff
@@ -41,7 +41,9 @@ def __init__(
             model_name (str): Name of the Azure AI model to use.
             temperature (float | None, optional): Sampling temperature for generation (e.g. 0.0–2.0).
                 If None, the provider default is used.
-            **kwargs: Additional arguments passed to the parent LiteLLMWrapper.
+            **kwargs: Additional arguments passed to the parent LiteLLMWrapper, including the
```
soulFood5632: for consistency this should be same signature as portkey?

packages/railtracks/src/railtracks/llm/models/_hyperparameter_support.py
```diff
@@ -0,0 +1,86 @@
+"""
+Per-model, per-hyperparameter support checks for the "common hyperparameters"
```
soulFood5632: I've been vehemently against these top of file things since they are such a pain to maintain. But I am understanding why they are valuable if an AI is reading it. For now can you please remove it. I will bring it up as a team discussion to determine on what our approach should be.

soulFood5632 (changes requested):
Looks good to me. No major issues, just a couple of stylistic and interface comments

packages/railtracks/src/railtracks/llm/providers.py
Aryan-Railtown: That is a good point, but i think we can just make the reasoning effort into a literal and completely remove this enum :)

packages/railtracks/src/railtracks/llm/models/cloud/azureai.py
Aryan-Railtown: yes good point

soulFood5632 (approved):
LGTM

PR #1312, opened 2026-07-29
Stop pre-rejecting PDF attachments on models litellm can't identify

Amir-R25:
Gate the `supports_pdf_input` check on `litellm.model_cost` membership, models with no catalog entry (custom Azure Foundry deployments etc.) now skip the pre-check and let the API decide, while known text-only models still raise upfront.


soulFood5632 (approved):
LGTM, as long as the comment was your own doing.

PR #1322, opened 2026-07-30
Fix kwargs not funneling through to litellm completion calls

Aryan-Railtown:
## Summary

`**kwargs` had no working path down to the actual `litellm.completion`/`acompletion` call at construction time. This fixes that, scoped to construction time only.
See the discussion below for why call-time support was deliberately left out.

No issue link,  found this out thanks to @CoronRing 

## What was broken

`AzureAILLM.__init__` and `OllamaLLM.__init__` already ended with `**kwargs` and forwarded it to `super().__init__(..., **kwargs)` — but `LiteLLMWrapper.__init__` (what they forward to) had no `**kwargs` parameter at all, so any extra construction-time kwarg would raise `TypeError`.

## What changed

LiteLLMWrapper.__init__ now accepts **kwargs, stores them, and merges them into every completion call 
- so litellm params we haven't named explicitly (e.g. seed) get forwarded automatically. 

## Why not call-time kwargs too

Earlier version also added **kwargs to chat()/chat_with_tools() etc. Logan pointed out nothing calls these with per-call overrides today, and litellm.drop_params=True means a typo'd kwarg silently rides along with zero warning instead of erroring. Reverted that part; kept construction-time only, which has a concrete use case.

## Verification
- all tests pass

### DIFF
--- a/packages/railtracks/src/railtracks/llm/model.py
+++ b/packages/railtracks/src/railtracks/llm/model.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from typing import (
-    Any,
     AsyncGenerator,
     Callable,
     Generator,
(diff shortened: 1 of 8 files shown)


soulFood5632 (approved):
LGTM

PR #1325, opened 2026-07-31
1206 V4 streaming (queue): per-call astream queue, remove bus streaming

CoronRing:
## Summary

Refactor streaming to a single, plain mechanism. `rt.astream(agent, ...)` streams an agent's LLM tokens through a **per-call `asyncio.Queue`** carried frame-locally in context, instead of routing chunks over the pub/sub bus. This removes the channel / callback / `stream_id` machinery that existed only to route chunks back to the right consumer across the task boundary, a per-call queue answers that by construction.

The user-facing `Stream` shape is unchanged:

```python
@rt.function_node
async def head(chunk: str, logger) -> str:
    stream = rt.astream(node1, chunk)      # node1 is an agent_node
    async for c in stream:
        logger.pipe(c)
    return stream.result.text

final = await rt.call(head, "Echo [abc]", logger)
```

Concurrent streams are just separate handles, no more channels, isolation is structural (each has its own queue):

```python
a = rt.astream(node1, "Echo [123]")
b = rt.astream(node1, chunk)
a_res, b_res = await asyncio.gather(drain(a), drain(b))
```

## What changed

**New mechanism.** The `Stream` creates a queue and passes it on `RequestCreation(stream_queue=...)`; it rides the existing plumbing (`state → call_nodes → _run_request → Task → update_parent_id`) into the agent frame's `InternalContext` (frame-local, never inherited). `ModelInvoker` drains `model.astream_*` directly onto that queue. The bus still delivers run completion; the `Stream` enqueues a `("done", ...)` marker on `RequestFinished` so the consumer knows when to resolve `.result`.

**Agent-only.** `rt.astream` now requires an agent node (`Node.type() == "Agent"`); a `@function_node`/tool node raises `NodeCreationError`. Because streaming never propagates through nested frames, `stream_id` inheritance and channels are unnecessary. One can return a generator directly.

**Removed** (all greps to zero across `src` + `tests`): `broadcast_stream`, `stream_callback` (Session / Flow / `set_config` / `ExecutorConfig`), `StreamingKind`, `Streaming.kind`/`.channel`/`.stream_id`, `get_stream_id`, `is_streaming_enabled`, `stream_id` inheritance, `RequestCreation.stream`/`.current_stream_id`.

**Kept.** The `rt.broadcast` + `broadcast_callback` (`Streaming` simplified to `{streamed_object, node_id}`; `stream_subscriber` → `event_subscriber`). Also untouched: the model-layer `astream_chat*` methods (the token source), `Response.text`, `TOOL_CALLING_STREAMING_BLACKLIST`, and `couple` (middleware).

**Docstring** Changes I own the framework from last PR

This is a clean inverse of #1262 (which wired astream onto the bus)  every symbol that PR introduced is gone, while the #1257 model-level plumbing it built on remains intact.

## Notes

- `rt.astream` remains the only public streaming entry point; `Stream` is the return type only.
- `stream_queue` carries only token *data*; results and errors still flow through the bus completion path.

### DIFF
--- a/packages/railtracks/src/railtracks/interaction/_astream.py
+++ b/packages/railtracks/src/railtracks/interaction/_astream.py
@@ -93,7 +93,7 @@ class Stream(Generic[_TOutput], AsyncIterator[Any]):
         self._args = args
         self._kwargs = kwargs
 
-        # the request id doubles as the stream scope id used to tag every chunk of this run
+        # identifies this run so the completion subscriber can match its terminal message
         self._request_id = str(uuid4())
 
         self._queue: asyncio.Queue[tuple[str, Any]] | None = None
@@ -113,8 +113,8 @@ class Stream(Generic[_TOutput], AsyncIterator[Any]):
     # ------------------------------------------------------------------ lifecycle
 
     async def _start(self) -> None:
-        """Lazily starts the run on first iteration: subscribes to the bus, then publishes
-        the entry request with streaming enabled."""
+        """Lazily starts the run on first iteration: subscribes for the run's completion,
+        then publishes the entry request with its chunk queue attached."""
         if self._started:
             return
         self._started = True
--- a/packages/railtracks/src/railtracks/interaction/_astream.py
+++ b/packages/railtracks/src/railtracks/interaction/_astream.py
@@ -53,33 +53,20 @@ class Stream(Generic[_TOutput], AsyncIterator[Any]):
     """
     The handle returned by `rt.astream(...)`.
 
-    A `Stream` is an async iterator over the chunks emitted during a streamed invocation. It
-    yields *only* chunks (typically `str` tokens); the node's final return value is exposed
-    separately so there is never any ambiguity between a chunk and the final result:
+    Async-iterate a `Stream` to receive the chunks (typically `str` tokens) emitted during
+    the run, then read the node's final return value from `.result`. The final result may
+    differ from the concatenation of the chunks (e.g. when output guardrails correct the
+    buffered response).
 
-    ```python
-    stream = rt.astream(agent, user_input="Write a poem.")
-    async for chunk in stream:
-        print(chunk, end="", flush=True)
-    final = stream.result  # the node's complete return value (e.g. StringResponse)
-    ```
+    A `Stream` is also awaitable: `await stream` runs it to completion (discarding unread
+    chunks) and returns the final result, which is useful when you break out of iteration
+    early but still want the result. The run always finishes; breaking does not cancel it.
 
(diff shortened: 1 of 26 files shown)


docs/system_internals/concepts.md
soulFood5632, line 1: We discussed previously these docs have been deleted and are out of scope. Please for this PR keep changes to system_internals out of the PR

docs/system_internals/session.md
soulFood5632, line 1: ditto

packages/railtracks/src/railtracks/interaction/_astream.py
soulFood5632: shouldn't we do the check up here to verify if its of type node, instead of unpacking it first

soulFood5632: same here why are we allowing an input of RTFunction node

packages/railtracks/src/railtracks/interaction/_astream.py
```diff
@@ -327,6 +297,19 @@ def astream(
         # not an RTFunction (no `node_type`), so it is already a Node subclass
         node = cast("type[Node[_P, _TOutput]]", node_)
 
+    # rt.astream streams an agent's LLM tokens, so it only accepts agent nodes. Anything else
+    # (a @function_node / tool node) has no token stream to surface use rt.call instead.
+    node_kind = node.type() if hasattr(node, "type") else None
```
soulFood5632: with proper typing you should be able to avoid this `hasattr` headache

soulFood5632 (changes requested):
LGTM, some super minor comments around type verification and docs

docs/system_internals/concepts.md
CoronRing: I thought that was advanced and system_internal is suppose to replace that anew, maybe I confused myself. Will move notion

packages/railtracks/src/railtracks/interaction/_astream.py
CoronRing: I think getting node before` .type() == "Agent"` is a cleaner solution. Otherwise, it might be chaotic checking try-catch attempting to parse each. I guess what we can do is whether the item passed in is of `  type() == Node` only here.

packages/railtracks/src/railtracks/interaction/_astream.py
CoronRing: Sounds good

packages/railtracks/src/railtracks/interaction/_astream.py
CoronRing: Updated, a leftover item.

PR #1327, opened 2026-07-31
1307 session messages

### LINKED ISSUE #1307: Create SessionStarted and SessionCompleted messages
### Description

Create message similar to the others in the setup. This message should include the major items relevant to the payload.

Pooria90:
## Summary

Adds the `SessionStarted` / `SessionCompleted` event family and emits it from `_call.py::_start`, so every top-level run is bracketed by a `session.started` / `session.completed` pair in the observability stream. `_start` is the only viable emission site today because it also owns the observer's lifetime (`ensure_started()` … `shutdown()`). Anything emitted outside that window is dropped.

To make `session.completed` reliable, the wait-for block is restructured so the completion event and the observer teardown both run in a `finally`; previously `shutdown()` was skipped whenever the top-level node raised, leaving the observer running and its writers undrained. Flow/session identity (`flow_name`, `flow_id`, `session_name`) now travels on `SessionContext` so the emitter can stamp it.

Stacked on #1303.

Closes #1307

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

- `session.started` attributes: `session_id`, `flow_name`, `flow_id`, `session_name`, `entry_point_name`, plus the effective config the run executes under (`timeout`, `end_on_error`, `save_state`).
- `session.completed` attributes: `session_id`, `status` (`"success"` / `"failure"`), `error`, `duration_seconds`.

**Context plumbing:** `SessionContext` gains `flow_name` / `flow_id` / `session_name` (all optional, so existing construction sites keep working), threaded through both `with_scope_pushed` and `restore_scope`, and read at the emit site via the new `get_session_identity()`.

**Known-open, not addressed here:**
1. Only the first top-level `rt.call` per `Session` is observable: `shutdown()` stops the process-wide observer while the publisher keeps running, so a second call takes the `_run` path and every event is swallowed by `emit`.
2. `node.destruction` fires on the failure path (via the `finally` in `nodes.py::wrapped_invoke`); the repaired hot-path test asserts that as current behavior, so changing it will surface there.


soulFood5632 (approved):
LGTM

PR #1328, opened 2026-08-04
1313 name for middleware

soulFood5632:
## Summary

Add name suplimentation to creation of middleware 

```python 
@rt.after_node(name="print_after")
    return result
```

I needed to use overloads to make it so we can support:
```
@rt.after_node 
# and 
@rt.after_node(name)
```

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

### DIFF
--- a/packages/railtracks/src/railtracks/context/internal.py
+++ b/packages/railtracks/src/railtracks/context/internal.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+import asyncio
+from typing import TYPE_CHECKING, Any
 
 from railtracks.utils.config import ExecutorConfig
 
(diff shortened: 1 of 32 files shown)


packages/railtracks/src/railtracks/interaction/_astream.py
```diff
@@ -296,8 +296,8 @@ def astream(
     ```python
     stream = rt.astream(agent, user_input="Write a short poem about rain.")
     async for chunk in stream:
-        print(chunk, end="", flush=True)   # str token chunks
-    final = stream.result                  # the complete StringResponse
+        print(chunk, end="", flush=True)  # str token chunks
```
soulFood5632: inentional changes due to ruff formatting rules.

packages/railtracks/src/railtracks/interaction/broadcast_.py
```diff
@@ -12,7 +12,7 @@ async def broadcast(item: str):
     """
     Broadcasts a one-off **event** to the session bus.
 
-    This triggers the `broadcast_callback` you have already provided. Events are 
+    This triggers the `broadcast_callback` you have already provided. Events are
```
soulFood5632: ditto

packages/railtracks/src/railtracks/context/internal.py
```diff
@@ -92,7 +92,7 @@ def stream_queue(self) -> asyncio.Queue[Any] | None:
 
         When set, the frame is the entry of a streamed invocation (see `rt.astream`): its
         LLM node writes each token chunk directly onto this queue, which the `Stream` handle
-        on the calling side drains. 
+        on the calling side drains.
```
soulFood5632, line 95: ruff format

Amir-R25 (approved):

PR #1332, opened 2026-08-04
Support bound methods and functools.partial as function_node tools (#1318/#1336)

### LINKED ISSUE #1318: Support Bound method and Partial methods as tools.
### Description

# Bound and patitial
Right now, `function_node` only support the conversion of a standard method to `tool_node`, however, bound method and partital method is especially common to support, especially in agentic application.

## Tasks
| Case | What's needed |
| --- | --- |
| **Bound methods** | Just relax the guard. They already carry `__name__`, `__qualname__`, and the real `__doc__`. ~1 line. |
| **functools.partial** | Relax the guard + a small helper to source name/qualname/doc from the callable, unwrapping `.func` when they're missing (a partial's `__doc__` is the useless `"partial(func, ...)"` boilerplate, and it has no `__name__`/`__qualname__` at all). |

### LINKED ISSUE #1336: [Feature Branch] Uncertain function node `name` functionality
### Description

Tool.from_function accepts a name kwarg but the builder never forwards it, so the tool name is always function.__name__.

```

rt_cube = rt.function_node(cube, name="cube", manifest=rt.ToolManifest(
    description="Calculates the cube of a number.",
    parameters=[
        rt.llm.Parameter(name="base", param_type="object", description="The base number to be cubed.")
    ]
))
```

cube gets ignored

CoronRing:
## Summary

Closes #1318, #1336.

`function_node` previously accepted only plain functions and coroutine functions. Passing a **bound method** or a **`functools.partial`** raised `NodeCreationError` (for the sync cases) or blew up later on missing metadata. This PR makes both first-class tool sources, and fixes a related naming bug that surfaced immediately once partials worked.

## What changed

**1. Accept bound methods and `functools.partial` (`built_nodes/function/node.py`)**

- Relaxed the callable guard so `inspect.ismethod` (bound/class methods) and `functools.partial` are accepted alongside plain functions and coroutines. Bound methods already carry correct `__name__`/`__qualname__`/`__doc__`, so they only needed the guard opened up.
- Added `_partial_with_resolved_metadata`: a `functools.partial` exposes no `__name__`/`__qualname__` and only a boilerplate `"partial(func, ...)"` `__doc__`, which the tool-schema extractor reads and would either crash on or turn into garbage. The helper unwraps `.func` (recursively, for nested partials) to recover the real metadata and stamps it onto a **fresh copy**, leaving the caller's object untouched. The copy keeps the partial's **reduced signature**, so already-bound *positional* args are correctly excluded from the inferred tool parameters.
- Extracted the validation + normalization block into `_validate_and_normalize_callable` (partial resolution, dict-parameter validation, manifest cross-check, builtin wrapping, and the callable-type guard), keeping `_single_function_node` under the mccabe complexity limit and isolating the accepted-callable policy in one place.

**2. Forward the `name=` override to the LLM-facing tool name (`built_nodes/function/node_builder.py`)**

`FunctionNodeBuilder.function` set the node id and class name from `name=`, but always derived the **tool** name from `function.__name__` — so `name=` renamed the node without renaming the tool the LLM dispatches on. This left distinct nodes sharing one tool name; most visibly, two partials of the same parent both collapsed to the parent's `__name__` and **collided at tool-dispatch time** (`get_node_from_name`), an error unreachable by any static check. Fixed by passing `name` into `Tool.from_function` so the override reaches the tool name and such nodes can be disambiguated.

## Note
- **Keyword-bound partial args still appear in the tool schema.** `inspect.signature` drops *positionally*-bound args but keeps *keyword*-bound ones (e.g. `partial(power, exp=2)` still lists `exp` as an optional param). Dropping those would mean explicitly filtering names in `partial.keywords`; left for a follow-up so this PR stays focused on making bound methods/partials work and the naming collision go away.
- **No static uniqueness check** across independently-created nodes, tool-name collisions still surface only at dispatch. Broader change touching the LLM node; out of scope here. Ie multiple partials always trigger #1337.

### DIFF
--- a/packages/railtracks/src/railtracks/built_nodes/function/node.py
+++ b/packages/railtracks/src/railtracks/built_nodes/function/node.py
@@ -116,10 +116,8 @@ def _validate_and_normalize_callable(
 ) -> Callable[_P, Coroutine[None, None, _TOutput]] | Callable[_P, _TOutput]:
     """Validate ``func`` and normalise it into a node-ready callable.
 
-    Runs dict-parameter validation, cross-checks an optional ``manifest`` against
-    the signature, resolves ``functools.partial`` metadata, wraps builtins so the
-    node type can be attached, and rejects anything that is not a supported
-    callable.
+    Resolves partial metadata, validates parameters against an optional
+    ``manifest``, wraps builtins, and rejects unsupported callables.
 
     Args:
         func: The callable to validate and normalise.
@@ -131,26 +129,19 @@ def _validate_and_normalize_callable(
     Raises:
         NodeCreationError: If ``func`` is not a supported callable type.
     """
-    # functools.partial objects carry no usable __name__/__qualname__ and only a
-    # boilerplate __doc__, both of which downstream tool-schema extraction relies
-    # on. Normalise into a copy that sources these from the wrapped callable
-    # before any further handling so the rest of the pipeline is oblivious to it.
+    # partials lack real __name__/__qualname__/__doc__; recover them onto a copy
     if isinstance(func, functools.partial):
         func = _partial_with_resolved_metadata(func)
 
-    if not isinstance(
-        func, BuiltinFunctionType
-    ):  # we don't require dict validation for builtin functions, that is handled separately.
-        validate_function(func)  # checks for dict or Dict parameters
+    # dict-parameter validation; builtins are validated separately
+    if not isinstance(func, BuiltinFunctionType):
+        validate_function(func)
 
-    # Validate tool manifest against function signature if manifest is provided
     if manifest is not None:
         validate_tool_manifest_against_function(func, manifest.parameters)
 
     if inspect.isbuiltin(func):
-        # builtin functions are written in C and do not have space for the addition of metadata like our node type.
-        # so instead we wrap them in a function that allows for the addition of the node type.
-        # this logic preserved details like the function name, docstring, and signature, but allows us to add the node type.
+        # builtins are C-level and can't hold our node-type attribute; wrap them
         return _function_preserving_metadata(func)
 
     if not (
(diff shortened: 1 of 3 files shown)


CoronRing:
[tool_call_tutorial.ipynb](https://github.com/user-attachments/files/30722016/tool_call_tutorial.ipynb)

Notebook to illustrate changes.

packages/railtracks/src/railtracks/built_nodes/function/node.py
```diff
@@ -110,6 +110,56 @@ def validate_function_parameters(
         validate_tool_manifest_against_function(func, manifest.parameters)
 
 
+def _validate_and_normalize_callable(
```
soulFood5632, line 128: we might as well type this with generics here or using overloads to make sure we know we aren;t gonna turn a corountine into a regular callable

packages/railtracks/src/railtracks/built_nodes/function/node.py
```diff
@@ -289,3 +320,36 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _TOutput:
...
+
+    Returns:
+        A copy of ``func`` with metadata sourced from the underlying callable.
+    """
+    underlying: Callable = func
+    while isinstance(underlying, functools.partial):
+        underlying = underlying.func
+
+    # discard the partial's boilerplate __doc__ in favour of the underlying one
+    name = getattr(underlying, "__name__", "partial")
```
soulFood5632, line 365: would this be a potential problem where these partial functions would conflict if they both didn't have names and were put into a tool node

soulFood5632 (changes requested):
LGTM, some super minor concerns. 

one question I have is why we are merging into the feature-branch and not directly into `main` for this PR? Nothing in here looks coupled to new logic

CoronRing:
@soulFood5632 No strong reason why its to feature branch, largely because _node_builder file have latest commit in this feature (latest modify in the feature) and we are merging in weeks anyways, so I just PRed here. We can cherry-pick if necessary.

packages/railtracks/src/railtracks/built_nodes/function/node.py
CoronRing: No name is unlikely, as names are from parents. But duplication is the issue mentioned in #1337, as thats the same name, and will be resolved next. 

ie in notebook: tool_nodes=[rt_square, rt_cube] should only work if rt_square, rt_cube are named differently.

packages/railtracks/src/railtracks/built_nodes/function/node.py
CoronRing: ([node.py:205](vscode-webview://0vk3m3lj215mfeeaemhumgn15b5ho6nnbcvtt6cucbapf45m53s0/packages/railtracks/src/railtracks/built_nodes/function/node.py#L205)) func is already declared as the union, so overload resolution collapses back to a union and the casts downstream stay. Not super helpful, but I will overload it for consistency.

soulFood5632: Yea I see that now. I think its still a good idea to overload it for type accuracy anyway

soulFood5632 (approved):
LGTM

CoronRing:
closes #1336

PR #1338, opened 2026-08-04
Add WebSearch prebuilt toolset

### LINKED ISSUE #1234: Web Search Prebuilt Tools
### Description

Build a web search toolset that agents can call to retrieve up-to-date information from the internet.

### Requirements

At minimum we should ship two layers:

- Search: take a query, return a ranked list of results (title, URL, snippet).
- Fetch: take a URL (typically one returned by Search) and return cleaned page content (text extraction, not raw HTML) for the agent to reason over.

For each of those layers we should provide a backend of choice:

- Search backends: e.g. Tavily, Brave Search API, Bing, SerpAPI; one can be default, but the interface should be backend agnostic
- Fetch backends: a simple HTTP + HTML-to-text extractor as default, with room to swap in something heavier later (headless browser) for JS-heavy pages

Other things to scope:

- Rate limiting / result caps (avoid an agent looping and burning API quota)
- Basic domain allow/block list support
- Consistent error handling when a backend is unavailable or a fetch fails (paywalled/blocked pages, etc.)

Aryan-Railtown:
## Summary

Adds a `WebSearchToolSet` prebuilt tool with `search`, `fetch`, and `search_and_fetch`, following the same swappable-backend `ToolSet` pattern as `KeyValueMemoryToolSet`/`ToDoToolSet`. Default backends are `TavilySearch` (search) and `HttpFetch` (httpx + trafilatura, fetch); `BraveSearch` is included as a second search backend, and both `SearchBackend`/`FetchBackend` are protocols so custom backends can be swapped in without touching the toolset.

Closes #1234.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

- [x] Lint & format pass (`ruff check . --no-fix && ruff format --check .`, scoped to files this PR touches under `packages/railtracks/src`; `docs/`, `examples/`, and `tests/` are excluded from ruff by project config)
- [x] Tests added/updated and pass locally (39 new tests; full suite: 2138 passed, 4 skipped)
- [x] Docs updated (new Web Search Tool page under Prebuilt Tools)
- [ ] Breaking changes include migration notes (n/a, no breaking changes)

## Notes

- New optional dependency group `websearch` (`httpx`, `trafilatura`) in `packages/railtracks/pyproject.toml`, added to `all`.
- Per the issue's own "Other things to scope" list: rate limiting, result caps beyond `top_k`, and domain allow/block lists are intentionally deferred to a follow-up. Consistent error handling (backend unavailable, fetch failures) is implemented, backend/network errors are caught and returned as descriptive strings rather than raised.
- A repo-wide `ruff check .` (without `--no-fix`) auto-fixes pre-existing style nits in unrelated files elsewhere in the repo, since this project sets `fix = true` by default. This PR's diff is intentionally scoped to the websearch feature only and does not include those unrelated fixes.
- Also added `examples/tutorials/web_research_agent.py`, a two-agent (Researcher/Summarizer) demo verified live against real Tavily/OpenAI APIs.

### DIFF
--- a/packages/railtracks/pyproject.toml
+++ b/packages/railtracks/pyproject.toml
@@ -54,10 +54,13 @@ visual = [
 # For each Integration package dependencies go here.
 portkey = ["portkey_ai >= 2.0.2"]
 
-# Prebuilt WebSearchToolSet: Tavily search (plain REST via httpx) + trafilatura
-# HTML-to-text extraction for the default fetch backend.
+# Prebuilt WebSearchToolSet: the official tavily-python SDK for the default
+# search backend (Brave, an alternate backend, uses httpx directly since
+# Brave has no official SDK), httpx + trafilatura for the default fetch
+# backend's HTML-to-text extraction.
 websearch = [
     "httpx >= 0.27.0",
+    "tavily-python >= 0.7.0",
     "trafilatura >= 1.8.0",
 ]
 
--- a/packages/railtracks/pyproject.toml
+++ b/packages/railtracks/pyproject.toml
@@ -54,6 +54,13 @@ visual = [
 # For each Integration package dependencies go here.
 portkey = ["portkey_ai >= 2.0.2"]
 
+# Prebuilt WebSearchToolSet: Tavily search (plain REST via httpx) + trafilatura
+# HTML-to-text extraction for the default fetch backend.
+websearch = [
+    "httpx >= 0.27.0",
+    "trafilatura >= 1.8.0",
+]
+
 # Retrieval-stack extras. These are not pulled in by `[all]` — install
 # `railtracks[retrieval]` to opt in.
 tiktok = ["tiktoken>=0.10.0, <=0.12.0"]
@@ -104,7 +111,7 @@ retrieval = [
     "railtracks[stores-all]",
 ]
 
-all = ["railtracks[visual]", "railtracks[portkey]", "railtracks[retrieval]"]
+all = ["railtracks[visual]", "railtracks[portkey]", "railtracks[retrieval]", "railtracks[websearch]"]
 
 [project.scripts]
 railtracks = "railtracks.cli:main"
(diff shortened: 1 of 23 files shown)


packages/railtracks/pyproject.toml
```diff
@@ -54,6 +54,13 @@ visual = [
 # For each Integration package dependencies go here.
 portkey = ["portkey_ai >= 2.0.2"]
 
+# Prebuilt WebSearchToolSet: Tavily search (plain REST via httpx) + trafilatura
+# HTML-to-text extraction for the default fetch backend.
+websearch = [
+    "httpx >= 0.27.0",
```
soulFood5632, line 62: Do we want to provide an additional allowance for peeps to install tavily's lib?

soulFood5632: Comment recinded, see above comment for more disucssion :)

packages/railtracks/src/railtracks/prebuilt/tools/websearch/search/tavily.py
```diff
@@ -0,0 +1,60 @@
...
+        self.api_key = api_key or os.getenv("TAVILY_API_KEY")
+        if not self.api_key:
+            raise ValueError(
+                "Tavily API key not provided. Pass api_key=, or set the "
+                "TAVILY_API_KEY environment variable."
+            )
+        self._base_url = base_url
+
+    async def search(self, query: str, *, top_k: int = 5) -> list[SearchResult]:
+        async with httpx.AsyncClient() as client:
```
soulFood5632: Using a http client directly is probably fine but there is a tradeoff here between using the tavily client library. The worry is if tavily ever changes their API we will need to be able to migrate the changes ourselves instead of relying on their SDK. Just wanted to make sure you thought through that.

soulFood5632 (dismissed):
LGTM, one comment made for consideration.

packages/railtracks/src/railtracks/prebuilt/tools/websearch/search/tavily.py
Aryan-Railtown: i hadn't considered that but i believe the official SDK would be better for long term maintenance, let e make that change

soulFood5632 (approved):
LGTM

PR #1358, opened 2026-08-06
1331 recreate old json trace

### LINKED ISSUE #1331: Recreate the old JSON trace
### Description

We should use the new event models to create the old JSON trace required for the visualizer for the backward compatibility purposes.

Pooria90:
## Summary

The middleware rewrite dropped the per-node `DebugDetails` store, so `state/node.py` now serializes `details: {"internals": None}` into every session document. That block is the only thing consumers read for LLM activity, and both of them break on it: 

1. the visualizer silently renders agents with no model, cost, tokens or messages.
2. `railtracks.evaluations` crashes outright. This restores the block by folding it back out of the event stream, leaving the graph serializer untouched.

Closes #1331.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

### What's added

- **`NodeInternalsCollector`** (`observability/node_internals.py`): buffers the
  ~10 relevant event types per session and folds them into the legacy
  `{guard_details, llm_details, latency}` block. Buffer-then-fold, because guard
  events arrive *before* the LLM event that identifies their node.

- **Inline listeners** (`observability/configure.py`, `publish.py`): a
  synchronous tap that runs on the publishing coroutine, ahead of the `Observer`.
  Used instead of a `Writer` for two reasons: (1) writer queues are bounded with
  `DROP_OLDEST` and would silently hole a long run's `llm_details`, (2) each
  writer gets a consumer `asyncio.Task` that binds to the loop that created it;
  under MCP or threads that breaks with `Task ... attached to a different loop`. (The latter needs further investigation)

- **Model identity + node duration on events** (`events/llm.py`, `events/node.py`): `llm.response` now carries `model_name`/`model_provider` (preferring the
  unprefixed name the provider echoes, as the legacy trace recorded), and
  `node.destruction` carries `duration_seconds`. Neither was reconstructable
  before.

- **Injection** (`_session.py`): `payload()` refills `details.internals` after
  `graph_serialization()`. No signature changes anywhere in `state/`, so the shim
  deletes in one diff once consumers read events directly.

- **Null-safety fix** (`evaluations/point.py`): `.get("internals", {})` never
  applied its default, since the key was present with value `None`.

### Running example

```python
import railtracks as rt
from railtracks.llm import Parameter

_POPULATION = {"tokyo": 37_000_000, "delhi": 34_000_000, "vancouver": 2_700_000}
_COUNTRY = {"tokyo": "Japan", "delhi": "India", "vancouver": "Canada"}

llm = rt.llm.AnthropicLLM("claude-sonnet-4-6")


@rt.function_node
def get_population(city: str) -> int:
    """Look up the metro-area population of a city.

    Args:
        city: Name of the city, e.g. "Tokyo".

    Returns:
        The population as a whole number. Returns 0 for unknown cities.
    """
    return _POPULATION.get(city.lower(), 0)


@rt.function_node
def get_country(city: str) -> str:
    """Look up which country a city is in.

    Args:
        city: Name of the city, e.g. "Tokyo".

    Returns:
        The country name, or "unknown" if the city isn't known.
    """
    return _COUNTRY.get(city.lower(), "unknown")


CityResearcher = rt.agent_node(
    name="City-Researcher",
    tool_nodes=[get_population, get_country],
    llm=llm,
    system_message=(
        "You look up facts about a single city. Use both tools, then answer in one "
        "sentence."
    ),
    manifest=rt.ToolManifest(
        description="Look up the population and country of one city.",
        parameters=[
            Parameter(
                name="city",
                description="The city to research.",
                param_type="string",
            )
        ],
    ),
)

Orchestrator = rt.agent_node(
    name="Orchestrator",
    tool_nodes=[CityResearcher],
    llm=llm,
    system_message=(
        "You answer comparison questions about cities. Delegate each city to the "
        "City-Researcher tool, then compare the results in two sentences."
    ),
)

flow = rt.Flow(name="Subagent-Tool-Calls-Example", entry_point=Orchestrator)

if __name__ == "__main__":
    result = flow.invoke("Which is bigger, Tokyo or Vancouver, and where are they?")
    print(result.content)
```

### For the reviewer

- **`Observer` consumer tasks are not loop-safe.** Reproduces on
  `rt_mcp/test_node_to_mcp.py` and `run/test_exterior_call.py` as soon as any
  writer is registered; latent today only because nothing registers one by
  default. Going inline routes around it; worth a separate issue.
- **Guardrails are unverified end-to-end.**
- **`agent_node(tool_nodes=..., output_schema=...)` is broken on this branch**,
  independently of this work: it skips the structuring pass and returns a
  `StringResponse`.


packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
```diff
@@ -24,39 +24,62 @@
...
+        message_history: MessageHistory,
+        schema: type[BaseModel] | None,
+        tools: list[Tool] | None,
+    ) -> Response:
+        prev_message_history = deepcopy(message_history)
+        model = get_model()
+        model_name = model.model_name()
+        model_provider = model.model_provider()
+
+        invocation_event = LLMInvocationEvent(
```
soulFood5632: The invocation even was intentionally not including the model_name and model_provider. You should be able to get this info from a LLMCreationObject and then referencing the ids

packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
```diff
@@ -24,39 +24,62 @@
...
+            model_provider=model_provider,
+        )
+        await emit(invocation_event)
+        try:
+            response: Response = await call(message_history, schema, tools)
+        except Exception as e:
+            event = LLMFailureEvent.from_exception(
+                e,
+                message_input=prev_message_history,
+                model_name=model_name,
```
soulFood5632: ditto

packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
soulFood5632: The main logic is that all of our events follow this pattern so none of them have to worry about accessing state that may not be relevant to them.

packages/railtracks/src/railtracks/events/llm.py
```diff
@@ -18,7 +18,11 @@
 
 @dataclass(kw_only=True)
 class LLMMessageBase(ParentEventBase[NodeSpatialParent, LLMParent]):
+    """Base for events about a single LLM round trip."""
+
     message_input: MessageHistory
+    model_name: str | None = None
```
soulFood5632: see above

packages/railtracks/src/railtracks/observability/configure.py
```diff
@@ -2,11 +2,17 @@
...
+from typing import Callable
+
+from .models import Event
 from .observer import Observer
 from .writers.base import Writer
 
 observer: Observer = Observer()
 
+# Called synchronously before the event reaches the Observer's per-writer queues
+_inline_listeners: list[Callable[[Event], None]] = []
```
soulFood5632, line 14: this gonna cause a bug: lets say someone spawns 2 seperate flows. this global variable will pick up events from both flows and shove them into the session objects `internals`. we will need to use some sort of context variables so that they are only listening to their own events or you could do some filtering on it

soulFood5632: In reference to the above, it would be ideal if they listeners were sectioned off by session. However since this is a stop gap fix. We may want to just say screw it and put in a simple filtering mechanism.

packages/railtracks/src/railtracks/observability/__init__.py
```diff
@@ -21,10 +27,12 @@
     "QueuePolicy",
     "Writer",
     "JsonlWriter",
+    "NodeInternalsCollector",
```
soulFood5632: I don't think this should be a part of the public API. it should be internal to our module

packages/railtracks/src/railtracks/observability/__init__.py
```diff
@@ -21,10 +27,12 @@
...
     "SCOPE_SESSION",
     "SCOPE_RETRIEVAL",
     "SCOPE_EVALUATION",
     "configure_writers",
+    "add_inline_listener",
```
soulFood5632: same here, prob should be internal

packages/railtracks/src/railtracks/observability/node_internals.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+
+    def __init__(self) -> None:
+        self._events: dict[str, list[Event]] = {}
+
+    def record(self, event: Event) -> None:
+        if event.event_type in _RELEVANT:
+            self._events.setdefault(event.scope_id, []).append(event)
+
+    def internals_for(self, session_id: str) -> dict[str, dict[str, Any]]:
```
soulFood5632, line 58: ok I now see where it si doing the filtering. This is fine for the stop gap

soulFood5632 (changes requested):
A couple of minor things.

packages/railtracks/src/railtracks/observability/configure.py
soulFood5632: See below comment (disregard above)

soulFood5632 (approved):
LGTM

packages/railtracks/src/railtracks/built_nodes/llm/model_invoker.py
```diff
@@ -49,6 +49,7 @@ async def _llm_observe(
     event = LLMResponseEvent(
         message_input=prev_message_history,
         output=response.message,
+        reported_model_name=response.message_info.model_name,
```
Amir-R25, line 52: There's just one thing we need to handle here, I'll create an issue for it. Any time an event is updated we need to also update the registry but registry is in my PR that hasn't landed yet

packages/railtracks/src/railtracks/events/llm.py
```diff
@@ -45,7 +47,10 @@ def event_type(self) -> str:
 
 @dataclass(kw_only=True)
 class LLMResponseEvent(LLMMessageBase):
+    """A completed round trip, plus what the provider reported about it."""
+
     output: Message | None
+    reported_model_name: str | None = None
```
Amir-R25, line 53: Let me know when you merge this, I'll update my branch

Amir-R25 (approved):
LGTM just ping me when you merge please so I'll update my branch

PR #1360, opened 2026-08-06
What the duck is happening?

Amir-R25:
### Registry (`events/registry.py`)
- Namespace-keyed table `NAMESPACE_COLUMNS: dict[str, dict[str, ColumnSpec]]` hard-coded for optimal visibility and maintainability (ironic I know)
- Reusable column blocks (`_CREATION`, `_PARENT_EVENT`, `_FAILURE`) mirror the event class hierarchy.
- ENUM members from`SpatialType`/`ParentType`.
- Completeness test walks every event class and asserts each field is reachable → drift caught at CI.

### Schema (`query/schema.py`)
- Translates `ColumnSpec` to a DuckDB type string. Registry stays DB-agnostic.
- `ENUM(...)` rendered with single-quote escaping.
- Potential for other schema mappings ie Postgres

### DuckDB connection (`query/connect.py`)
- `connect(path, namespaces)` opens an `EventQuery` and builds one typed view per namespace.
- Unknown namespaces raise `ValueError` pointing at `registry.py`.
- Import-time guard rejects a registry that shadows an envelope column.
- `refresh()` rebuilds views when new data lands.

### Demo

```python
from railtracks.query import connect

with connect(".railtracks/data/session.jsonl", ["llm"]) as q:
    total, avg_latency = q.con.execute(
        "SELECT SUM(input_tokens), AVG(latency) "
        "FROM llm WHERE event_type = 'llm.response'"
    ).fetchone()
    print(f"input tokens: {total}, avg latency: {avg_latency:.2f}s")
```

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes
Tests failing are not related to this work


packages/railtracks/src/railtracks/query/__init__.py
```diff
@@ -0,0 +1,5 @@
+from railtracks.events.registry import namespaces as list_namespaces
+
+from .connect import EventQuery, connect
+
+__all__ = ["connect", "list_namespaces", "EventQuery"]
```
soulFood5632, line 5: total nitpick but make this a verticial list so our future diffs are cleaner

packages/railtracks/src/railtracks/query/connect.py
```diff
@@ -0,0 +1,174 @@
...
+            f"railtracks.query requires the 'duckdb' package. Install with: {_INSTALL_HINT}"
+        ) from e
+    return duckdb
+
+
+def connect(path: Path | str, namespaces: list[str]) -> EventQuery:
+    """Open an ``EventQuery`` over the JSONL at ``path``, registering one view
+    per requested namespace. Every entry in ``namespaces`` must be present in
+    the registry's ``NAMESPACE_COLUMNS`` — unknown names raise ``ValueError``."""
+    duckdb = _require_duckdb()
```
soulFood5632: This feels like a strange way to imports but if you think its fine, I am good

packages/railtracks/src/railtracks/query/connect.py
```diff
@@ -0,0 +1,174 @@
...
+
+class EventQuery:
+    def __init__(
+        self,
+        con: DuckDBPyConnection,
+        path: Path | str,
+        namespaces: list[str],
+    ) -> None:
+        known = set(_registry_namespaces())
+        unknown = sorted(set(namespaces) - known)
```
soulFood5632, line 71: another total nitpick but why can't we just do equlaity here. If you are thinking that its possible we would have more known than namespaces, I'm thinking failing fast could be usefull there to catch things. I'll leave it up to you

packages/railtracks/src/railtracks/query/connect.py
```diff
@@ -0,0 +1,174 @@
...
+    def _teardown_views(self) -> None:
+        """Drop the namespace views this query owns, plus the events view."""
+        for ns in self.namespaces:
+            self.con.execute(f"DROP VIEW IF EXISTS {_sql_identifier(ns)}")
+        self.con.execute("DROP VIEW IF EXISTS events")
+
+    def _build_events_view(self, files: list[Path]) -> None:
+        """Create a view named ``events`` that unions all the JSONL files together."""
+        self.con.read_json(
+            [str(p) for p in files],  # type: ignore[arg-type] typing issue in duckdb
```
soulFood5632, line 114: did you have the abiltiy to post this issue on their repo

soulFood5632 (commented):
Will finish the review later but overall looks great, have a couple of funny nits

packages/railtracks/src/railtracks/query/connect.py
Amir-R25: Yeup: https://github.com/duckdb/duckdb-python/issues/576

packages/railtracks/src/railtracks/query/__init__.py
Amir-R25: `ruff format packages/railtracks/src/railtracks/query/__init__.py` reformats it to one line when I make that change

packages/railtracks/src/railtracks/query/connect.py
Amir-R25: yeup, sorry, moved the try/except inline

packages/railtracks/src/railtracks/query/connect.py
Amir-R25: It's intentional to enable subsetting. For instance, if it were an equality the demo script I posted in the PR message would break and raise a `ValueError`. It's a bit of ceremony to support subsets though.

packages/railtracks/src/railtracks/events/registry.py
Pooria90, line 1: I added these two fields to my last PR. We should register them:
`NodeDestruction.duration_seconds and LLMResponseEvent.reported_model_name`

packages/railtracks/src/railtracks/events/registry.py
```diff
@@ -0,0 +1,144 @@
...
+_FAILURE = {"exception_name": STRING, "exception_message": STRING}
+
+
+# ---- the table -------------------------------------------------------------
+
+NAMESPACE_COLUMNS: dict[str, dict[str, ColumnSpec]] = {
+    "llm": {
+        **_PARENT_EVENT,
+        "llm_id": STRING,
+        "model_provider": JSON,
```
Pooria90: This is a ModelProvider type, which is a str enum. Should we change to STRING?

packages/railtracks/tests/unit_tests/query/test_connect.py
```diff
@@ -0,0 +1,265 @@
...
+
+from __future__ import annotations
+
+import json
+from contextlib import closing
+from pathlib import Path
+
+import pytest
+
+pytest.importorskip("duckdb")
```
Pooria90, line 11: Is this needed here? Because it will skip the tests if people have not installed duckdb; but they want to experiment with it. It may give a notion of "passing".

packages/railtracks/src/railtracks/query/connect.py
```diff
@@ -0,0 +1,172 @@
...
+from .read import resolve_data_files
+from .schema import duckdb_columns
+
+if TYPE_CHECKING:
+    from duckdb import DuckDBPyConnection
+
+
+_INSTALL_HINT = "pip install 'railtracks[visual]'"
+
+_ENVELOPE_COLUMNS: dict[str, str] = {
```
Pooria90, line 19: I'm assuming this reflects the `observability/models.py::Event`. We should either build it from the events or have a unit test that validates it against that dataclass.

packages/railtracks/src/railtracks/query/read.py
```diff
@@ -0,0 +1,12 @@
...
+from pathlib import Path
+
+
+def resolve_data_files(path: Path | str) -> list[Path]:
+    p = Path(path)
+    if p.is_file():
+        return [p]
+    if p.is_dir():
+        return sorted(p.glob("*.jsonl"))
+    return []
```
Pooria90, line 12: should we raise a warning or error in case that there is no file or directory for `path`?

packages/railtracks/src/railtracks/query/connect.py
```diff
@@ -0,0 +1,172 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from railtracks.cli.io import print_error, print_status
```
Pooria90, line 6: why are we using these, not the library logger?

packages/railtracks/tests/unit_tests/query/_events.py
```diff
@@ -0,0 +1,94 @@
+"""Reusable event dicts for the query test modules.
+
+Fixtures use the real event shapes from ``railtracks.events`` — flat payloads with
+``spatial_parent_*`` / ``parent_*`` subfields as the write side emits them.
+"""
+
+from __future__ import annotations
+
+LLM_RESPONSE = {
```
Pooria90, line 9: I think it's better if we use the actual event classes to build these fixtures and test their construction and serialization,.

packages/railtracks/src/railtracks/query/connect.py
```diff
@@ -0,0 +1,172 @@
...
+    - ``ENUM(...)`` casts ``payload->>'key'`` (unquoted text) into the enum; casting
+      from JSON keeps the surrounding quotes and would break the enum lookup.
+    - Everything else casts ``payload->'key'`` (native JSON) to the target duckdb type.
+    """
+    if duckdb_type == "VARCHAR":
+        return f"payload->>{_sql_string(key)} AS {_sql_identifier(key)}"
+    if duckdb_type == "JSON":
+        return f"payload->{_sql_string(key)} AS {_sql_identifier(key)}"
+    if duckdb_type.startswith("ENUM("):
+        return f"CAST(payload->>{_sql_string(key)} AS {duckdb_type}) AS {_sql_identifier(key)}"
```
Pooria90: This line breaks "SELECT * FROM ..." if there is an unrecognized value for enums. My Claude suggested using `TRY_CAST`. But I'm not sure if this ever happens on the user side.

Code for error reproduction:
```python
import json, tempfile, pathlib
from railtracks.query import connect

d = pathlib.Path(tempfile.mkdtemp()); f = d / "e.jsonl"
f.write_text(json.dumps({
    "event_id": "e1", "event_type": "llm.response", "scope_type": "session",
    "scope_id": "s1", "parent_scope_id": None, "stamp": "2026-01-01T00:00:00+00:00",
    # any value not in the registry: a new SpatialType member, a renamed one,
    # or an older session file already on disk
    "payload": {"spatial_parent_spatial_type": "tool", "input_tokens": 5},
}) + "\n")

with connect(f, ["llm"]) as q:
    print(q.con.execute("SELECT SUM(input_tokens) FROM llm").fetchone())  # (5,) — fine
    print(q.con.execute("SELECT * FROM llm").fetchone())                  # raises; `tool` type does not exist.
```

Pooria90 (changes requested):

packages/railtracks/src/railtracks/query/__init__.py
soulFood5632: so dumb, ok sounds good. `ruff` should have a different behavior for `__init__`'s but I digress.

packages/railtracks/src/railtracks/query/connect.py
soulFood5632: ok sounds good to me.

packages/railtracks/src/railtracks/query/connect.py
soulFood5632: I can answer that one. command line prints are different from logs. We have employed this pattern for our cli stuff.

soulFood5632 (approved):
LGTM, Pooria has some good comments to review.

packages/railtracks/src/railtracks/query/connect.py
Amir-R25: So this was what we were doing initially but the code became too complex especially with maintainability cc @soulFood5632

It is far easier as maintainers to make sure we maintain this data contract. But yes there is a unit test that verifies this.

packages/railtracks/src/railtracks/events/registry.py
Amir-R25: I switched so it uses the `_enum_` just in case we change things in the future

packages/railtracks/tests/unit_tests/query/test_connect.py
Amir-R25: Pytest will mark it as "SKIPPED" in the results. I also would not worry too much because how will they be able to experiment with it if it's not installed?

packages/railtracks/src/railtracks/query/read.py
Amir-R25: `refresh()` supports openning a query before a session has written events so this is to protect against that.

packages/railtracks/tests/unit_tests/query/_events.py
Amir-R25: That will be quite involved. It was how we initially were doing registry but decided against it since it was really hard to both understand and maintain.

We still catch new/renamed event fields not reachable in `test_registry`

packages/railtracks/src/railtracks/query/connect.py
Amir-R25: Very good point, this would have messed up "older sessions" if we ever updated our events. Addressed

packages/railtracks/src/railtracks/events/registry.py
Amir-R25: Done

PR #1365, opened 2026-08-06
Add tool_nodes() accessor and creation-time duplicate tool name check

### LINKED ISSUE #1334: Expose an agent's tool list
### Description

# Expose an agent's tool list

Today, when you build an agent with `rt.agent_node(..., tool_nodes=[...])`, the tool nodes are captured inside the agent's `_invoke` closure (`llm_invoke_factory` in `built_nodes/llm/llm_helpers.py`) and are **not** stored anywhere inspectable. The built node class only carries `invoke`, `type`, `name`, `tool_info` (when applicable), `prepare_args`, and the middleware lists there is no public way to ask an agent "what tools do you have?". Its existence is both helpful from debugging, static reference (guards ensure some tool exist) and dynamic generation (conditional tool inject).

The only current workarounds are to keep the `tool_nodes` list yourself, or to
reach into `agent.invoke.__closure__` cells, which is private and fragile.

## Proposed API

Add a classmethod accessor on the agent (LLM) node, consistent with the
existing classmethod accessors (`tool_info()`, `name()`, `type()`):

```python
agent = rt.agent_node("agent", llm=..., tool_nodes=[fn_a, fn_b])

agent.tool_nodes()        # -> list[type[Node]]  (the connected tool node classes)

# convenience, if we want it:
[t.tool_info() for t in agent.tool_nodes()]   # the parsed Tool schemas the LLM sees Return a fresh list (defensive copy) so callers can't mutate internal state. An agent with no tools returns [] (never None). Sub-agents used as tools (agent-as-tool) appear in the list like any other node.
```

## Task
1. Create a internal varaible _tool_nodes, written by _closure_ accessable in other class method
2. Introduce public interface tool_nodes()

## Open questions / decisions
- Name: tool_nodes() matches the existing no-get_ classmethod style (tool_info(), name(), type()). 
- Return type: Node classes copy (list[type[Node]]) are the most flexible; callers can derive schemas via .tool_info().  Optionally also add a tool_schemas() helper returning list[Tool], but already iterable with .tool_info. A copy is returned to prevent mutation. (although may actually be useful to allow custom tool mutation), but not for the intended use here. 
- Home of the method: Base Node (uniform, default []) vs. agent-only (concept is really agent-specific). Leaning base Node with [] default for a uniform contract.

### LINKED ISSUE #1337: Same name tool trigger a run-time issue instead of static.
### Description

The uniqueness assertion lives at runtime in (llm_helpers.py#L191-L200), which filters tool_nodes by tool_info().name when the LLM dispatches a call. Each function_node() call mints a Node type in isolation, there is no shared registry, so nothing at creation time can see that a sibling node will collide. The clash only materializes when both are handed to the same agent and the model calls that name. 

This was never surfaced to the user/system as an exception at check time. Potentially related to #1336

agent = rt.agent_node("agent", llm=rt.llm.OpenAILLM("gpt-4o"), tool_nodes=[rt_square, rt_cube])
flow = rt.Flow("flow", entry_point=agent)
r = await flow.ainvoke("What is 5 squared and what is 5 cubed? If there is an error message, show the error message and the stack trace.")
print(r)
``` Output
LLMResponse(There was an error while trying to compute both 5 squared and 5 cubed. The error message was:

```
AssertionError("Multiple tool nodes found with name 'power'. This should not happen, please ensure all tool nodes have unique names. Offending nodes: [<class 'abc.PowerNode'>, <class 'abc.PowerNode'>]")
```

This error indicates that there are multiple tool nodes with the same name "power", which is causing a conflict.)

```

Aryan-Railtown:
## Summary

Closes #1334, closes #1337.

- Adds a `tool_nodes()` classmethod accessor (base `Node`, default `[]`) so an agent's connected tool nodes are inspectable, instead of only living inside the `llm_invoke` closure. Returns a fresh list (defensive copy) so callers can't mutate internal state.
- Adds a creation-time check (`_check_duplicate_tool_names`, raised as `NodeCreationError`) so two tool nodes sharing a `tool_info().name` fail immediately at `agent_node(...)` construction, naming every offending node class, instead of only surfacing later at LLM tool-dispatch time. This is most commonly hit by two `functools.partial` of the same parent function without a disambiguating `name=`.
- The pre-existing runtime dispatch check in `get_node_from_name` is kept as a defensive backstop (now an explicit `RuntimeError` instead of a bare `assert`) for tool lists assembled outside `agent_node(...)`.
- Docs: added a short "Tool names must be unique" note + `tool_nodes()` usage example to `function_tools.md`.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

Kept deliberately minimal/scoped to the two tickets . A same-callable reuse-caching mechanism that was explored alongside this work was dropped after confirming it wasn't restoring any actual prior behavior (the code path it resembled was dead/unreachable) and wasn't required by either ticket.

```python
square = rt.function_node(functools.partial(power, exp=2), name="square")   # name is imp here, otherwise we throw error
cube = rt.function_node(functools.partial(power, exp=3), name="cube")

agent = rt.agent_node("agent", llm=..., tool_nodes=[square, cube])
agent.tool_nodes()                          # -> [SquareNode, CubeNode]
[t.tool_info() for t in agent.tool_nodes()] # tool schemas the LLM sees
```


packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
```diff
@@ -195,9 +195,11 @@ def get_node_from_name(tool_name: str, tool_nodes: list[type[Node]]) -> type[Nod
         raise RuntimeError(
             f"LLM called tool '{tool_name}' which was not found in the provided tool nodes {[x.tool_info().name for x in tool_nodes]}",
         )
-    assert len(candidate_list) == 1, (
-        f"Multiple tool nodes found with name '{tool_name}'. This should not happen, please ensure all tool nodes have unique names. Offending nodes: {candidate_list}"
-    )
+    # This is a cirtical issue that breaks all tools, Agent should not allow this, a hard Error
+    if len(candidate_list) > 1:
+        raise RuntimeError(
```
Aryan-Railtown: this is just a backstop check, we should never reach here because the creation checks should catch the name duplication early on, but we still kept this just in case

packages/railtracks/src/railtracks/built_nodes/llm/node_builder.py
```diff
@@ -94,10 +95,16 @@ def llm(
             list(deepcopy(middleware)) if middleware is not None else []
         )
 
+        tool_nodes = list(connected_nodes) if connected_nodes else []
```
CoronRing: I am not sure if the connected node is deep copied (path: _build_dynamic_agent). As tool_node() promises immutability and a "fresh copy". 

I am fine having it as a dynamic one as well, so multi-thread becomes safe without the need of lock(Right now its not safe, we can get -> (update external) -> act and lead to issue. ), more like a design decision, but must be consistent with docstring.

Aryan-Railtown: changed this to deepcopy instead, but in this particular case deepcop works exactly the same way as shallow copy because it returns the exact same class object. 
Good point Guan!!

packages/railtracks/src/railtracks/built_nodes/function/node.py
```diff
@@ -237,12 +237,17 @@ async def wrapped_function(*args: _P.args, **kwargs: _P.kwargs) -> _TOutput:
     completed_node_type = builder.build()
 
     if issubclass(completed_node_type, Node):
+        rt_function: (
```
soulFood5632: I know this is pedantic but what is the point of this change. It's obvs fine to leave in, just curious haha

packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
```diff
@@ -195,9 +195,11 @@ def get_node_from_name(tool_name: str, tool_nodes: list[type[Node]]) -> type[Nod
         raise RuntimeError(
             f"LLM called tool '{tool_name}' which was not found in the provided tool nodes {[x.tool_info().name for x in tool_nodes]}",
         )
-    assert len(candidate_list) == 1, (
-        f"Multiple tool nodes found with name '{tool_name}'. This should not happen, please ensure all tool nodes have unique names. Offending nodes: {candidate_list}"
-    )
+    # This is a cirtical issue that breaks all tools, Agent should not allow this, a hard Error
```
soulFood5632: spelling error on critical

packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
soulFood5632: Lets use an assertion here so it removes the check when compiled in non-debug mode instead of a `RunTime` error

packages/railtracks/src/railtracks/built_nodes/llm/node_builder.py
```diff
@@ -94,10 +95,26 @@ def llm(
...
+        # deepcopy for consistency with the middleware lists above; in practice this is
+        # equivalent to a shallow list() here since the elements are classes, and
+        # copy.deepcopy treats `type` as atomic (returns the same class object rather than
+        # trying to duplicate it) -- see https://docs.python.org/3/library/copy.html.
+        # The container itself is still a fresh list either way, which is what tool_nodes()
+        # actually needs to protect against caller mutation.
+        tool_nodes = list(deepcopy(connected_nodes)) if connected_nodes else []
+        _check_duplicate_tool_names(tool_nodes)
+
+        # Kept on the builder so the built class can expose it via tool_nodes(). This list is
```
soulFood5632: lets get rid of this comment. Its bad style.

packages/railtracks/src/railtracks/built_nodes/llm/node_builder.py
```diff
@@ -94,10 +95,26 @@ def llm(
...
+        # a frozen snapshot taken at construction time -- there is currently no supported way
+        # to add/remove tools on an already-built agent. If dynamic tool mutation is added
+        # later, it will need its own synchronization; don't assume this attribute is safe to
+        # write to concurrently.
+        casted_instance._tool_nodes = tool_nodes
```
soulFood5632, line 106: cc: @Amir-R25 here is another example of what I spoke about on Friday. I think we should just take on this fix now before we come up with all these ad hoc solutions.

@Aryan-Railtown we can leave this in here for this PR but I am thinking we need rethink the typing approach of nodes a bit. 

> There are many properties that are specific to LLM nodes like the one you just added. I got rid of all the inheritance in this feature branch for all the nodes. However, I think the best solution has a thin LLMBase type that allows for the autocompletion which exposes `tool_nodes`, `model` etc. if and only if the node is an agent. I will be adding an issue to handle this and migrate the logic you added over.

packages/railtracks/src/railtracks/built_nodes/llm/node_builder.py
```diff
@@ -94,10 +95,26 @@ def llm(
             list(deepcopy(middleware)) if middleware is not None else []
         )
 
+        # deepcopy for consistency with the middleware lists above; in practice this is
+        # equivalent to a shallow list() here since the elements are classes, and
+        # copy.deepcopy treats `type` as atomic (returns the same class object rather than
+        # trying to duplicate it) -- see https://docs.python.org/3/library/copy.html.
+        # The container itself is still a fresh list either way, which is what tool_nodes()
+        # actually needs to protect against caller mutation.
+        tool_nodes = list(deepcopy(connected_nodes)) if connected_nodes else []
```
soulFood5632: be very careful here, changing it to an empty list could cause a major issue downstream. I think it would be better to keep as `None`

My concern is elsewhere when we decide whether it is a `StructuredLLM` or `ToolCallLLM` things could get really bad since we have decided to remove the ability for `StructuredToolCallingLLM`. This is complicated by us not differntiating an empty list from one that has tools and an empty list could prevent any chance for a type of `StructuredLLM` to be created, instead erroring out. 

So for know can we keep the same functionality as before.

packages/railtracks/src/railtracks/nodes/nodes.py
```diff
@@ -178,6 +184,19 @@ def tool_info(cls) -> Tool:
             "You must implement the tool_info method in your node"
         )
 
+    @classmethod
```
soulFood5632, line 187: cc: @Amir-R25 see above comment

soulFood5632 (changes requested):
Clever design, we may have a bit of a structural issue with the way we are differentiating the behavior of FunctionNode's and Agent's that we should address sooner rather than later. I would like to discusss with @Amir-R25 before we go ahead and merge.

packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
Aryan-Railtown: fixed, thanks

packages/railtracks/src/railtracks/built_nodes/function/node.py
Aryan-Railtown: this change was made originally when Guan introduced the reuse cache, but we talked and came to a conclusion that instead of taking in duplicate functions, we just raise errors. At that time i believe there was an extra step before returning the sync function node, hence the variable. 
TLDR; this was an old change which we can revert to the original :D

soulFood5632 (approved):
LGTM

PR #1367, opened 2026-08-07
Add Flow.connect() for post-run context and message history access (#1192)

### LINKED ISSUE #1192: [Feature] Many reasons to expose context variable object.
### Feature Request

# Context Variables Should Be Exportable (Public Live Access)
## Summary

Today the user context (`rt.context.*` → `MutableExternalContext`) is only reachable
from *inside* an active session, via the thread-local `runner_context` `ContextVar`
([`context/central.py:48`](../../packages/railtracks/src/railtracks/context/central.py#L48)).
There is no supported way for code *outside* the running flow to read or update the
context of an in-flight execution.

This ticket proposes exposing a **live, per-invocation handle** to the running context so
external code can observe it mid-execution and, optionally, write to it.

---

## Reason 1: Supports `yield`-like behaviour

Exposing the context lets events be surfaced mid-execution so the system can behave like a
generator that `yield`s out and is fed back in.

RT does not currently support `yield` natively — a node is invoked and its coroutine is
awaited to completion (a generator body is never driven; see the probe in the
`yield_sample` notebook). What *does* work today is awaiting an external async object:

```python
inbox: asyncio.Queue = asyncio.Queue()   # outside  -> flow  (answers)
outbox: asyncio.Queue = asyncio.Queue()  # flow     -> outside (questions)


@rt.function_node
async def ask_human(question: str) -> str:
    """Surface a question to the outside world and YIELD until an answer arrives.

    Args:
        question (str): The question to show to whoever is driving the flow.
    """
    await outbox.put(question)   # tell the outside world what we need
    answer = await inbox.get()   # <-- YIELD: suspend right here until input is fed in
    return answer


@rt.function_node
async def greeter() -> str:
    """A small multi-step flow that pauses for human input between steps."""
    name = await rt.call(ask_human, "What's your name?")
    colour = await rt.call(ask_human, f"Hi {name}! What's your favourite colour?")
    return f"{name} likes {colour}."
```

This works, but the in/out plumbing is passed around by hand — it is not natural. If the
context were a first-class, externally reachable object, the same "pause and exchange data
with the outside" pattern could be expressed through the context surface instead of ad-hoc
queues threaded through node signatures.

## Reason 2: No harm — and we cannot block it anyway

A determined user can already obtain the context object. Because `copy_context()` is a
**shallow** copy, every node in a session shares the *same* `MutableExternalContext`
instance by reference; a user can capture that reference from inside any node (e.g. stash
it in a closure or module global) and read/write it from anywhere afterwards. We are not
actually preventing external access — only making it undocumented and awkward.

Given that, an official accessor is strictly better: it makes the supported surface
explicit, lets us attach the right lifecycle/safety semantics (see Reason added below and
the concurrency ticket), and removes the incentive to reach into internals.

## Reason 3: Live observability and an external control plane

Right now the only external view of a run is the serialized graph written at **session
exit** (`save_state`, [`_session.py:166`](../../packages/railtracks/src/railtracks/_session.py#L166)) —
strictly post-mortem. Exposing the live context unlocks two classes of use that have no
clean path today:

- **Live observability:** a dashboard, the viz server, or a debugger can read in-flight
  state (progress counters, intermediate artifacts, "current step") *while* the flow runs,
  not after it finishes.
- **External control plane:** outside code can write signals the nodes poll — cancellation
  flags, feature toggles, rate/budget limits, a live "config" blob, or the running-clock
  example. This is the lightweight, read-mostly cousin of the queue/HIL pattern in
  Reason 1: for pushing values *in*, a shared context cell is simpler than a bespoke queue.

---

## Design: where to expose it

The non-obvious part. `Flow` is a **stateless template**, not a live run:
`ainvoke` does `Session(context=deepcopy(self._context))` per call
([`flow.py:76-91`](../../packages/railtracks/src/railtracks/orchestration/flow.py#L76-L91)),
so a single `Flow` can have **0, 1, or N** concurrent live sessions, each with its own
context. Therefore a bare `flow.get_context()` is ambiguous — *which* invocation?

The live context lives on the **`Session`**, so the handle should belong to the running
invocation, not the template. Options, roughly in order of preference:

note for proposed methods below, they do not replace the current, but offer new method.

### Option A — Run handle (recommended)

Give `Flow` a way to start a run that yields a handle while it executes:

```python
async with flow.start(*args) as run:      # run owns the live Session
    ctx = run.get_context()               # this invocation's MutableExternalContext
    ctx.put("clock", t)                   # external write
    snapshot = ctx.snapshot()             # safe read (see concurrency ticket)
    result = await run.result()
```

Unambiguous under concurrency (each `run` is one session), and gives a natural place to
scope the handle's lifetime (invalidate on `__aexit__`).

Same logic, but different form:
```python
ctx = rt.Context()                 # or a plain dict / MutableExternalContext whatever
task = asyncio.create_task(flow.ainvoke(vars, context=ctx))
```
We can make context and pass it in

### Option B — Return/expose the Session

`ainvoke` already has a sibling precedent: the `@session` decorator returns
`(result, session)` ([`_session.py:412-428`](../../packages/railtracks/src/railtracks/_session.py#L412-L428)).
A `Session.get_context()` accessor plus an opt-in way for `ainvoke` to surface its session
(callback, or an `ainvoke(..., return_session=True)`) would work. Lower-ceremony, but the
caller only gets the handle *after* the run for plain `ainvoke`, which defeats live access
unless paired with a callback.

### Option C — Active-session registry

A registry keyed by `session_id`, with `flow.get_context(session_id=...)` or a
"current single active session" shortcut. Convenient for the common single-run case, but
the global registry and the "exactly one active" assumption are fragile; treat as a
pragmatic shortcut layered on A/B, not the primary API.

### Cross-cutting requirements

- **Lifetime:** the handle must be invalidated at session teardown (`delete_globals()`,
  [`central.py:279`](../../packages/railtracks/src/railtracks/context/central.py#L279)) so
  reads after the run fail cleanly instead of returning stale ghosts.
- **Read vs write split:** consider exposing a safe **read snapshot** as the default public
  surface and treating external **writes** as an explicit, documented opt-in — the cost of
  writes (atomicity, determinism) is higher than reads. See the concurrency ticket.
- **Determinism caveat:** document that external writes between nodes make a run no longer a
  pure function of its inputs, which weakens replay/repro from the saved graph.

---

## Open questions

- Should external writes be allowed by default, or gated behind an explicit flag on the run
  handle (e.g. `flow.start(..., allow_external_writes=True)`)?
- Do we expose the raw `MutableExternalContext` or a narrower facade (`get`/`put`/`snapshot`
  only) to keep the public contract small?
- How does this interact with nested sub-flows, which share the parent's external context by
  reference via `prepare_new`? (A child write is visible to the parent — intended?)

## Recommended next step

Prototype **Option A** (`flow.start()` → run handle with `get_context()`), backed by a
`Session.get_context()` accessor, with reads returning a snapshot and writes behind an
opt-in flag. Pair with the concurrency-safety ticket before documenting writes as stable.

### Additional Context

_No response_

CoronRing:
Closes #1192.

`invoke`/`ainvoke` return only the flow's result, so a run's context — and the conversations its agents held — are unreachable once it finishes.

`flow.connect()` returns a `FlowConnection`, which you invoke in place of the Flow:

```python
conn = flow.connect()
result = await conn.ainvoke("text")

conn.context.get("progress")
```

## Surface

| accessor | |
|---|---|
| `.context` | the run's context; a live reference, readable mid-flight |
| `.message_histories()` | every model conversation, oldest first |
| `.session` | the backing `Session`, for anything without a dedicated accessor |
| `.session_id` | |
| `.connected` | whether anything has run yet |

`.message_histories()` covers **nested agents**, which the return value cannot: an `LLMResponse` carries only its own history, so a flow delegating to sub-agents has no way to surface theirs. Each entry is a `NodeMessageHistory` with `node_name`, `node_id`, `request_id` and `message_history`; nodes that made no model calls are omitted.

Accessors stay readable after an invocation raises, which is often when you most want them.

## Notes

- **Nothing else changes.** The plain `invoke`/`ainvoke` path behaves exactly as before and return values are untouched — no attribute is attached to `LLMResponse`, and custom return types work without modification. Both now delegate to a connection, so session setup lives in one place and `flow.py` gets shorter.
- **One invocation at a time.** A connection raises if you start a second while the first is in flight; open one per concurrent run. Sequential reuse is fine — accessors then describe the most recent invocation.
- **State is out of scope.** `session.payload()` and `session.info` are reachable through `.session`, but are deliberately not `FlowConnection` accessors, since that schema is expected to change.

## Testing

27 new tests in `test_flow_connection.py`. Everything else in the suite is unaffected — the failures on this branch (guardrails, middleware, `test_state.py`, MCP, live-LLM) are all present at baseline.

## Docs

New *Inspecting a Run* section in `documentation/invocation/flows.md`, with snippets in `docs/scripts/flows_sessions.py`.

### DIFF
--- a/docs/documentation/invocation/flows.md
+++ b/docs/documentation/invocation/flows.md
@@ -35,7 +35,7 @@ Sometimes you may want generic context items to be available across all runs of 
 Nothing about your flow or its return type has to change, and the plain `invoke`/`ainvoke` path is unaffected.
 
 ### Message Histories
-`connection.message_histories` gives you every model conversation in the run, oldest first. This includes nested agents — an `LLMResponse` carries only its own history, so a flow that delegates to sub-agents cannot surface theirs through its return value.
+`connection.message_histories` gives you every model conversation in the run, oldest first. This includes nested agents: an `LLMResponse` carries only its own history, so a flow that delegates to sub-agents cannot surface theirs through its return value.
 
 ```python
 --8<-- "docs/scripts/flows_sessions.py:connection_message_histories"
@@ -59,7 +59,8 @@ A connection handles one invocation at a time and raises if you start a second w
 
 !!! note
     A connection can be invoked repeatedly. Its accessors always describe the **most recent** invocation.
-
+    
+<!-- 
 ### Reaching Further
 `connection.session` exposes the underlying `Session` for anything without a dedicated accessor.
 
@@ -69,4 +70,4 @@ connection.session.payload() # the JSON that save_state writes
 ```
 
 !!! warning
-    `Session.info` and `Session.payload()` expose the run's state representation, whose shape is not yet stable. Prefer an accessor on `FlowConnection` where one exists.
+    `Session.info` and `Session.payload()` expose the run's state representation, whose shape is not yet stable. Prefer an accessor on `FlowConnection` where one exists. -->
--- a/docs/documentation/invocation/flows.md
+++ b/docs/documentation/invocation/flows.md
@@ -24,3 +24,49 @@ Sometimes you may want generic context items to be available across all runs of 
 
 !!! warning
     The context dictionaries used by Flows are passed by value. If a specific run mutates its context dictionary, those changes will not affect the original context or be passed to other runs.
+
+## Inspecting a Run
+`invoke` and `ainvoke` return only the flow's result, so anything a run built up along the way is gone once it finishes. Use `flow.connect()` when you need more than the result. It returns a `FlowConnection`, which you invoke in place of the Flow.
+
+```python
+--8<-- "docs/scripts/flows_sessions.py:connecting"
+```
+
+Nothing about your flow or its return type has to change, and the plain `invoke`/`ainvoke` path is unaffected.
+
+### Message Histories
+`connection.message_histories` gives you every model conversation in the run, oldest first. This includes nested agents — an `LLMResponse` carries only its own history, so a flow that delegates to sub-agents cannot surface theirs through its return value.
+
+```python
+--8<-- "docs/scripts/flows_sessions.py:connection_message_histories"
+```
+
+Each entry is a `NodeMessageHistory` with `node_name`, `node_id`, `request_id` and `message_history`. Nodes that made no model calls are omitted.
+
+### Failed Runs
+The accessors stay readable after an invocation raises, which is often when you most want them.
+
+```python
+--8<-- "docs/scripts/flows_sessions.py:connection_failure"
+```
+
+### Concurrency
+A connection handles one invocation at a time and raises if you start a second while the first is in flight. Open a connection per concurrent run.
+
+```python
+--8<-- "docs/scripts/flows_sessions.py:connection_concurrent"
+```
+
+!!! note
+    A connection can be invoked repeatedly. Its accessors always describe the **most recent** invocation.
+
+### Reaching Further
+`connection.session` exposes the underlying `Session` for anything without a dedicated accessor.
+
+```python
+connection.session.info      # the run graph: nodes, requests, timing
+connection.session.payload() # the JSON that save_state writes
+```
+
+!!! warning
+    `Session.info` and `Session.payload()` expose the run's state representation, whose shape is not yet stable. Prefer an accessor on `FlowConnection` where one exists.
--- a/docs/documentation/invocation/flows.md
+++ b/docs/documentation/invocation/flows.md
@@ -35,13 +35,13 @@ Sometimes you may want generic context items to be available across all runs of 
 Nothing about your flow or its return type has to change, and the plain `invoke`/`ainvoke` path is unaffected.
 
 ### Message Histories
-`connection.message_histories` gives you every model conversation in the run, oldest first. This includes nested agents: an `LLMResponse` carries only its own history, so a flow that delegates to sub-agents cannot surface theirs through its return value.
+`connection.message_histories()` gives you every model conversation in the run, in the order the runs were recorded. This includes nested agents: an `LLMResponse` carries only its own history, so a flow that delegates to sub-agents cannot surface theirs through its return value.
 
 ```python
 --8<-- "docs/scripts/flows_sessions.py:connection_message_histories"
 ```
 
-Each entry is a `NodeMessageHistory` with `node_name`, `node_id`, `request_id` and `message_history`. Nodes that made no model calls are omitted.
+Each entry is a `NodeMessageHistory` with `node_name`, `node_id`, `request_id` and `message_history`. Nodes that made no model calls are omitted. Nodes called concurrently have no guaranteed order between them.
 
 ### Failed Runs
 The accessors stay readable after an invocation raises, which is often when you most want them.
@@ -59,7 +59,7 @@ A connection handles one invocation at a time and raises if you start a second w
 
 !!! note
     A connection can be invoked repeatedly. Its accessors always describe the **most recent** invocation.
-    
+
 <!-- 
 ### Reaching Further
 `connection.session` exposes the underlying `Session` for anything without a dedicated accessor.
--- a/docs/scripts/flows_sessions.py
+++ b/docs/scripts/flows_sessions.py
@@ -40,3 +40,53 @@ flow = rt.Flow(
 context_injected_flow = flow.update_context({"run_specific_key": "run_specific_value"})
 response = context_injected_flow.invoke("What is the value of shared_key and run_specific_key?")
 # --8<-- [end: injecting_context]
+
+
+# --8<-- [start: connecting]
+# .connect() gives you a FlowConnection, which you invoke in place of the Flow.
+connection = flow.connect()
+response = connection.invoke("What is the capital of France?")
+
+# The run's context is still readable afterwards.
+print(connection.context.get("shared_key"))
+# --8<-- [end: connecting]
+
+
+# --8<-- [start: connection_message_histories]
+connection = flow.connect()
+response = connection.invoke("What is the capital of France?")
+
+for history in connection.message_histories:
+    print(history.node_name)
+    for message in history.message_history:
+        print(f"  {message.role}: {message.content}")
+# --8<-- [end: connection_message_histories]
+
+
+# --8<-- [start: connection_failure]
+connection = flow.connect()
+
+try:
+    connection.invoke("What is the capital of France?")
+except Exception:
+    # The context is readable even though the run raised.
+    print("failed at stage:", connection.context.get("stage", default="unknown"))
+# --8<-- [end: connection_failure]
+
+
+# --8<-- [start: connection_concurrent]
+import asyncio  # noqa: E402
+
+questions = ["Capital of France?", "Capital of Japan?", "Capital of Peru?"]
+
+
+async def ask_all():
+    # One connection per concurrent run.
+    connections = [flow.connect() for _ in questions]
+    await asyncio.gather(*(c.ainvoke(q) for c, q in zip(connections, questions)))
+    return connections
+
+
+for connection in asyncio.run(ask_all()):
+    print(connection.session_id, connection.context.get("shared_key"))
+# --8<-- [end: connection_concurrent]
--- a/docs/scripts/flows_sessions.py
+++ b/docs/scripts/flows_sessions.py
@@ -56,7 +56,7 @@ print(connection.context.get("shared_key"))
 connection = flow.connect()
 response = connection.invoke("What is the capital of France?")
 
-for history in connection.message_histories:
+for history in connection.message_histories():
     print(history.node_name)
     for message in history.message_history:
         print(f"  {message.role}: {message.content}")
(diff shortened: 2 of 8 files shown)


CoronRing:
A notebook walking through the interface will be added here shortly.

CoronRing:
[1192_interface_sketch.ipynb](https://github.com/user-attachments/files/30815983/1192_interface_sketch.ipynb)

Note book that illustrate the interface changes and advanced dynamic use.

docs/documentation/invocation/flows.md
```diff
@@ -23,4 +23,51 @@ Sometimes you may want generic context items to be available across all runs of
...
+A connection handles one invocation at a time and raises if you start a second while the first is in flight. Open a connection per concurrent run.
+
+```python
+--8<-- "docs/scripts/flows_sessions.py:connection_concurrent"
+```
+
+!!! note
+    A connection can be invoked repeatedly. Its accessors always describe the **most recent** invocation.
+    
+<!--
```
soulFood5632, line 63: lets just get rid of this. No need to keep in commented markdown

docs/scripts/flows_sessions.py
```diff
@@ -39,4 +39,54 @@
...
+try:
+    connection.invoke("What is the capital of France?")
+except Exception:
+    # The context is readable even though the run raised.
+    print("failed at stage:", connection.context.get("stage", default="unknown"))
+# --8<-- [end: connection_failure]
+
+
+# --8<-- [start: connection_concurrent]
+import asyncio  # noqa: E402
```
soulFood5632: no need for noqa here. The files in the docs section should be ignored by `ruff` linters. If that is not the case we should modify it to be that way/

docs/scripts/flows_sessions.py
```diff
@@ -39,4 +39,54 @@
...
+
+questions = ["Capital of France?", "Capital of Japan?", "Capital of Peru?"]
+
+
+async def ask_all():
```
soulFood5632: i feel like stylically we could come with a better example that does not require such abstraction. I think its nice if the docs docs code avoid the making of functions and keeps things as scripts. Something like:

```python 
connections = []
futures = []
for question in ["Capital of France?", " Capital of Japan?"]:
  c = flow.connect()
  connections.append(c)
  f = c.ainvoke(question)
  futures.append(f)


results = await asyncio.gather(*futures)

print([... for c in connections])
```

packages/railtracks/src/railtracks/orchestration/connection.py
```diff
@@ -0,0 +1,189 @@
...
+
+        A live reference, not a copy. Context is not intended to be modified.
+        
+        """
+        return self._require_session().context
+
+    @property
+    def session_id(self) -> str:
+        """Identifier of the session backing the most recent invocation."""
+        return self._require_session()._identifier
```
soulFood5632: we shouldn't be accessing protected variables of other objects. We should either make the property public or come up with some other way to access it through a `@property` wrapper.

packages/railtracks/src/railtracks/orchestration/connection.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @property
+    def session(self) -> Session:
```
soulFood5632, line 140: Honestly, unless you see a good reason to include this in a public API, I feel like this an uneeded thing to add. The users should never really be interacting with the Session type as its designed to be internal.

packages/railtracks/src/railtracks/orchestration/connection.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        The session backing the most recent invocation, for inspection.
+
+        Already closed; the connection owns its lifecycle. `Session.payload()`
+        and `Session.info` expose the run's state representation.
+        """
+        return self._require_session()
+
+    @property
+    def message_histories(self) -> List[NodeMessageHistory]:
```
soulFood5632, line 149: I really like the idea of this but I don't think it should be a property. the graph traversals is fairly expensive, and I think it should be a called on demand property.

packages/railtracks/src/railtracks/orchestration/connection.py
```diff
@@ -0,0 +1,189 @@
...
+                    if node_type is not None
+                    else "<unknown>",
+                    node_id=request.sink_id,
+                    request_id=request.identifier,
+                    message_history=history,
+                )
+            )
+
+        # heap ordering is insertion-based; step ordering is the run's own
+        histories.sort(key=lambda h: info.request_forest[h.request_id].stamp.step)
```
soulFood5632: I cannot for life of me figure out what this comment is saying what does "step ordering the run's own" mean?

packages/railtracks/src/railtracks/orchestration/flow.py
```diff
@@ -73,31 +71,26 @@ def update_context(self, context: dict[str, Any]) -> Flow[_P, _TOutput]:
         new_obj._context.update(context)
         return new_obj
 
+    def connect(self) -> FlowConnection[_P, _TOutput]:
```
soulFood5632, line 74: Marking this for discussion because I don't know how I feel about `connect`. To me connect just doesn't really describe what you are doing. I feel like something like:
- `container`
- `persist`
- `store`
- `executor`

worthy of discussion

soulFood5632 (changes requested):
Looks very good. Very clever design. I left a couple of comments around some nitpicky comments. Take a look at my comments around the name. I think it would be good to discuss.

docs/documentation/invocation/flows.md
CoronRing: This was to access a session object, which we are not "exposing" to the user, however, right now it contains some stuff that user will 100% need, for example tokens, number of runs and most importantly toolcalls. 

I thought it would be useful as a partial read for now, agents/humans might read if they really must. Session is indeed exposed, not saying it is somehow is a bit weird.

docs/scripts/flows_sessions.py
CoronRing: I copied over from code, forgot to remove, done

packages/railtracks/src/railtracks/orchestration/connection.py
CoronRing: As above, we otherwise have no way of getting message history, ie after one flow, i want to save state and continue chatting, right now we need message history. Imagine

who is steve
Agent 1: Searching steve->doc 1
Agent 2: read(doc 1) -> steve is XYZ

We need to save/restore history of both agent 1 and agent 2. Right now this is strictly not possible becase invoke only pass in one set. Could expand in future, but right now, message history is a must to get it

packages/railtracks/src/railtracks/orchestration/connection.py
CoronRing: It means: iterating the heap dict gives you whatever order keys were inserted, we don't trust it. Instead we should sort by the run's own stamp.step counter instead.

But when I had claude confirm, I realized I was wrong, stamp.step is not creation order, its some "last updated" order, which is actually completion order. So in theory we need to invert sort.

invoke order        : ['A', 'B', 'C'] 
model finish order: ['C', 'B', 'A']
heap order        : ['A', 'B', 'C'] 
by current step   : ['C', 'B', 'A']    <- what we returned
New order         : ['A', 'B', 'C'] 

I updated the code to actually removed that line that depend on stamp.step because I am not certain of its case and dict insert order should work 99% of the time. I added statement "Concurrently called nodes have no guaranteed order." Which I think stamp.step do not promise as well.

packages/railtracks/src/railtracks/orchestration/connection.py
CoronRing: made public

docs/documentation/invocation/flows.md
soulFood5632: I was referencing the fact that the docs are escaped. they didn't show up when I ran `mkdocs`

packages/railtracks/src/railtracks/orchestration/connection.py
soulFood5632: I like the property. I was just suggesting it should be accessed via `message_histories()` instead of `message_histories`

packages/railtracks/src/railtracks/orchestration/connection.py
soulFood5632: Yea I like getting rid of the logic. Good catch thanks.

docs/documentation/invocation/flows.md
soulFood5632: and to answer your question, lets just keep it.

packages/railtracks/src/railtracks/orchestration/connection.py
CoronRing: the code actually did updated it to message_histories(), this part was to justify the PR body changes, which I realized I was not clear as it does not prompt an update message on edit.

docs/documentation/invocation/flows.md
CoronRing: I understand, I meant say that when agents reference the docs its the .md version, they can still see it, as I agree we should not let user know too much about session.

docs/scripts/flows_sessions.py
CoronRing: Example updated

packages/railtracks/src/railtracks/orchestration/connection.py
CoronRing: Resolving as sessions expose advanced parameters required for advanced use. (the commented out component in docs)

soulFood5632 (approved):
LGTM

PR #1382, opened 2026-08-12
fix: emit per-call ToolFailure and a single Runtime result in ToolUseEvaluator

### LINKED ISSUE #1250: `ToolUseEvaluator` misfiles per-call `Runtime` results and emits them twice
### Description

**File:** `packages/railtracks/src/railtracks/evaluations/evaluators/tool_use_evaluator.py`
  **Function:** `ToolUseEvaluator._extract_tool_stats` (lines ~117–144)

  ### Summary
  In the per-tool-call loop, the per-call `Runtime` result is appended to the wrong
  bucket and is also emitted twice. This mislabels the `ToolFailure` metric, duplicates
  `Runtime` leaves in the aggregate forest, and leaves orphan nodes behind.

  ### Details

  **1. Per-call `Runtime` result is appended to the `ToolFailure` bucket.**
  Lines 117–126 build a `ToolMetricResult` with `Runtime`'s name/id and
  `value = tool.runtime if tool.runtime is not None else 0.0`, but append it to
  `results[METRICS["ToolFailure"]]`:
  ```python
  metric_result = ToolMetricResult(
      result_name=f"{METRICS['Runtime'].name}/{tool_name}",
      metric_id=METRICS["Runtime"].identifier,
      ...
      value=tool.runtime if tool.runtime is not None else 0.0,
  )
  forest.add_node(metric_result)
  results[METRICS["ToolFailure"]].append(metric_result)   # <-- wrong bucket
  Consequences:
  - The ToolFailure metric ends up holding runtime values, not 0/1 failure indicators.
  - No genuine per-call failure indicator is ever produced — ToolFailure is effectively
  unimplemented.

  2. Runtime result is emitted twice.
  When tool.runtime is not None (lines 132–144) a second, near-identical Runtime
  result is created and appended to results[METRICS["Runtime"]]. Each runtime leaf then
  exists twice (two ids): once (with 0.0 fallback) in the ToolFailure bucket, once in
  the Runtime bucket.

  Impact

  - ToolFailure entries in metric_results are mislabeled (they carry runtimes).
  - The aggregate forest contains duplicate Runtime leaves.
  - The first runtime leaves (added at line 125) are never referenced by any aggregate node
  or root, so they become orphan nodes in forest.nodes.
  - This does not break the visualizer (the report stays schema-valid); it produces
  incorrect/duplicated data.

  Expected behavior

  - Emit exactly one Runtime result per call → results[Runtime] (used by aggregation).
  - Emit a ToolFailure result per call with
  value = 1 if tool.status == Status.FAILED else 0 → results[ToolFailure].
  - Leave FailureRate and UsageCount unchanged.

  Suggested fix

  Compute failure status first, then emit one Runtime leaf and one ToolFailure leaf into
  their correct buckets, and remove the duplicate emit.

DavidGao520:
## Summary

Fixes the per-call metric emission in `ToolUseEvaluator._extract_tool_stats`. Closes #1250.

**Root cause** — `packages/railtracks/src/railtracks/evaluations/evaluators/tool_use_evaluator.py:117-144` (pre-fix). Inside the per-tool-call loop the evaluator built a `ToolMetricResult` carrying `Runtime`'s name and `metric_id` but appended it to `results[METRICS["ToolFailure"]]`, then — when `tool.runtime is not None` — built a second, near-identical result and appended that one to `results[METRICS["Runtime"]]`. Three consequences:

- `ToolFailure` held runtime values instead of 0/1 indicators, so the metric was effectively unimplemented.
- Every runtime leaf existed twice in `forest.nodes` under two ids.
- The first copy was never referenced by an aggregate node or root (`_aggregate_per_run` walks `results[Runtime]`, which only ever contained the second copy), so it stayed in `forest.nodes` as an orphan.

**Fix** — compute the failure status once, emit a real `ToolFailure` leaf (`value = 1 if tool.status == Status.FAILED else 0`, `result_name = "ToolFailure/<tool>"`) into `results[ToolFailure]`, and drop the misfiled first emit so exactly one `Runtime` leaf per call goes into `results[Runtime]` and the forest. `FailureRate` and `UsageCount` are untouched, and the `failure_count` bookkeeping that feeds `FailureRate` now reuses the same status check.

The `ToolFailure` leaf is deliberately **not** added to the forest: nothing aggregates per-call failures (`_aggregate_per_run` handles `Runtime`; `_aggregate_across_runs` handles `FailureRate`/`UsageCount`), so adding it would recreate the orphan-node problem the issue calls out. This matches `JudgeEvaluator`, which puts its reasoning results in `results` only (`judge_evaluator.py:97-108`). Flagging it because the issue text doesn't specify either way — happy to add it to the forest instead if you'd rather every leaf be present there.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [x] Lint & format pass (`ruff check . --no-fix && ruff format --check .`) — "All checks passed!" / "261 files already formatted"
- [ ] Docs updated if user-facing behavior changed — no docs reference these metric buckets
- [x] Breaking changes include migration notes — n/a, but see Notes: `ToolFailure` values in `metric_results` change meaning

## Testing

Three tests added to `packages/railtracks/tests/unit_tests/evaluations/evaluators/test_tool_use_evaluator.py`, each failing before the change and passing after:

- `test_run_tool_failure_results_are_failure_indicators` — before: `["Runtime/get_price", ...] != ["ToolFailure/get_price", ...]`; after: values `[0, 1]` for one completed + one failed call.
- `test_run_emits_one_runtime_result_per_tool_call` — before: `[0.1, 0.1, 0.3, 0.3] != [0.1, 0.3]`; after: one leaf per call.
- `test_run_forest_has_no_orphan_nodes` — walks `forest.roots` through `children` and asserts every node in `forest.nodes` is reachable. Before: one extra unreachable node; after: passes.

```
$ uv run pytest packages/railtracks/tests/unit_tests/evaluations/evaluators/test_tool_use_evaluator.py -q   # before the fix
3 failed, 10 passed in 0.06s

$ uv run pytest packages/railtracks/tests/unit_tests/evaluations packages/railtracks/tests/integration_tests/evaluations -q   # after
196 passed in 0.11s

$ uv run pytest packages/railtracks/tests/unit_tests/ packages/railtracks/tests/integration_tests/ -q
1 failed, 2144 passed, 7 skipped, 283 warnings, 2 errors in 88.75s

$ ruff check . --no-fix && ruff format --check .
All checks passed! / 261 files already formatted
```

The one failure (`unit_tests/llm/test_type_mapping.py::test_builtin_function_raises_runtime_error`) and two collection errors (`integration_tests/rt_mcp/test_mcp.py`) are pre-existing — I confirmed they reproduce identically on an unmodified checkout of the same base commit (`67045b1`) and are unrelated to this change.

## Notes

- **Behavior change to be aware of:** `ToolFailure` entries in `EvaluatorResult.metric_results` previously contained runtimes; they now contain `0`/`1`. Anything downstream that was (accidentally) reading runtimes out of that bucket will see different numbers — that is the point of the fix, but worth a look if you have consumers I can't see.
- Aggregate `Runtime` statistics (mean/median/std on `ToolAggregateNode`) were already computed from `results[Runtime]` only, so removing the duplicate leaf does not change any reported aggregate value — it only removes duplicate/orphan entries from `forest.nodes`.
- Test files are excluded from ruff (`exclude = [..., "tests"]` in the root `pyproject.toml`), so the new tests follow the surrounding file's existing style rather than `ruff format` output.

---
This contribution was developed with AI assistance (Claude Code). I reviewed the change and the test results before submitting.


soulFood5632:
I would like to confirm this with @Amir-R25 as the code owner of evals before we allow the merge but it looks good from my end.

Amir-R25:
Thanks for the PR @DavidGao520, please update your branch from main when you get a chance and we're good to merge! 🚀

Amir-R25 (approved):
Amazing, thanks, looks good on my end!

PR #1383, opened 2026-08-12
fix: keep the text a model returns alongside its tool calls

### LINKED ISSUE #1235: Railtracks do not gracefully handle multi-part agent response.
### Description

# Issue
Often times, agents return response in multiple chunks: 

ie, in anthropic:
```
"content": [
    {
      "type": "text",
      "text": "The current weather in London is: "
    },
    {
      "type": "tool_use",
      "id": "toolu_01EsQDnwhMTqLymnSdfA9frB",
      "name": "get_weather",
      "input": {
        "location": "London"
      },
      "caller": {
        "type": "direct"
      }
    }
  ],
```
Note how context includes context and tool use. Currently if we notice a tool call, the actual text content is dropped. 

## Expected
```
The current weather in London is: 
>tool call
(tool call result returned)
15 degrees
```
## Current
```
>tool call
(tool call result returned)
15 degrees
```

## Reproduce step:
Any agent with prompt: 

```
Can you check weather for me?? london? State you are making call before call

I will call XX with YY
>call
Show result
```
Tool schema:
```
{
  "type": "object",
  "properties": {
    "location": {
      "type": "string",
      "description": "The city and state, e.g. San Francisco, CA"
    },
    "unit": {
      "type": "string",
      "enum": [
        "celsius",
        "fahrenheit"
      ],
      "description": "The unit of temperature"
    }
  },
  "required": [
    "location"
  ]
}
```

### Additional Context

_No response_

### Operating System

_No response_

### railtracks Version

_No response_

### Python Version

_No response_

DavidGao520:
## Summary

Models frequently answer with prose *and* a tool call in the same message (the Anthropic
example in #1235 returns a `text` block followed by a `tool_use` block; OpenAI does the same
with `message.content` set next to `message.tool_calls`). Railtracks kept only the tool calls
and threw the text away, so "I will check the weather in London for you" never reached the
message history, the saved session, or the model's own next turn.

Closes #1235.

**Root cause**

`LiteLLMWrapper._chat_with_tools_handler_base` builds the assistant message from the tool
calls alone and never looks at `choice.message.content`
(`packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py:691`). Because
`Content = Union[str, List[ToolCall], ToolResponse, BaseModel]`, an assistant message could
hold *either* text *or* tool calls, so there was nowhere for the text to go. The same drop
happened in three more places:

- the streaming handlers, which accumulate the text deltas and then discard
  `accumulated_content` as soon as any tool call is present
  (`_litellm_wrapper.py:480` and `:574`);
- `_to_litellm_message`, which hardcoded `base["content"] = ""` for a tool-call message
  (`_litellm_wrapper.py:875`), so even a preserved text would not be sent back to the
  provider on the next request;
- `OutputLessToolCallLLMBase._handle_response`
  (`packages/railtracks/src/railtracks/built_nodes/concrete/_tool_call_base.py:205`) and
  `ChatToolCallLLM.invoke` (`chat_tool_call_llm.py:52`), which rebuild a fresh
  `AssistantMessage` from `message.content` before appending it to the history.

**Fix**

`AssistantMessage` gains an optional `text` field for the prose that accompanied a list of
tool calls (`llm/message.py:363`). This keeps the existing `Content` union and the
`isinstance(content, list)` checks that the whole tool-calling loop is built on — no caller
that ignores `text` changes behaviour. The four drop sites now carry the text through, and
`state.serialize.encode_message` emits it as an additive `"text"` key so saved sessions (and
the visualizer reading them) can show it.

Nothing about the fields the FE already reads changed: `"text"` is only present when the
model actually returned prose with its tool calls.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [ ] Docs updated if user-facing behavior changed — no docs page describes tool-call message
      content; the new field is documented in the `AssistantMessage` docstring
- [ ] Breaking changes include migration notes — n/a, the new argument is optional

## Testing

```
uv run pytest -s -v packages/railtracks/tests/unit_tests/ packages/railtracks/tests/integration_tests/
```
→ `1 failed, 2148 passed, 7 skipped, 2 errors`

The one failure (`unit_tests/llm/test_type_mapping.py::test_builtin_function_raises_runtime_error`)
and the two errors (`integration_tests/rt_mcp/test_mcp.py`) are present on an unmodified
checkout of this branch's base commit as well — they are unrelated to this change.

```
uv run ruff check packages/railtracks   # All checks passed!
uv run ruff format --check packages/railtracks   # 257 files already formatted
```

New tests, all of which fail on `main` and pass with the fix:

- `unit_tests/llm/models/test_litellm_wrapper.py::TestCompletionMethods::test_chat_with_tools_keeps_text_returned_with_tool_calls[sync|async]`
  — a provider response carrying content *and* `tool_calls` in one message; before the fix
  `result.message` has no trace of the text at all.
- `...::test_prepare_response_keeps_streamed_text_with_tool_calls` — same for the streaming path.
- `...::TestHelpers::test_to_litellm_message_tool_call_list_with_text` — the text is sent back
  to the provider instead of `""`.
- `integration_tests/library/test_tool_calling_llms.py::TestSimpleToolCalling::test_text_returned_with_tool_call_is_kept`
  — full agent run: the assistant message in `response.message_history` keeps the preamble.
  This one covers the node-layer drop independently: with only `_tool_call_base.py` reverted it
  fails with `assert None == 'I will call secret_phrase to find out.'`.
- `unit_tests/utils/serialization/test_serialize.py::test_encode_message_keeps_text_sent_with_tool_calls`
  and `::test_encode_message_omits_text_when_there_is_none`.

## Notes

- **Round-trip to the provider.** `_to_litellm_message` now sends the real text instead of
  `""` for an assistant tool-call turn. Both OpenAI and Anthropic accept content next to
  tool calls (it is what they returned in the first place), and the previous `""` was
  arguably the more fragile value for Anthropic. Worth a sanity check against a live model,
  since the test suite mocks litellm.
- **Why a field and not a mixed content list.** Making `Content` allow
  `List[str | ToolCall]` would touch every `assert all(isinstance(x, ToolCall) ...)` and the
  `Message.tool_calls` property across the tool-calling loop. The optional field keeps the
  diff to the places that actually dropped data.
- **Not included:** displaying the preamble in the chat UI. `interaction/interactive.py`
  sends only the final `response.content` to `chat_ui`, and `ChatToolCallLLM.chat_ui` is
  never assigned anywhere in the package (`chat_ui = None`), so wiring the text into that
  surface is a separate change. The data it needs is now in the history and in the serialized
  session. Same for the visualizer, which the issue thread says is being handled separately.

---
This contribution was developed with AI assistance (Claude Code). I reviewed the change and the test results before submitting.


soulFood5632:
Hey David,

Thanks for the PR, and sorry for the slow response. We've been pushing to get some work through for our `1.5.0` release, so we didn't want to merge anything before that was polished off.

I wanted to take the time to write a proper response since this is a very clever solution. However, I have some design concerns about your approach that I'd like to work through to make sure we're on the same page.

You added a field called `text`, which contains the added string. A better design might be to create a new type that allows both `ToolCalls` and text to live within the content type `_T`. That way, we stay consistent in keeping the LLM's response entirely within content.

Let me know your thoughts on this approach.

DavidGao520:
Thanks — that's a better shape, and you're right that a parallel text field breaks the "the response lives in content" invariant. Pushed as e4ac353.

There's now a ToolCalls content type holding both:

class ToolCalls(List[ToolCall]):
    def __init__(self, tool_calls: Iterable[ToolCall] = (), text: str | None = None):
        super().__init__(tool_calls)
        self.text = text

AssistantMessage loses the text parameter entirely; a tool-calling turn is AssistantMessage(content=ToolCalls(calls, text=...)), and everything reads the text off msg.content.text.

The one judgement call I'd like your read on: I made it a list[ToolCall] subclass rather than a fresh model. That keeps for tc in msg.content, msg.content[0], msg.content == [call], isinstance(content, list), and AssistantMessage(content=[ToolCall(...)]) all working exactly as they do today — Message.tool_calls, LLMResponse.tool_invocations and _to_litellm_message needed no changes to their list handling, and neither does any user code. AssistantMessage.__init__ normalizes a bare list into a ToolCalls, so internally a tool-calling turn is always one shape.

The tradeoff is that a list subclass carrying an attribute is less explicit than, say:

class ToolCalls(BaseModel):
    tool_calls: list[ToolCall]
    text: str | None = None

which would sit better next to ToolCall/ToolResponse, at the cost of a breaking change to content. If you'd rather have that and are happy to take the break in 1.5.0, say the word and I'll switch it over — it's a small change from here.

One place the split still shows: AssistantMessage.encode() lifts the text to a top-level "text" key, because a ToolCalls serializes as a bare JSON array and the text would be dropped from traces otherwise. That mirrors how providers put it on the wire (content + tool_calls), but if you'd prefer the encoded form to nest instead, that's easy to change too.

Checks on my end: ruff check / ruff format clean, and the full unit + integration suite passes — 2590 passed. The three that don't (test_builtin_function_raises_runtime_error, and the two test_mcp.py errors) fail the same way on a clean checkout of main, so they're not from this branch.

Added coverage for the new type: list behaviour, deepcopy preserving the text, normalization of a plain list, and both encode cases.

soulFood5632:
> The one judgement call I'd like your read on: I made it a list[ToolCall] subclass rather than a fresh model. That keeps for tc in msg.content, msg.content[0], msg.content == [call], isinstance(content, list), and AssistantMessage(content=[ToolCall(...)]) all working exactly as they do today — Message.tool_calls, LLMResponse.tool_invocations and _to_litellm_message needed no changes to their list handling, and neither does any user code. AssistantMessage.**init** normalizes a bare list into a ToolCalls, so internally a tool-calling turn is always one shape.
> 
> The tradeoff is that a list subclass carrying an attribute is less explicit than, say:
> 
> class ToolCalls(BaseModel): tool_calls: list[ToolCall] text: str | None = None
> 
> which would sit better next to ToolCall/ToolResponse, at the cost of a breaking change to content. If you'd rather have that and are happy to take the break in 1.5.0, say the word and I'll switch it over — it's a small change from here.
> 
> One place the split still shows: AssistantMessage.encode() lifts the text to a top-level "text" key, because a ToolCalls serializes as a bare JSON array and the text would be dropped from traces otherwise. That mirrors how providers put it on the wire (content + tool_calls), but if you'd prefer the encoded form to nest instead, that's easy to change too.

Thanks for your throurough review of my comment. I think your current solution is fine. we can flag it as a future improvement if the design becomes a bit too brittle

soulFood5632 (approved):
LGTM

PR #1385, opened 2026-08-12
Add shared CLAUDE.md and comment-consistency skill

Aryan-Railtown:
## Summary
- Un-gitignore `CLAUDE.md` so it's the shared, git-tracked project file for the team
- Add `.claude/skills/comment-consistency` ,  auto-applies our recurring PR review nitpicks (imports, docstring drift, no bare `Any`, no inline TODOs, etc.) whenever code is edited

## Test plan
- [ ] Confirm `CLAUDE.md` shows up for teammates after a `git pull`
- [ ] Make a trivial code edit and confirm the skill gets invoked automatically

### DIFF
--- a/.claude/skills/comment-consistency/SKILL.md
+++ b/.claude/skills/comment-consistency/SKILL.md
@@ -1,0 +1,28 @@
+---
+name: comment-consistency
+description: >
+  Apply railtracks' code/comment/docstring consistency rules whenever writing or editing Python code in
+  this repo — new functions, modified signatures, docstrings, comments, or error messages. Use this
+  automatically as part of any code change, not just when explicitly asked to clean up comments.
+---
+
+# Comment & code consistency
+
+Before finishing any code change in this repo, check the diff against these rules. Each one comes from a
+recurring, repeated PR review comment on railtracks, not a hypothetical style preference.
+
+- **Imports**: top-level only. No local/inline imports inside functions.
+- **Docstrings**: must match the actual signature, param names, types, `| None` / `Iterable[...]` etc.
+  Update the docstring any time a signature changes.
+- **Module-level comment strings**: a docstring at the top of a file is welcome. Avoid long inline
+  paragraph-style comments in the body of the code — keep inline comments short and targeted.
+- **Typing**: no bare `Any` when a concrete type is available or already imported nearby.
+- **TODOs**: no inline `# TODO` comments. Either fix it now or file a tracked GitHub issue and reference it.
+- **Cruft**: no dead/commented-out code, no stray debug/scratch files left in a change.
+- **Mutable defaults**: don't store a mutable default (`x or []`) by reference and mutate it later — copy
+  defensively.
+- **Errors**: use `repr(e)` rather than `str(e)` in error/log messages.
+- **`__init__.py`**: don't add docstrings to `__init__.py` files — not this repo's convention.
+- **Consistency with siblings**: when adding a new variant of something that already has 2+ siblings
+  (chunkers, loaders, LLM providers, etc.), match their existing shape/behavior rather than introducing an
+  ad hoc one-off.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,151 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Repository Overview
+
+Railtracks (`rt`) is a Python framework for building agentic systems: agents, tools, and multi-step flows
+defined entirely in standard Python (no YAML/DSLs). This is a `uv` workspace monorepo:
+
+```
+Root
+├── packages/railtracks/          # The actual "railtracks" PyPI package
+│   ├── pyproject.toml            # Package dependencies (add new deps here, in optional-dependencies for extras)
+│   ├── src/railtracks/           # Source (module is railtracks, package dir uses underscore)
+│   └── tests/                    # unit_tests/, integration_tests/, end_to_end/, llm_live_tests/
+├── pyproject.toml                # Root workspace: dev-tooling deps only (docs/test/lint groups), NOT package deps
+├── docs/                         # mkdocs documentation source
+├── examples/                     # Example scripts
+└── scripts/                      # CI helper scripts (dependency sorting, license checks, docs validation)
+```
+
+## Ticket Workflow
+
+When working on a ticket/task with the user, follow this sequence:
+
+1. **Branch first**: create a new branch dedicated to the ticket before any other work, and stay on it for
+   the entire ticket unless told otherwise.
+2. **Explore & understand**: read the relevant code, experiment as needed, and scope out what the ticket
+   actually requires before proposing anything. For bug fixes, this step and planning (below) may merge
+   into one.
+3. **Plan**: once the shape of the work is clear, write the plan.
+4. **Record decisions in Notion**: capture the plan and key decisions in a Notion page for this ticket
+   (create one page per ticket, under the scratchpad space, via the Notion MCP) before writing any
+   implementation code.
+5. **Tests first**: if the ticket needs new tests, write them first and confirm they fail for the expected
+   reason.
+6. **Implement**: make the change, keeping existing tests and behavior intact.
+7. **Commit only**: create commits when asked, but leave `git push` to the user — they push code themselves.
+   They may ask for explanations along the way to stay current with what's changed.
+
+Local, gitignored scratch docs (one pair per branch, not committed):
+- `flow.md` — the running log of what changed and how it affects the codebase.
+- `decisions.md` — major decisions made during the ticket, with brief rationale.
+
+**Never publish anything externally visible** — PR creation, PR/issue comments, or any other
+web-visible change — without explicit consultation first, even if a prior similar action was approved.
+Each instance needs its own go-ahead unless the user has said otherwise in advance.
+
+## Setup
+
+```bash
+uv sync --group dev
+uv pip install -e "packages/railtracks[all]"   # or a specific extra, e.g. [visual], [retrieval]
+```
+
+## Common Commands
+
+```bash
+# Lint / format (must pass before commit — CI enforces this)
+ruff check --fix
+ruff format
+
+# Unit tests only (fast, ~10s)
+pytest packages/railtracks/tests/unit_tests/ -v --timeout=30
+
+# Unit + integration tests (excludes llm_live_tests and retrieval e2e, per root pyproject.toml addopts)
+pytest -s -v packages/railtracks/tests/unit_tests/ packages/railtracks/tests/integration_tests/
+
+# Single test file / test
+pytest packages/railtracks/tests/unit_tests/nodes/test_x.py -v
+pytest packages/railtracks/tests/unit_tests/nodes/test_x.py::test_name -v
+
+# Dependency sort check (CI enforced)
+python scripts/check_dependencies_sorted.py
+
+# Docs
+mkdocs serve            # local preview at localhost:8000
+mkdocs build --strict --verbose
+```
+
+CI (`.github/workflows/pr_tests.yaml`) runs, in order: `ruff check . --no-fix`, `ruff format --check .`,
+license check, `check_dependencies_sorted.py`, then the full pytest suite. Run these locally before pushing.
+
+Note: `end_to_end` and `llm_live_tests` require real API keys/network and are excluded from the default
+pytest run via root `pyproject.toml` `addopts`. `RAILTRACKS_TEST_MODE` is auto-enabled during tests
+(via `conftest.py`) to disable session persistence to disk; opt into persistence testing with
+`RAILTRACKS_ALLOW_PERSISTENCE=1` and the `allow_persistence` fixture.
+
+## Architecture
+
+### Core building blocks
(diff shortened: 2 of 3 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,151 @@
...
+   implementation code.
+5. **Tests first**: if the ticket needs new tests, write them first and confirm they fail for the expected
+   reason.
+6. **Implement**: make the change, keeping existing tests and behavior intact.
+7. **Commit only**: create commits when asked, but leave `git push` to the user — they push code themselves.
+   They may ask for explanations along the way to stay current with what's changed.
+
+Local, gitignored scratch docs (one pair per branch, not committed):
+- `flow.md` — the running log of what changed and how it affects the codebase.
+- `decisions.md` — major decisions made during the ticket, with brief rationale.
```
Aryan-Railtown, line 43: i feel like this would be a nice to have per branch, but we should make sure to not merge this into main. What do you guys think?

CLAUDE.md
```diff
@@ -0,0 +1,151 @@
...
+│   ├── pyproject.toml            # Package dependencies (add new deps here, in optional-dependencies for extras)
+│   ├── src/railtracks/           # Source (module is railtracks, package dir uses underscore)
+│   └── tests/                    # unit_tests/, integration_tests/, end_to_end/, llm_live_tests/
+├── pyproject.toml                # Root workspace: dev-tooling deps only (docs/test/lint groups), NOT package deps
+├── docs/                         # mkdocs documentation source
+├── examples/                     # Example scripts
+└── scripts/                      # CI helper scripts (dependency sorting, license checks, docs validation)
+```
+
+## Ticket Workflow
```
Aryan-Railtown, line 22: The workflow for agent might be constrictive, if you guys think this is narrowing it down too much, we can get rid of this section.

Pooria90:
Let's keep this open until 1.5 is released so we can publish a clean one based on the new architecture and changes.

For example these details get deprecated very soon:
```
  | `tool_nodes`? | `output_schema`? | Resulting type |
  |---|---|---|
  | No | No | `TerminalLLM` |
  | No | Yes | `StructuredLLM` |
  | Yes | No | `ToolCallLLM` |
  | Yes | Yes | `StructuredToolCallLLM` |
```

CoronRing:
I want to make sure that dev-focused skill (ie some part of CLAUDE and all of comment) is clearly isolated from user-focused descriptions. While it is somewhat clear that the CLAUDE.md exists as part of the repo and is dev-focused for developers, this may not be the case for beginners. I would propose that this doc be user-focused and have a 1-line hint to use DEV_CLAUDE.md if the agent is developing the framework instead of being a user. 

A user's agent should not need to know the system structure nor follow our convention (ie comments) on their own projects.

.claude/skills/comment-consistency/SKILL.md
```diff
@@ -0,0 +1,28 @@
...
+  this repo — new functions, modified signatures, docstrings, comments, or error messages. Use this
+  automatically as part of any code change, not just when explicitly asked to clean up comments.
+---
+
+# Comment & code consistency
+
+Before finishing any code change in this repo, check the diff against these rules. Each one comes from a
+recurring, repeated PR review comment on railtracks, not a hypothetical style preference.
+
+- **Imports**: top-level only. No local/inline imports inside functions.
```
CoronRing, line 14: - **Imports**: top-level only. No local/inline imports inside functions.

This one is hard due to various import issues and passive imports to avoid expensive repo. In fact, I will create a new ticket thats says all expensive import must happen at runtime, instead of import time, which impacts even feature no one use. (ie litellm is always imported, even if a custom router is used.). I would update this line (some other) to jsut say run lint/ruff, which we require for any PR anyways.

.claude/skills/comment-consistency/SKILL.md
```diff
@@ -0,0 +1,28 @@
...
+- **Docstrings**: must match the actual signature, param names, types, `| None` / `Iterable[...]` etc.
+  Update the docstring any time a signature changes.
+- **Module-level comment strings**: a docstring at the top of a file is welcome. Avoid long inline
+  paragraph-style comments in the body of the code — keep inline comments short and targeted.
+- **Typing**: no bare `Any` when a concrete type is available or already imported nearby.
+- **TODOs**: no inline `# TODO` comments. Either fix it now or file a tracked GitHub issue and reference it.
+- **Cruft**: no dead/commented-out code, no stray debug/scratch files left in a change.
+- **Mutable defaults**: don't store a mutable default (`x or []`) by reference and mutate it later — copy
```
CoronRing, line 22: Some of these item does not seem "comment" related, I would move it to another file or alike. It might be worthy to merge some item from `CONTRIBUTING.md` which is for human but should also apply to AI.

CLAUDE.md
```diff
@@ -0,0 +1,151 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
```
CoronRing, line 3: Lets remove this line and the header (# CLAUDE.md) as this file maybe used by other agents like codes or antigravity, which also read CLAUDE.md because its so common

CLAUDE.md
```diff
@@ -0,0 +1,151 @@
...
+1. **Branch first**: create a new branch dedicated to the ticket before any other work, and stay on it for
+   the entire ticket unless told otherwise.
+2. **Explore & understand**: read the relevant code, experiment as needed, and scope out what the ticket
+   actually requires before proposing anything. For bug fixes, this step and planning (below) may merge
+   into one.
+3. **Plan**: once the shape of the work is clear, write the plan.
+4. **Record decisions in Notion**: capture the plan and key decisions in a Notion page for this ticket
+   (create one page per ticket, under the scratchpad space, via the Notion MCP) before writing any
+   implementation code.
+5. **Tests first**: if the ticket needs new tests, write them first and confirm they fail for the expected
```
CoronRing, line 35: This thes not make full sense to me. I would change this to : when working on a issue, create a reproduce procedure that trigger the error. After the patch ensure the issue is addressed by going through the reproduce code and turn it into a test.

CLAUDE.md
```diff
@@ -0,0 +1,151 @@
...
+   reason.
+6. **Implement**: make the change, keeping existing tests and behavior intact.
+7. **Commit only**: create commits when asked, but leave `git push` to the user — they push code themselves.
```
CoronRing, line 38: Not even commit, stage it as you can no longer diff when you commit, but you can still do it when code is staged

CLAUDE.md
CoronRing: The isntrutcion may not be not super clear, maybe just to me (〒︿〒),. Is this suggesting the agent should create this for the user to see? I think this will make things worse, because agents will start writing to these files without the user knowing and leave changes out of its inline response.

CLAUDE.md
```diff
@@ -0,0 +1,151 @@
...
+- `evaluations/` — evaluators and runners for scoring agent behavior.
+- `observability/` — event model, `Observer` pub/sub, and `writers/` (writer protocol) for run tracing.
+- `retrieval/` — optional RAG stack: `chunking/`, `embedding/`, `loaders/`, `stores/` (install via
+  `railtracks[retrieval]`).
+- `human_in_the_loop/`, `prebuilt/` (ready-made tools/agents), `prompts/`, `validation/` (node
+  creation/invocation checks), `utils/` (logging, serialization, visuals, config), `cli/`.
+- `_session.py` — `Session`/`session()`/`ExecutionInfo`, the top-level run/session lifecycle.
+
+### Things to avoid when writing or modifying agent code
+- Don't write vague tool docstrings — it's the literal tool description the LLM sees.
```
CoronRing, line 136: DRY with the comment rule file + other part of this file like where flow is explained, types, ruff etc

CLAUDE.md
```diff
@@ -0,0 +1,151 @@
...
+- Don't omit type hints on tool functions — they define the JSON schema parameters.
+- Don't create a `Flow` *and* a manual top-level `await rt.call(...)` for the same agent — one entry point.
+- Don't add tools an agent doesn't need.
+
+## Code conventions
+
+See `.claude/skills/comment-consistency/SKILL.md` for this repo's code/comment/docstring conventions —
+it's applied automatically whenever code is written or edited here.
```
CoronRing, line 144: Not sure what this means: (ﾟ∀。)? Auto applied?

.claude/skills/comment-consistency/SKILL.md
Aryan-Railtown: fair point, didnt think about the lazy-import-for-expensive-deps case (litellm etc). lets soften this to just say follow ruff/lint, which is already required for every PR anyway, instead of a blanket no-local-imports rule. feel free to link your new ticket here once you make it

.claude/skills/comment-consistency/SKILL.md
Aryan-Railtown: yeah fair, some of these arent strictly "comment" related (mutable defaults, repr(e) etc), might just rename the skill to something broader like code-consistency instead of splitting into a separate file. agree on pulling relevant bits from CONTRIBUTING.md in too so we're not maintaining the same rules twice

CLAUDE.md
Aryan-Railtown: good call, will drop the header + that line. no reason to call out claude.ai/code specifically since its basically a defacto standard other tools read too

CLAUDE.md
Aryan-Railtown: https://github.com/RailtownAI/railtracks/pull/1385#discussion_r3769230179

i think it would be better to just get rid of this section all together, we can give the agent some space to work without clobbering it too much

CLAUDE.md
Aryan-Railtown: same as above, going to remove the section

CLAUDE.md
Aryan-Railtown: agreed, will just link to one from the other instead of repeating the same info in two places

CLAUDE.md
Aryan-Railtown: this just means claude should auto-invoke the skill whenever it edits code here, based on the skill's description matching...not a hard enforced hook, more like a strong nudge. will reword so its less ambiguous

Aryan-Railtown:
> I want to make sure that dev-focused skill (ie some part of CLAUDE and all of comment) is clearly isolated from user-focused descriptions. While it is somewhat clear that the CLAUDE.md exists as part of the repo and is dev-focused for developers, this may not be the case for beginners. I would propose that this doc be user-focused and have a 1-line hint to use DEV_CLAUDE.md if the agent is developing the framework instead of being a user.
> 
> A user's agent should not need to know the system structure nor follow our convention (ie comments) on their own projects.


this is repo-root `CLAUDE.md`, not shipped anywhere - checked pyproject.toml, sdist only includes docs/README/LICENSE, so users never get this file at all. and the skill lives at .claude/skills/comment-consistency here in the repo, completely separate from packages/railtracks/src/railtracks/cli/skills/ which is what railtracks add claude:<skill> actually bundles for users. so there's no real overlap here, dont think we need a DEV_CLAUDE.md for this one

.claude/skills/comment-consistency/SKILL.md
```diff
@@ -0,0 +1,28 @@
...
+
+Before finishing any code change in this repo, check the diff against these rules. Each one comes from a
+recurring, repeated PR review comment on railtracks, not a hypothetical style preference.
+
+- **Imports**: top-level only. No local/inline imports inside functions.
+- **Docstrings**: must match the actual signature, param names, types, `| None` / `Iterable[...]` etc.
+  Update the docstring any time a signature changes.
+- **Module-level comment strings**: a docstring at the top of a file is welcome. Avoid long inline
+  paragraph-style comments in the body of the code — keep inline comments short and targeted.
+- **Typing**: no bare `Any` when a concrete type is available or already imported nearby.
```
soulFood5632, line 19: we may actually want to be even more stringent than that and say "don't throw Any" in to fix something that is broken.

.claude/skills/comment-consistency/SKILL.md
soulFood5632: Yea I like that. You could name it something like code-style or something along those lines.

CLAUDE.md
```diff
@@ -0,0 +1,151 @@
...
+license check, `check_dependencies_sorted.py`, then the full pytest suite. Run these locally before pushing.
+
+Note: `end_to_end` and `llm_live_tests` require real API keys/network and are excluded from the default
+pytest run via root `pyproject.toml` `addopts`. `RAILTRACKS_TEST_MODE` is auto-enabled during tests
+(via `conftest.py`) to disable session persistence to disk; opt into persistence testing with
+`RAILTRACKS_ALLOW_PERSISTENCE=1` and the `allow_persistence` fixture.
+
+## Architecture
+
+### Core building blocks
```
soulFood5632, line 91: This we want to defer to the agent-builder skill instead of having it in 2 locations.

soulFood5632 (changes requested):
Overall looks good to me. Lots of good comments from the rest of the team as well. I love the setup.

Aryan-Railtown:
Closing — rebuilding this on a fresh branch off main now that the mega-merge has landed. Will open a new PR once ready.

PR #1386, opened 2026-08-12
fix: only fill bare {key} placeholders 

CoronRing:
Restricts prompt placeholder filling to bare `{key}` lookups, and adds
`rt.escape_braces` for text that should be delivered to the model as written.

`string.Formatter().parse()` is now used as a lexer only, and `fill_template` does the
substitution itself against a bare-key allowlist.

## Behaviour change

- `{a.b}`, `{a[b]}`, `{a!r}` and `{a:spec}` are no longer resolved. They render exactly as
  written and are logged at WARNING.
- Bare `{key}`, the missing-key echo-back, numeric keys like `{1}`, and `{{` / `}}` are
  unchanged. The existing tests in `test_prompt.py` pass without modification.
- Filling no longer raises. Unbalanced braces and unresolvable placeholders pass through
  instead of failing the run.

## `rt.escape_braces`

Lets one message hold both a template you wrote and a string you did not:

```python
prompt = f"The current time is {{time}}:\nUser Message:\n{rt.escape_braces(user_text)}"
```

`{time}` is filled from the context; everything inside `user_text` is delivered as written.
Escaping survives exactly one render, so it is not idempotent.

## Verification

- `1879 passed, 3 skipped` on the unit suite. The remaining errors are pre-existing missing
  optional extras (`sqlalchemy`, http-fetch, tavily, websearch, huggingface).
- `ruff format --check` and `ruff check`: clean.

Rationale and follow-up are tracked internally.

### DIFF
--- a/docs/tutorials/concepts/prompts_and_context.md
+++ b/docs/tutorials/concepts/prompts_and_context.md
@@ -23,6 +23,24 @@ Passing prompt details up the chain can be expensive in both **tokens** and **la
 2. Set values in the Railtracks context (see [Context Management](../../documentation/advanced/context.md) for details)
 3. When the prompt is processed, the placeholders are replaced with the corresponding values from the context
 
+Only a placeholder naming a context key on its own is replaced. A placeholder that reaches into a
+value, such as `{config.host}` or `{config[host]}`, is left in the prompt as written, as is a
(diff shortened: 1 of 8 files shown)


soulFood5632 (approved):
LGTM

Amir-R25 (approved):
Thanks Guan!

PR #1389, opened 2026-08-12
Pooria/warnings before 1 5

Pooria90:
## Summary

Adds `FutureWarning` notices to the public APIs that 1.5.0 removes or relocates, so users
on 1.4.7 get in-band warning before the breaking release.

Notices are emitted as `FutureWarning` rather than `DeprecationWarning`: the latter is
hidden by default outside `__main__`, so most users would never see it.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

## Notes

**Nothing breaks in this PR.** It only adds notices ahead of 1.5.0. The migration guide is
`docs/documentation/upgrading/1_5_0.md`, surfaced via a site-wide announcement bar rather
than a nav entry.

**Two design rules the implementation follows**, worth knowing when reviewing:

1. *A name that gains a notice must leave `__all__`.* pdoc resolves every `__all__` entry;
   a lazily-provided name yields `Found 'X' in __all__, but it does not resolve` and
   silently drops out of the API reference. Removed names stay in `__dir__` (REPL
   completion) with redundant `TYPE_CHECKING` aliases (type-checker resolution).
2. *Notices go on the documented path only.* `rt.guardrails.Guard` warns;
   `railtracks.guardrails.core` does not, because that's what railtracks' own code imports:
   a notice there would fire during `import railtracks`. Same reason the `input`/`output`
   guard notices live in their own `__init__` files rather than the parent package.

**Test filter.** `pyproject.toml` makes these notices *errors* in the suite, matched on
**message** (`error:.*in railtracks 1\.5\.0:FutureWarning`) rather than category. Matching
on `FutureWarning` alone would fail CI on any dependency's warning.

### DIFF
--- a/packages/railtracks/src/railtracks/built_nodes/__init__.py
+++ b/packages/railtracks/src/railtracks/built_nodes/__init__.py
@@ -1,0 +1,14 @@
+"""Node builders: Forward path
+
+``agent_node`` and ``function_node`` are exported from here. They previously lived in
+``railtracks.built_nodes.easy_usage_wrappers``, which is removed in railtracks 1.5.0.
+"""
+
+__all__ = [
+    "agent_node",
+    "function_node",
+]
+
+
+from .easy_usage_wrappers.agent import agent_node
+from .easy_usage_wrappers.function import function_node
--- a/packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
+++ b/packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
@@ -3,8 +3,7 @@ from typing import Callable, Iterable, Literal, Type, TypeVar, overload
 
 from pydantic import BaseModel
 
-from railtracks.built_nodes.concrete.function_base import RTFunction
-from railtracks.built_nodes.concrete.guarded_llm import (
+from railtracks.built_nodes.concrete import (
     GuardedStreamingStructuredLLM,
     GuardedStreamingTerminalLLM,
     GuardedStreamingToolCallLLM,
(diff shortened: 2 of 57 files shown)


packages/railtracks/src/railtracks/built_nodes/easy_usage_wrappers/agent.py
```diff
@@ -318,10 +319,23 @@ def agent_node(
         tool_nodes (set[Type[Node] | Callable | RTFunction] | None): If your agent is a LLM with access to tools, what does it have access to?
         output_schema (Type[_TBaseModel] | None): If your agent should return a structured output, what is the output_schema?
         llm (ModelBase): The LLM model to use. If None it will need to be passed in at instance time.
+            Deferring the model this way is going away: `llm` becomes a required keyword in
+            railtracks 1.5.0.
         system_message (SystemMessage | str | None): System message for the agent.
         manifest (ToolManifest | None): If you want to use this as a tool in other agents you can pass in a ToolManifest.
         guardrails (Guard | None): Guardrail config. When provided, the agent runs input/output guardrails.
     """
+    if llm is None:
```
Amir-R25, line 339: love it, thanks for being thorough

packages/railtracks/src/railtracks/built_nodes/__init__.py
```diff
@@ -0,0 +1,14 @@
+"""Node builders: Forward path
```
Amir-R25, line 1: I don't think this is necessary? Users were calling `railtracks.agent_node` and `railtracks.function_node` so they wouldn't care about this warning, we can skip this one?

Amir-R25 (commented):
Thanks!

I think we might be missing the "StructuredToolCallLLM" deprecation? cc @soulFood5632

soulFood5632 (dismissed):
LGTM

packages/railtracks/src/railtracks/built_nodes/__init__.py
Pooria90: Just a precaution; mostly for contributors not users.

Amir-R25 (approved):
image

Issue #1267, opened 2026-07-17
General HIL structure (example workflows)


soulFood5632:
### Revised scope proposal (2026-08-25)

Originally scoped -> "ship a suite of concrete verifier backends": Python input / webhook / custom (original description preserved below). 

Revised scope-> **give people the tools to build HIL themselves**, rather than ship one opinionated HIL system.

Concretely:
- `pre_verifier` + `post_verifier` as prebuilt middleware (today's `verifier`, renamed/re-homed as `pre_verifier` with no behavior change, plus a new `post_verifier` that gates a call's *output* after it's already run). Tracked as a sub-issue of #1266: #1485.
- Maybe a prebuilt "looping HIL" node/pattern for continuously-running human-gated loops (e.g. a chat that keeps going until a human confirms ending it). 
- Concrete backends (CLI/local-input, webhook, LLM-mediated, custom/composable) stay as **documented examples and demos**, not built-in features. Matches what we already found: any callable satisfying `Callable[P, Verdict]` already works with `verifier` as-is, so "custom" isn't something to build, that generalizes to the other backends too.

NOTE: Working demos for all of the above are in `examples/human_in_the_loop/` on branch `1267-general-hil-structure`, including a non-runnable API mockup (`proposal_verifier.py`) for the pre/post split. Full notes on the "General HIL structure (#1267)" Notion page under "Human in the loop".

---

### Original description

A suite of tools that allow you create verifier functions that take in some sort of input and output some idea confirm, confirm with comments, decline, and decline with comments.

some concrete ways to implement these verifier functions will include:
- Python's input
- webhook
- custom

some of those TBD

> Issue is not complete please talk to @soulFood5632

soulFood5632:
LGTM

Issue #1266, opened 2026-07-17
General Verifier


soulFood5632:
### Description

In reference to the parent #1265:

Implement a verifier middleware which takes in a `callable[P, Bool]` where P is the input type of the node. This function will run every time the node is called and will raise special exception when returned false. 

Note although not fully fleshed out you could return something like accept, accept with comments, decline, and decline with comments

Issue #1265, opened 2026-07-17
Implement Naive Human in the loop


soulFood5632:
### Description

We should add HIL (human-in-the-loop) support as `@wrap_node` node middleware.

There are two possible implementations below. Only Option 1 should be implemented unless a coherent argument for Option 2 is made.

The approach is to first design a general verifier, then build a specific module on top of it to support HIL. Each will be broken into sub-issues.

### Option 1: Node Verification (node-level middleware)

Build one middleware that gates the input to a node with a custom function of type `Callable[_P, bool]`, where `_P` matches the node's own input signature. This general verifier can later be specialized into a HIL-specific implementation.

### Option 2: Tool Call Verification (LLM node-level config parameter)

This could instead be implemented by modifying the LLM node's tool-calling logic to trigger a verification flow before a tool call is dispatched.

**Suggestion Option 1**: this approach makes verification a property of the agent that calls tools, rather than a property of the node itself. Option 1 keeps verification composable and swappable via middleware, independent of any specific node's internals.

PR #1401, opened 2026-08-17
Add general verifier middleware (#1266)

Aryan-Railtown:
## Summary
- Implements the general verifier from #1266 (sub-issue of #1265, Naive Human in the Loop) as an ordinary `wrap_node` middleware -- no new execution mechanism needed.
- `Verdict` (accept / accept-with-comments / decline / decline-with-comments, via `accepted: bool` + optional `comment`/`args`/`kwargs` override) lives in its own module (`middleware/verdict.py`) for a cleaner split from the middleware wrapper itself.
- `verifier(approve_fn, *, timeout=None, name=None)` binds `approve_fn` to the wrapped node's own args/kwargs (sync or async), raises `VerifierRejectedError` on decline before the node's `invoke` ever runs, and treats a timeout as a decline with reason `"timeout"`.
- Exported as `rt.verifier` / `rt.middleware.{Verdict, VerifierRejectedError}`.
- Includes a small unrelated fix: `test_state.py`'s `_create_node_and_request` regression test (added on `feature-branch-node-add-on`, predates this work) still asserted the old single-value return; updated it to match main's intentional `(request_id, node_instance)` tuple contract, surfaced while merging main in.
- HIL itself (#1267, pluggable ask-a-human backends) is a separate follow-up ticket -- explicitly out of scope here, and marked incomplete pending a design conversation with @soulFood5632.

## Test plan
- [x] `packages/railtracks/tests/unit_tests/middlewares/test_verifier.py` -- 12 new tests (unit-level against `verifier(...).wrap(core)`, plus end-to-end through the real `rt.call` path)
- [x] Full unit suite: 2260 passed, 4 skipped
- [x] `ruff check .` / `ruff format --check .` clean


soulFood5632 (approved):
LGTM, amazing work. We will want to add a pair of future work items. (Docs and how to send these as specific events)

Aryan-Railtown:
> LGTM, amazing work. We will want to add a pair of future work items. (Docs and how to send these as specific events)

Yes, we can make separate issues for them.

PR #1392, opened 2026-08-13
Logan great merge mid summersnight dream

soulFood5632:
## Summary

The big mega merge of `feature-branch-node-add-on` and `feature-branch-observability-update`

There were a couple of merging shenanigans worth paying attention to:
1. Streaming had to be fit to work within the new scope link structure. It used many APIs which were removed. 

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs
- [ ] Refactor / chore / build / tests

## Checklist

- [ ] Lint & format pass (`ruff check . && ruff format .`)
- [ ] Tests added/updated and pass locally (`pytest tests`)
- [ ] Docs updated if user-facing behavior changed
- [ ] Breaking changes include migration notes

PR #1406, opened 2026-08-17
Fix 1405 mcp close loop race

### LINKED ISSUE #1405: [Bug] `MCPServer.close()` races the background thread's `loop.close()`, masking setup errors with `RuntimeError: Event loop is closed`
## Summary

When MCP server setup fails, the background thread signals shutdown, then tears down
its own event loop. The main thread, woken by the same signal, calls `self.close()`,
which calls `call_soon_threadsafe` on that loop. Nothing orders these two, so if the
background thread closes the loop first, `close()` raises
`RuntimeError: Event loop is closed` and the real setup error `FileNotFoundError` dorps.

## Where observed

`tests/unit_tests/rt_mcp/test_error_handling.py::TestErrorContextAndMessages::test_file_not_found_has_helpful_message`
failed on PR #1396 (an unrelated change to `llm/tools`, which merely merged `main` in):

<https://github.com/RailtownAI/railtracks/actions/runs/31928503850/job/95119836667>

```
packages/railtracks/tests/unit_tests/rt_mcp/test_error_handling.py:59: in test_file_not_found_has_helpful_message
    MCPServer(
packages/railtracks/src/railtracks/rt_mcp/main.py:306: in __init__
    self.close()
packages/railtracks/src/railtracks/rt_mcp/main.py:392: in close
    self._loop.call_soon_threadsafe(self._shutdown_event.set)
.../asyncio/base_events.py:798: in call_soon_threadsafe
    self._check_closed()
E   RuntimeError: Event loop is closed

self = <_UnixSelectorEventLoop running=False closed=True debug=False>
```

`main`'s own run of the same commit passed, so this presents as a flaky test.

## Repro

Deterministic. The `sleep` only *chooses* the interleaving that CI hit by chance —
it does not create the fault.

```python
import time

from railtracks.rt_mcp import MCPStdioParams
from railtracks.rt_mcp.main import MCPServer

orig_close = MCPServer.close


def slow_close(self, *a, **kw):
    time.sleep(0.3)  # let the background thread reach self._loop.close() first
    return orig_close(self, *a, **kw)


MCPServer.close = slow_close

try:
    MCPServer(config=MCPStdioParams(command="missing_mcp_server", args=[]), setup_timeout=5)
except FileNotFoundError:
    print("FileNotFoundError (expected)")
except RuntimeError as e:
    print(f"RuntimeError: {e}")
```

### Actual

```
RuntimeError: Event loop is closed
```

### Expected

```
FileNotFoundError (expected)
```

i.e. `FileNotFoundError: MCP server command not found: 'missing_mcp_server'. Please
verify the command is installed and in your PATH.`

CoronRing:
## Summary

Closes #1405.

`MCPServer.close()` signals shutdown via `self._loop.call_soon_threadsafe(...)`. On a
setup failure the background thread has already signalled itself and closed that loop,
so the call raises `RuntimeError: Event loop is closed` which replaces the real
`FileNotFoundError`/`TimeoutError` the caller needs. Wrapping the signal in
`except RuntimeError` lets the genuine setup error propagate.

## Why this is a one-line change (and why it is in `close()`)

The bug is not that `close()` is called at the wrong time it is that `close()` is not
safe to call once the background thread has disposed of the loop, which its own
docstring already promises ("Safe to call even if initialization failed early").

### Two alternatives were considered and rejected:

- **Reverting the `self.close()` call at line 306** (added by #1392)  the timeout branch
  at line 297 calls `close()` the same way against the same possibly-closed loop, so this
  would leave the defect reachable. Fixing `close()` covers both callers.
- **Guarding with `if not self._loop.is_closed()`**  check-then-act. The loop can close
  between the check and the call, leaving the same window open. The `except` has to wrap
  the call itself.
## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [x] Lint & format pass (`ruff check packages/ && ruff format --check packages/`)
- [x] Docs updated if user-facing behavior changed (n/a — `close()` now matches its
      documented contract; no signature or behaviour change on the success path)
- [x] Breaking changes include migration notes (n/a)
.

### DIFF
--- a/packages/railtracks/src/railtracks/rt_mcp/main.py
+++ b/packages/railtracks/src/railtracks/rt_mcp/main.py
@@ -389,7 +389,11 @@ class MCPServer:
                 turns that into a logged warning instead of an indefinite hang.
         """
         if self._loop is not None and self._shutdown_event is not None:
-            self._loop.call_soon_threadsafe(self._shutdown_event.set)
+            try:
+                self._loop.call_soon_threadsafe(self._shutdown_event.set)
+            except RuntimeError:
+                # The background thread already signalled itself and closed the loop
+                pass
             self._thread.join(timeout=join_timeout)
             if self._thread.is_alive():
                 logger.warning(
--- a/packages/railtracks/tests/unit_tests/rt_mcp/test_error_handling.py
+++ b/packages/railtracks/tests/unit_tests/rt_mcp/test_error_handling.py
@@ -8,6 +8,7 @@ including invalid commands, timeouts, and exception propagation from background 
 
 import threading
 import time
+from unittest.mock import patch
 
 import pytest
 from railtracks.rt_mcp import MCPStdioParams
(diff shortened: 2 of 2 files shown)


soulFood5632 (approved):
LGTM

PR #1419, opened 2026-08-18
feat: add timeout middleware

### LINKED ISSUE #1411: Timeout Middleware
### Description

A prebuilt middleware that can be used to wrap a node in a timeout so it will fail instead of halting after a long operation. 

The DX could look something like the following:
```python
agent_node(..., middleware=[timeout_mw(time=10)])
```

yibiao42-cloud:
## Summary

- Add a slot-agnostic `Timeout` middleware that cancels calls exceeding the configured deadline.
- Export `Timeout` from `railtracks.prebuilt.middleware`.
- Add unit tests covering successful calls, argument forwarding, cancellation, and exception propagation.
- Add a runnable documentation example and catalog entry.

## Testing

- `pytest packages/railtracks/tests/unit_tests/middlewares -q` — 56 passed
- `ruff check packages/railtracks/src/railtracks/prebuilt/middleware`
- `python -m py_compile docs/scripts/prebuilt_middleware.py`

Closes #1411


soulFood5632 (dismissed):
Thanks for the contribution looks great on my end :)

Amir-R25:
Thanks @yibiao42-cloud, could you please update from base branch when you get a chance

packages/railtracks/src/railtracks/prebuilt/middleware/timeout.py
```diff
@@ -0,0 +1,23 @@
...
+
+    Args:
+        seconds: Maximum number of seconds to wait for the wrapped call.
+    """
+
+    def __init__(self, seconds: float):
+        self._seconds = seconds
+        super().__init__(self._middleware_fn)
+
+    async def _middleware_fn(self, call, *args, **kwargs):
```
CoronRing: Note `asyncio.exceptions.TimeoutError` is its own class deriving from Exception: it is not the builtin TimeoutError (which derives from `OSErro`r). The alias only landed in 3.11. The middleware raises the asyncio one while the test catches the builtin one, so nothing catches it and it escapes` pytest.raises`. 

If you python ==3.10 you can reproduce this. 
Hope that helps, and great work!

Ciao =D

soulFood5632 (approved):
Great work!

PR #1423, opened 2026-08-18
Removed Unneeded Function Wrap in Quick Start

### LINKED ISSUE #1261: Docs Wrap an Agent in a Function for no Reason
### What's on your mind?

in our docs the block of code wraps the agent in a function for no major reason. 

We should just get rid of the `function_node` instead
```
import railtracks as rt

# To create your agent, you just need a model and a system message. 
Agent = rt.agent_node(
    system_message="You are a helpful AI assistant."
)

# Create a function node that will be the entry point of our flow. 
# This is where we will call our Agent.
@rt.function_node
async def main(message: str):

    # Now to call the Agent, we just need to use the `rt.call` function
    result = await rt.call(
        Agent,
        message,
    )
    return result

# Create your flow and set the entry point to the function we just created. 
# Then we can invoke the flow with a the input to the function node. 
flow = rt.Flow("Quickstart Example", entry_point=main)

result = flow.invoke("Hello, what can you do?")
```

### Additional Context

_No response_

soulFood5632:
## Summary

See title

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [x] Docs

### DIFF
--- a/docs/scripts/documentation/quickstart.py
+++ b/docs/scripts/documentation/quickstart.py
@@ -7,21 +7,10 @@ Agent = rt.agent_node(
     system_message="You are a helpful AI assistant."
 )
 
-# Create a function node that will be the entry point of our flow. 
-# This is where we will call our Agent.
-@rt.function_node
-async def main(message: str):
-
-    # Now to call the Agent, we just need to use the `rt.call` function
-    result = await rt.call(
-        Agent,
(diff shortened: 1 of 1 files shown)


Aryan-Railtown (approved):
PR number 7 for today letss gooo

PR #1441, opened 2026-08-18
1438 cleanup inject prompt leftover

### LINKED ISSUE #1438: Clean up inject_prompt leftover.
### Description

Now that the inject_prompt functionality is shifted from hardcoded behaviour to middleware as optional importable, we want to clean up existing items, such as not-functioning inject_prompt at message level + doc and prompt_inject at flow level.

CoronRing:
## Summary

Cleans up the leftovers from the context-injection migration. `closes #1438`

Injection used to be hardcoded on for every agent, so a set of switches existed to turn it
back off. Now that it is opt-in through `ContextInjection` middleware, those switches are
redundant — and strictly weaker than the middleware, since every one of them could only
ever say *no*:

- **Message level** — `inject_prompt` on `Message` / `UserMessage` / `SystemMessage` /
  `AssistantMessage`, plus the property and setter. `inject_values` no longer gates on the
  flag or clears it after filling.
- **Flow level** — `prompt_injection` on `rt.Flow(...)` and its pass-through into `Session`.
- **Global / session level** — `prompt_injection` on `rt.set_config`, `rt.Session`,
  `@session`, and `ExecutorConfig`, plus the gate it fed in `inject_context`. Setting it
  `True` was a no-op in every configuration (it is the default and cannot enable anything);
  only `False` did something, and only for an agent that had already opted in via
  middleware.

`ContextInjection()` in `model_middleware` is now the single switch. To disable injection,
omit the middleware.

Also fixes a broken instruction in the prompts walkthrough: it told users to add
`rt.middleware.ContextInjection()`, which does not exist and raises `AttributeError`. The
class is at `rt.prebuilt.middleware.ContextInjection`. The walkthrough now uses the correct
path and links the middleware's own reference page, which is otherwise only reachable from
the prebuilt catalog.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Breaking change
- [x] Docs

## Checklist

- [x] Breaking changes include migration notes

## Notes

**No deprecation shim, by choice.** Intercepting a removed keyword means accepting it, which
would add a `**kwargs` catch-all to five config entry points and four message constructors —
permanently vaguer signatures, and weaker static checking, to carry a notice that expires at
1.5.0. The interface stays clean instead: passing the old arguments raises Python's own
`TypeError`, which already names both the argument and the call site.

```
TypeError: set_config() got an unexpected keyword argument 'prompt_injection'
TypeError: UserMessage.__init__() got an unexpected keyword argument 'inject_prompt'
```

**Migration.** Add `ContextInjection()` to the agents that need placeholder filling:

```python
from railtracks.prebuilt.middleware import ContextInjection

Agent = rt.agent_node(
    name="my-agent",
    system_message="You are helping {user_name}.",
    model_middleware=[ContextInjection()],
)
```

Anything that passed `prompt_injection=False` or `inject_prompt=False` should drop the
argument  the effect is achieved by leaving the middleware off that agent.

**Blast radius.** `prompt_injection=False` only ever mattered for an agent that *also* had
`ContextInjection` in its middleware list, and that middleware only exists as of the
migration commit (`5e6454d0`). Code combining the two would be both new and
self-contradictory, so real usage should be close to none.

**Behaviour check.** Removing the gate also removed a `try/except ContextError` that made
`inject_context` a no-op outside a `Session`. Verified this is still safe: `fill_template`
already swallows the failed lookup, so placeholders come back written as-is rather than
raising.

### DIFF
--- a/.github/ISSUE_TEMPLATE/general.yaml
+++ b/.github/ISSUE_TEMPLATE/general.yaml
@@ -12,11 +12,3 @@ body:
     validations:
       required: true
 
-  - type: textarea
-    id: additional_context
-    attributes:
-      label: Additional Context
-      description: Any other information that will help us understand your query, concern, or suggestion.
-      placeholder: Related documentation, links, or examples.
-    validations:
-      required: false
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -10,6 +10,7 @@ jobs:
   end_to_end_tests:
     name: End to End Tests
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v5
 
--- a/.github/workflows/pr_tests.yaml
+++ b/.github/workflows/pr_tests.yaml
@@ -16,6 +16,7 @@ jobs:
   ruff-lint:
     name: Check Code Formatting and Linting with Ruff
     runs-on: ubuntu-latest
+    timeout-minutes: 15
 
     steps:
       - name: Checkout Code
@@ -38,6 +39,7 @@ jobs:
   check-licenses:
     name: Check Licenses
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Checkout Code
         uses: actions/checkout@v5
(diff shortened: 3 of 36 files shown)


Aryan-Railtown (approved):
LGTM

PR #1442, opened 2026-08-18
fix: stop hardcoding openai/ prefix in OpenAICompatibleProvider (#1437)

### LINKED ISSUE #1437: We are hardcoding f"openai/ for a non-openai model
### Description

In `class OpenAICompatibleProvider(ProviderLLMWrapper, ABC):`, we currently hardcode `def full_model_name(self, model_name: str) -> str:
        return f"openai/{model_name}"`

This is incorrect, right now litellm strips this, but it records, visalizes wrong. the patern itself is also incorrect. 

Observe `OpenAICompatibleProvider`'s full_model_name function

Image

Aryan-Railtown:
## What
`OpenAICompatibleProvider.full_model_name` baked a fake `openai/` prefix into the configured model name so litellm could route calls to gateway-style endpoints (Portkey, Telus, custom deployments). That mutated name is what
`model.model_name()` reports for telemetry, and what visualization falls back to before a real response lands, so every gateway-routed model showed up as `openai/<model>` regardless of what was actually behind the gateway.

## Fix
`full_model_name` is now identity. The `openai` routing hint litellm needs is passed separately as a `custom_llm_provider` completion kwarg instead (routes identically on the wire, verified against litellm 1.89.0) — so `self._model_name`
stays the true configured name everywhere it's read.

## Tests
Added coverage for `OpenAICompatibleProvider` and `PortKeyLLM`/`TelusLLM` model name + routing behavior. 

**Note**: `TelusLLM` had zero test coverage before this, those tests are incidental coverage added alongside the fix, not themselves part of the bug.

Fixes #1437

### DIFF
--- a/packages/railtracks/src/railtracks/llm/models/api_providers/_openai_compatable_provider_wrapper.py
+++ b/packages/railtracks/src/railtracks/llm/models/api_providers/_openai_compatable_provider_wrapper.py
@@ -38,14 +38,9 @@ class OpenAICompatibleProvider(ProviderLLMWrapper, ABC):
             every hyperparameter, valid or not, is passed straight through and any
             error surfaces from the gateway or upstream provider directly.
         """
-        # litellm's own routing (get_llm_provider) hard-fails on model names it
-        # doesn't recognize unless told what protocol to use, and gateway model
-        # names (Portkey, Telus, arbitrary custom deployments) are never in its
-        # catalog. custom_llm_provider is that hint, forced here (not left to
-        # kwargs) since "openai" is the entire premise of this wrapper class.
-        # Passed as a completion kwarg rather than baked into the model name
-        # string, so self._model_name stays the true model name for telemetry
-        # and visualization (see #1437).
+        # litellm needs to be told this is an OpenAI-compatible endpoint to route the
+        # call at all. Forced as a kwarg here instead of baked into the model name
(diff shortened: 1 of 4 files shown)


soulFood5632 (dismissed):
LGTM

packages/railtracks/src/railtracks/llm/models/api_providers/_openai_compatable_provider_wrapper.py
```diff
@@ -38,6 +38,10 @@ def __init__(
             every hyperparameter, valid or not, is passed straight through and any
             error surfaces from the gateway or upstream provider directly.
         """
+        # litellm needs to be told this is an OpenAI-compatible endpoint to route the
+        # call at all. Forced as a kwarg here instead of baked into the model name
+        # string, so the true model name survives for telemetry/visualization (#1437).
```
CoronRing: Lets not mention error ID here. Just for consistency. I am somewhat ok with docstring on test as no one read them anyways ()

Aryan-Railtown: removed!!

CoronRing:
LGTM!

CoronRing (approved):

PR #1444, opened 2026-08-19
fix: metric Result violates LSP

### LINKED ISSUE #1015: Metric Result violates LSP
In the above comment in a PR there was an issue with the `MetricResult` inheritance chain. Our types shoould probably follow LSP when they can

See Pydantic supports that but are we sure we want to design it that way. I feel like usually a geneirc pattern may be better. That way we don't violate LSP.

_Originally posted by @soulFood5632 in https://github.com/RailtownAI/railtracks/pull/1001#discussion_r2927147705_

Mr-Neutr0n:
Fixes #1015

Made MetricResult generic over its value type (Generic[TMetricValue]) and parameterized ToolMetricResult as MetricResult[float | int] and LLMMetricResult as MetricResult[float | int | None], removing the type-narrowing value override that violated LSP. (Issue text was external data; the suggested generic pattern was adopted as the fix.)

Could not run the suite locally: . Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.


soulFood5632 (approved):
Thanks for the contribution, LGTM.

packages/railtracks/src/railtracks/evaluations/result/metric_results.py
```diff
@@ -1,25 +1,27 @@
+from typing import Generic, TypeVar
 from uuid import UUID, uuid4
 
 from pydantic import BaseModel, Field
 
+TMetricValue = TypeVar("TMetricValue")
```
soulFood5632: This metric value should be bound to the original requirements. `float | int | str | None`

soulFood5632 (changes requested):
Sorry upon further review there is a minor regression I noticed.

Mr-Neutr0n:
Addressed the review: bounded MetricResult's generic value type to the original supported set, float | int | str | None, so specialized results cannot widen it accidentally. The focused evaluation tests passed (39 tests), Ruff passed, and mypy reports no issues. Commit: https://github.com/Mr-Neutr0n/railtracks/commit/f9c6f4e61b5fa72f9756178b039c0bb891e7a8bf

soulFood5632:
Just waiting on the release of 1.5.0 before we merge this in

soulFood5632 (approved):
LGTM

Issue #1343, opened 2026-08-05
Thorough Review of Coding Assistant support


soulFood5632:
### Description

Right now we have support of agent building through some very simple skills. I would like to get a sense of how we can improve the performance of these coding agents when using Railtracks. 

This issue should cover the following key tasks 
- Review similar agentic frameworks and understand the support they provide
- Complete an analysis of our support and identify any issues we currently have
- Implement the new changes (which could include new skills etc.)

CoronRing:
@Pooria90 If you can add one line to your next skill PR to include some variant of: 

```When working with docs like .md, do not hard line wrap. Keep each line a paragraph for the quality of reading and editability.```

Will be wasteful if I open a PR just for this. =D

Pooria90:
@CoronRing Sure, thanks for the note.

PR #1450, opened 2026-08-19
fix copilot and stuff

### LINKED ISSUE #1448: Copilot instructions need an update
### Description

The current `.github/copilot-instructions.md` is now stale because of the changes in the framework. Resolving this issue needs a polishing and update to this file.

Pooria90:
## Summary

Repairs `.github/copilot-instructions.md`, which had drifted badly: it documented `packages/railtracks-cli/` (deleted), told contributors to install from a `requirements-dev.txt` that doesn't exist, and shipped a validation snippet calling `rt.call_sync`, which isn't in the public API.

Rather than hand-fix a file that had already rotted once, the framework-concepts half is now **generated** from the bundled
`agent-builder` skill via `railtracks add copilot:agent-builder --force`, so it can't
fork again; the contributor half was rewritten against the live repo and CI config.

Making that generation viable required two fixes to the copilot install handler, and
turned up a correction to the `agent-builder` skill itself (the four named agent classes
it documented were removed in 1.5.0). Also adds `railtracks add --list`.

Part of the coding-assistant review in #1343.

Closes #1448 .

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

## Notes

**`railtracks add --list`** prints every bundled skill with its argument hint and
description, plus the supported tools.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -210,13 +210,14 @@ Steps 3 and 4 are gated on changed paths, so a docs-only PR skips the test suite
 - **Flows** wrap an agent or async function as the entry point and handle execution, config, and context.
 - **`rt.call()`** is used inside async workflows to call agents or nodes directly.
 
-### Agent Type Selection
-| Has `tool_nodes`? | Has `output_schema`? | Agent type |
-|---|---|---|
-| No | No | `TerminalLLM` — plain chat |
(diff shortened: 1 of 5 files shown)


soulFood5632 (approved):
I feel like we should just have a single file which is shared but this is a fine change.

PR #1452, opened 2026-08-20
fix: [Feature] Add Support for non-google docstring formats

### LINKED ISSUE #881: [Feature] Add Support for non-google docstring formats
### Feature Request

Right now we only support google doc strings. it would be valuable to support other dcostring formats in function_node automatic parsing

### Additional Context

_No response_

Mr-Neutr0n:
Fixes #881

Added NumPy and reST/Sphinx docstring format support to the docstring parser used by function_node/Tool automatic parsing, plus tests, with no new dependencies.

Could not run the suite locally: . Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.


packages/railtracks/src/railtracks/llm/tools/docstring_parser.py
```diff
@@ -35,13 +36,24 @@ def parse_docstring_args(docstring: str) -> Dict[str, str]:
...
 
-    # Parse the arguments and their descriptions
-    return parse_args_section(args_section)
+    # NumPy style
+    numpy_section = extract_numpy_args_section(docstring)
+    if numpy_section:
+        arg_descriptions.update(parse_numpy_args_section(numpy_section))
+
+    # reST/Sphinx style
+    rest_args = parse_rest_args_section(docstring)
```
CoronRing, line 54: runs all three parsers unconditionally over the same docstring and blind-merges them seem very inefficient, please explain design decision. 

This is various cases where the system fails entirely due to key words, ie inside a NumPy description	use ---, the whole section lost, in a Args, a key called Parameters exist, etc.

reST descriptions never get stripped from the tool detail. The PR adds reST arg parsing but not reST section awareness in extract_main_description, so a pure-reST function ships this to the model as its tool description:

'Fetches a URL.\n\n    :param url: The endpoint to hit.\n    :type url: str\n    :returns: The body.'

packages/railtracks/src/railtracks/llm/tools/docstring_parser.py
```diff
@@ -126,6 +138,145 @@ def parse_args_section(args_section: str) -> Dict[str, str]:
     return arg_descriptions
 
 
+def _is_numpy_section_header(lines: list, i: int) -> bool:
```
CoronRing, line 160: A quick run by claude returns: 

One style per docstring, first match wins (Google → NumPy → reST), with a UserWarning on mixed styles. Google keeping priority means zero behavior change for existing users.
Indentation-driven NumPy parsing — definitions at the section's base indent, descriptions below. Fixes 3, 4, 5, 6, 11. Handles *args, **kwargs, and combined x, y : type.
Section headers restricted to the 14 known numpydoc titles with a length-matched underline, so prose followed by dashes is no longer a false header.
Nesting awareness for both reST fields and NumPy headers, via a PEP 257 base-indent helper. Fixes 7, 8.
extract_main_description stops at reST field lists and at real NumPy headers only.
extract_args_section now terminates at a NumPy header or reST field. This was a pre-existing main bug — the Google section ran to the end of the docstring — and it's what made cases 1 and 2 unfixable without touching it.

Which of courseitself is not complete, but do represent that some Extensive testing and justification is needed to the design. At minimal should not impact current functionality

Amir-R25 (changes requested):
Thanks @Mr-Neutr0n, could you please update the docs in `function_tools.md`, both the warning and add examples for the additional doc styles supported please?

We'd also appreciate if you could star the repo so railtracks becomes more used by the community!

Mr-Neutr0n:
Docs updated. The warning now says Google, NumPy, and reST/Sphinx parameter sections all parse, with a note that typehints stay on the signature regardless of style. Added a Supported docstring styles section with the same tool written in each of the three styles.

I checked all three examples against the parser from this branch before writing them in; each returns the same name-to-description mapping.

soulFood5632 (dismissed):
Looks great. thanks a ton for the contribution :)

CoronRing (commented):
This ticket need some updates, simply parsing each args is not enough, as their composition and lead to false parsing and conflicts with args. The system needs more complex reasoning and design. 

this does not mean EVERY single conflict across docstring formats need to be resolved (even pylance cannot do that). However, parsing should at minimal be non-destructive and regressive (worse than having only args parser. 

Some review were form last week, maybe fixed.

Mr-Neutr0n:
Addressed your review (2af5e110):

- **reST main description** — `extract_main_description` now stops at reST fields (`:param`, `:type`, `:returns:` etc.), so a pure-reST function no longer ships raw `:param url: ...` as its tool description. Verified with the example from your comment: main description now correctly returns `Fetches a URL.`

- **One style per docstring, first-match-wins** — `parse_docstring_args` now returns on the first style that yields args (Google → NumPy → reST) instead of blind-merging all three. Google keeps priority, so existing users see zero behavior change. This also fixes the inefficiency of running all three parsers unconditionally — we now stop after the first hit.

- **Section bleed** — `extract_args_section` (Google) now also terminates at NumPy headers (`_is_numpy_section_header`) and reST fields. This prevents the case you noted where a Google `Args:` section would swallow a following NumPy `Parameters` block (e.g., the `a: google b Parameters ----------` bleed in the mixed-style test).

The more extensive header-restriction (14 known numpydoc titles, length-matched underline) and full indentation-driven NumPy parsing you outlined would be a larger change — happy to follow up with that if you’d like, but this keeps the current PR minimal and non-breaking as you suggested (“At minimal should not impact current functionality”).

Let me know if you’d like the remaining edge cases ( `---` inside descriptions, `Parameters` as a param name, `*args`/`**kwargs` etc.) covered here or in a follow-up.

Mr-Neutr0n:
Merged latest main into the branch and resolved the one test conflict by retaining both the upstream Google parser regression and the NumPy/reST helper imports from this PR. All 26 docstring parser tests pass and Ruff is clean. CI is rerunning. https://github.com/Mr-Neutr0n/railtracks/commit/c37d15fa

Mr-Neutr0n:
@Amir-R25 ready for re-review.

Verified on HEAD (`b3916c00`):

- **Docs** (`5cbb5ea8`) — `function_tools.md` warning covers Google/NumPy/reST, plus a Supported docstring styles section with the same tool in all three styles.
- **CoronRing: no blind-merge** (`2af5e110`) — `parse_docstring_args` is first-match Google → NumPy → reST; Google Args extraction also stops at NumPy headers / reST fields.
- **CoronRing: reST main description** (`2af5e110`) — `extract_main_description` stops at `:param` / other reST fields (example tool description is `Fetches a URL.`, not the raw field list).
- **Regression tests** (`b3916c00`) — locked the above in `test_docstring_parser.py` (31 passing). Ruff clean.

Happy to resolve the two open review threads once you’ve taken another look.

packages/railtracks/src/railtracks/llm/tools/docstring_parser.py
Mr-Neutr0n: Addressed in `2af5e110` (parser) and locked in by regression tests in `b3916c00`. See latest PR comment for details.

packages/railtracks/src/railtracks/llm/tools/docstring_parser.py
Mr-Neutr0n: Addressed in `2af5e110` (parser) and locked in by regression tests in `b3916c00`. See latest PR comment for details.

CoronRing (changes requested):
@Mr-Neutr0n  Lots of thanks for taking this on! and the docs page with the three side-by-side
examples is a nice touch.

The overall shape is right: Google first, then NumPy, then reST, first
non-empty match wins. That keeps existing users on the same path. My main
concern is that the new break conditions were added into the *existing*
Google code paths rather than alongside them, so a few Google docstrings
that work on `main` now parse differently. Since this PR is meant to be
purely (as best as we can) additive, I'd like those fixed before merge.

I diffed the branch parser against `main`'s on each case below.

### Blocking: Google-style regressions

**1. `break` on a `:`-leading line truncates the rest of the Args block**
(`docstring_parser.py:90`)

```python
"""Do a thing.

Args:
    a: first
    :note: careful
    b: second
"""
```

`main` gives `{'a': 'first :note: careful', 'b': 'second'}`, this branch
gives `{'a': 'first'}`. `b` disappears from the tool schema with no warning.
This wants `continue` (or an indentation check) rather than `break`.

**2. The numpy-header check fires on the first Google arg line**
(`docstring_parser.py:92`)

Same docstring with `---` in place of the `:note:` line: `main` gives both
params, this branch returns `{}`. `_is_numpy_section_header` sees `a: first`
followed by a dashes-only line and aborts the whole section. Worth gating
this check so it can't trigger on an already-open `Args:` block.

**3. `extract_main_description` truncates ordinary prose**
(`docstring_parser.py:317`)

```python
"""Summary line.

Details
-------
more prose

Args:
    x: the x
"""
```

Description drops to `"Summary line."`. Same thing for a paragraph starting
with a role, e.g. ``:class:`Foo` is the entry point.``. The equivalent
truncation on NumPy/reST docstrings is clearly intentional and good; it just
needs to not apply when the docstring is Google-style.

### NumPy parser

`^\s*(\w+)\s*:` at `docstring_parser.py:221` is doing a lot of work, and a
few things slip through:

- No indentation guard, so a `Note: ...` line inside a description is read as
  a new parameter. Verified output: `{'x': 'the x value.', 'Note': 'keep this
  in mind.'}`. That phantom parameter reaches the LLM, and `x`'s description
  is silently clipped. The Google parser was hardened against exactly this at
  lines 130-139; the same guard would fix it here.
- Valid numpydoc forms are rejected: grouped `x, y : int`, untyped `z`, and
  `*args`/`**kwargs`. A docstring using grouped params returns `{}`.
- `docstring_parser.py:190` matches only the literal `Parameters`, so
  numpydoc's `Other Parameters` yields `{}`.
- The extractor stops only at *underlined* headers, so a trailing `Examples:`
  block leaks in and is parsed as a parameter:
  `{'x': 'the x', 'Examples': 'foo(1)'}`.
- `docstring_parser.py:201` has the same dashes-line abort as (2).

### reST

`docstring_parser.py:262`: the type group `(?:(?:\w+)\s+)?` can't match
subscripted or dotted types, so `:param List[int] items:` is skipped
silently. `(?:[\w.\[\], ]+\s+)?` or similar would cover it.

### Minor

- `_is_numpy_section_header(lines: list, i: int)` uses a bare `list`. Repo
  convention (see AGENTS.md) is to parameterize: `list[str]`.
- `tool.py:161` warns on `docstring.count("Args:") > 1`. Since the parser now
  understands two more styles, that check could cover `Parameters` and
  `:param` too.

All 67 tests pass, so each of the above is currently uncovered. A regression
test per numbered item would be the useful addition, particularly for the
three Google cases.

If you feel strongly about any of the cases being con over benefit, explain the case and I am happy to discuss further! 

Apologies for not doing in-line comments, but many of our contributors seem to prefer this way, let me know if you prefer the classic approach.

Mr-Neutr0n:
Pushed `2d83233` addressing every item, with a regression test per case (the parser suite goes from 31 to 40 tests).

**Google**
1. The `:` break now only fires at the indentation of `Args:` itself, so an indented `:note:` is continuation text again: `b` is no longer dropped and `a` keeps `:note: careful`.
2. The numpy-header abort is replaced by a section-name check, so a dashes-only line after `a: first` no longer empties the section.
3. `extract_main_description` detects a Google docstring and only stops at known section names there, so `Details` + `-------` and role-leading paragraphs stay in the description.

**NumPy**: added the indentation guard for `Note:`-style lines; grouped (`x, y : int`), untyped (`z`) and `*args`/`**kwargs` definitions are parsed and expanded per name (so they match `inspect` parameter names); `Other Parameters` is accepted; the block ends at known section names, including non-underlined ones like `Examples:`.

**reST**: the type group now accepts `List[int]` and `module.Type`.

**Minor**: `_is_numpy_section_header` uses `list[str]`; the `tool.py` warning counts `Args:`, `Parameters` and `:param`.

Verified with `ruff check . --no-fix`, `ruff format --check .` and the full unit suite (2566 passed; the one `test_type_mapping.py::test_builtin_function_raises_runtime_error` failure also fails without these changes, it looks environment-specific).

Mr-Neutr0n:
Rebased onto current `main` (the branch had picked up conflicts) and built the feature on top of the current Google parser from #1494, so the diff is now only the NumPy/reST support plus the review fixes. Locally: all 101 parser tests pass, the full unit suite is green (2740 passed), and `ruff check . --no-fix` plus `ruff format --check .` pass.

CoronRing (changes requested):
@Mr-Neutr0n Thanks for the patience! this is a long and impactful PR so its reveiwed in depth:

**A. The docs changes are missing from this head.** The head I reviewed last round touched `docs/documentation/agent_design/tools/function_tools.md` (+62, the three side-by-side examples). `f8dbea24` touches only `docstring_parser.py`, `tool.py` and `test_docstring_parser.py`, and the page still says docstrings must be "formatted in Google-style ... as that is what Railtracks automatically parses". I am fairly sure this was dropped in the rebase rather than removed on purpose. It needs to come back, both because shipping without it leaves the docs contradicting the behaviour and because it was Amir's original ask.

**B. The new warning in `tool.py` counts markers rather than sections, and it now misfires on Google docstrings too.** This is the one place where the PR is regressive for existing users. Measured on both sides:

```
                                     base    head
rest_two_params                      quiet   WARNS
numpy_other_params                   quiet   WARNS
google_prose_mentions_parameters     quiet   WARNS
google_two_args_sections             WARNS   WARNS
```

`sum(docstring.count(m) for m in ("Args:", "Parameters", ":param")) > 1` fires on any reST docstring with two parameters, on any NumPy docstring containing `Other Parameters` (the substring double counts), and on a Google docstring whose prose merely contains the word "Parameters". Only the fourth row is a real duplicate section. What you want is a count of sections per style: occurrences of `Args:`, header lines matching `Parameters` or `Other Parameters`, and any number of `:param` fields treated as a single reST section. Nothing asserts on this warning today, so a green suite tells you nothing here.

This one is on me. I flagged the check as too narrow without saying what it should count.

**C. Point 3 is only half fixed,** The `is_google` guard keys off the presence of an `Args:` section, but the regression class is "a wrapped prose line that begins with a reST role", which mostly happens in docstrings that have no `Args:` section at all. Those still take the `:`-break path and truncate mid-sentence:

```python
"""Chunk a document asynchronously.

Offloads the synchronous :meth:`chunk` call to a thread via
:func:`asyncio.to_thread` so the event loop is not blocked.
"""
```

`main` returns the full description. This branch returns `"Chunk a document asynchronously.\n\nOffloads the synchronous :meth:`chunk` call to a thread via"`, cut mid-sentence. Adding an `Args:` section to the same docstring makes it correct again, which is what makes the guard look like it works.

This is not hypothetical: **14 docstrings in railtracks' own source regress this way**, including `base.py::achunk`, `lexical.py::LexicalSearch`, `protocol.py::KeyValueStore` and several of the `_exceptions.py` classes. All 14 are cases with no `Args:` section, and in every one the head output is a truncation of the base output. Since this feeds the tool description the model sees, a description ending in "to a thread via" is worse than the Google-only behaviour it replaces.

The style detection probably needs to be positive rather than negative: treat the docstring as reST only when a line matches the actual `:param`/`:type`/`:returns:` field pattern, instead of treating any line starting with `:` as a section break.

## Minor, not blocking

**a.** `extract_main_description` breaks on any bare `_NUMPY_SECTION_NAMES` line regardless of style, so a plain docstring containing a standalone prose line reading `Notes`, `References` or `Examples` truncates where `main` kept it. Confirmed: `"Summary line.\n\nNotes\nsome more prose here."` becomes `"Summary line."`. Same root cause as C, a cross-style rule applied unconditionally.

**b.** `parse_rest_args_section` leaves its locals unannotated while the NumPy sibling declares `Dict[str, str]`. Worth matching.

## Repro

Worked from `f8dbea24` and its base `00c592c6` in two isolated worktrees.

- Differential corpus: extracted all 1214 docstrings from `packages/railtracks/src`, `examples/` and `docs/scripts/` and ran both parsers over them. 290 of those produce parameters, 752 parameters in total. **Zero** differences in `parse_docstring_args`. That is the non-destructive bar I asked for and you cleared it. The 14 differences were all in `extract_main_description`, which is item C.

Sum: the parameter parsing is in good shape and the test coverage is solid. A, B and C are all about the edges bleeding into the other two styles, and C is the only one that needs real thought.

Ciao =D

Mr-Neutr0n:
Thanks for the careful pass, and sorry for the couple of days' turnaround. All five points are addressed in `0a4d4a8`.

**A** is back. The docs section (the three side-by-side examples plus the warning rewrite) was dropped in the rebase, not removed on purpose.

**B** now counts sections per style: `Args:` headers, `Parameters` and `Other Parameters` header lines, and any number of `:param` fields as one reST section. Your four measured cases behave as you asked: `rest_two_params` quiet, `numpy_other_params` quiet, `google_prose_mentions_parameters` quiet, `google_two_args_sections` warns. Four tests now assert exactly this, since nothing did before.

**C** is fixed with positive detection. `extract_main_description` ends only at a line matching the actual field pattern (`:param`, `:type`, `:returns:`, plus the other field-list names), so `:func:` and `:class:` lines in wrapped prose stay in the description.

I re-ran your differential corpus over the repo (1297 docstrings from `packages/railtracks/src`, `examples/`, `docs/scripts/`). `parse_docstring_args` is identical to both `main` and the previous head. `extract_main_description` now matches `main` on every docstring in it, and the 16 truncations in that corpus are gone, including all 14 you listed (`achunk`, `KeyValueStore`, `LexicalSearch`, the `_exceptions` classes).

**a.** A bare `Notes`/`References`/`Examples` line no longer ends the description. Only a real NumPy header, the name followed by a dashes underline, does.

**b.** The `parse_rest_args_section` locals are annotated like their NumPy sibling.

Local verification: unit suite 2805 passed (the `test_type_mapping.py::test_builtin_function_raises_runtime_error` failure is pre-existing on Python 3.14 and fails on `main` here too), integration 314 passed, repo-wide `ruff check` and `ruff format --check` clean. The workflow run on the new head is waiting for approval to start.

CoronRing:
@Mr-Neutr0n Thanks for the quick turnaround, this round is a big step forward. I checked each point from last time independently rather than from the summary, and A, C, a and b are all fixed. B is mostly fixed, with one gap left (item 1 below).

- **Repo corpus** (1274 docstrings from `packages/railtracks/src`, `examples/`, `docs/scripts/`): zero differences from base in both `parse_docstring_args` and `extract_main_description`. The only warning changes are 4 docstrings in `docstring_parser.py` itself that mention `'Args:'` in prose, where base warned and head is now correctly quiet.
- **Wider corpus**: I also ran both parsers over 137,761 docstrings from boto3, matplotlib, scipy, aiohttp, azure, litellm, etc. to exercise the NumPy and reST paths on real code. Zero changes to any Google-parsed parameters. All 32,886 description changes are truncations at a real reST field or a real underlined NumPy header, with no stray causes. C is properly fixed.

A correction on my side: in the first round I said the NumPy `Note:` phantom parameter "reaches the LLM". It doesn't. `Tool.from_function` iterates the signature and looks each name up, so names that aren't parameters are dropped. The real harm in all of these cases is a wrong or truncated description on a *real* parameter, which is what I measured below.

## Should fix before merge

**1. `Parameters` + `Other Parameters` warns, and drops the `Other Parameters` entries.** This is the standard numpydoc layout, and it's where B is still regressive. `count_parameter_sections` counts each header separately, so it returns 2 and warns, while base is quiet. And `extract_numpy_args_section` stops at `Other Parameters` because the name is in `_NUMPY_SECTION_NAMES`, so everything under it is lost.

```python
"""Do.

Parameters
----------
a : int
    first

Other Parameters
----------------
b : int
    second
"""
```

Head gives `{'a': 'first'}` and warns. In the wider corpus this hits 92 docstrings (mostly matplotlib and scipy): all 92 newly warn, and 53 real parameters lose their description. `test_numpy_other_parameters_do_not_warn` only covers `Other Parameters` on its own, which is why the suite didn't catch it. The fix is to count NumPy parameter headers as one section (the same way `:param` fields are treated), and to continue through `Other Parameters` instead of breaking on it.

**2. The NumPy block is entered without an underline, so Google-ish `Parameters:` docstrings get wrong descriptions.** `extract_numpy_args_section` enters on any line reading `Parameters` or `Parameters:`, underline or not. Napoleon-style docstrings that use `Parameters:` in place of `Args:` then go through the NumPy parser, which reads `name: text` as `name : type` and takes the description from the lines that follow. The result is descriptions attached to the wrong parameter:

```python
"""Do.

Parameters:
    a: first
    b (int): second
"""
```

Head gives `{'a': 'b (int): second'}`. Base gives `{}`. Real examples from the corpus: in `litellm.completion`, `frequency_penalty` gets `logit_bias`'s description; in `httpcore.URL`, `scheme` gets only its second line (`Typically either "http" or "https".`). 42 docstrings are affected. A wrong description is worse than none, so this falls on the wrong side of the non-destructive bar. Your `_is_numpy_section_header` helper already does the right check, so gating the entry on it fixes this. It should accept `=` underlines as well as `-`: 18 docstrings in the corpus (jupyter_client, fsspec) use `==========` and parse correctly today.

**3. reST parameter descriptions are cut at a wrapped line that starts with a role.** This is the same root cause as C, but inside `parse_rest_args_section`: `elif stripped.startswith(":")` treats any `:`-leading line as a new field, including an inline role.

```python
""":param chunker: The chunker to use, typically a
    :class:`TextChunker` instance.
"""
```

Head gives `{'chunker': 'The chunker to use, typically a'}`, cut mid-sentence. Base returns `{}` here, so strictly this isn't a regression, but it ships exactly the kind of truncated text C was about. The wider corpus has 343 real parameters cut this way, e.g. boto3's `Session.client` `service_name` ends with "You can get a list of available services via". I first tried switching the check to `_is_rest_field`, but its name list is too short: about 80 `:ivar:`, `:cvar:`, `:exception:` and `:meta private:` fields got absorbed into the previous parameter. The minimal safe fix is to keep the current break and exempt only lines that start with a role (`:name:` followed by a backtick).

## Suggested patch

This covers all three items. I checked it against both corpora: the repo corpus is still byte-identical to base, Google parameters are unchanged across all 137k docstrings, every description is still a prefix of base, and all 109 tool tests pass. Per parameter in the wider corpus it fixes the 343 role truncations, restores the 53 `Other Parameters` entries, removes the misattributed `Parameters:` descriptions, and clears all 92 false warnings. The only other effect is that the 18 `=`-underlined NumPy docstrings now end their tool description at the header, the same as `-` ones already do.

```diff
+# Inline reST roles such as ":class:`Foo`" or ":py:meth:`bar`". A wrapped
+# description line may start with one, and it is prose rather than a new field.
+_REST_ROLE_PATTERN = re.compile(r"^:[\w.+-]+(?::[\w.+-]+)*:`")

 def _is_numpy_section_header(lines: list[str], index: int) -> bool:
     ...
     underline = lines[index + 1].strip()
-    return bool(underline) and set(underline) == {"-"}
+    return set(underline) in ({"-"}, {"="})

 def count_parameter_sections(docstring: str) -> int:
     ...
     sections = sum(1 for line in lines if line.strip().startswith("Args:"))
-    sections += sum(
-        1
-        for line in lines
-        if line.strip().rstrip(":") in ("Parameters", "Other Parameters")
-    )
+    if any(
+        line.strip().rstrip(":") in ("Parameters", "Other Parameters")
+        and _is_numpy_section_header(lines, index)
+        for index, line in enumerate(lines)
+    ):
+        sections += 1

 def extract_numpy_args_section(docstring: str) -> str:
     ...
-    for line in split_lines:
+    for index, line in enumerate(split_lines):
         stripped = line.strip()
         if not in_params_section:
-            if stripped.rstrip(":") in ("Parameters", "Other Parameters"):
+            if stripped.rstrip(":") in ("Parameters", "Other Parameters") and (
+                _is_numpy_section_header(split_lines, index)
+            ):
                 in_params_section = True
             continue
     ...
+        if stripped.rstrip(":") == "Other Parameters":
+            continue
         if stripped and stripped.rstrip(":") in _NUMPY_SECTION_NAMES:
             break

 def parse_rest_args_section(docstring: str) -> Dict[str, str]:
     ...
-        elif stripped.startswith(":"):
+        elif stripped.startswith(":") and not _REST_ROLE_PATTERN.match(stripped):
```

Please add a regression test for each of the three docstrings above, plus a warning test for the `Parameters` + `Other Parameters` combination.

## Minor, fine as follow-ups

- **a.** `parse_rest_args_section` only accepts `:param` and `:parameter`. Sphinx also treats `:arg`, `:argument` and `:key` as parameter fields, so those docstrings currently yield `{}`.
- **b.** A trailing paragraph or directive after the field list is absorbed into the last reST parameter, e.g. `:param x: the x` followed by a blank line and `.. note:: ...` gives `{'x': 'the x .. note:: ...'}`. Continuation lines probably need to be indented, or end at a blank line.
- **c.** Napoleon accepts `Parameters:` and `Arguments:` as aliases for `Args:`. With the gate from item 2 these return `{}`, same as base, which is fine for this PR. Supporting them properly in the Google parser would be a nice follow-up.

Sum: the core parsing is solid and the Google path is untouched, which was the main bar. The three items above are all small and the patch is ready to drop in. Once they're in, I'm happy to approve.

Ciao =D

PR #1460, opened 2026-08-20
chore: remove personal scratch-doc entries from shared .gitignore

Aryan-Railtown:
## Summary
- Removes the "Personal ticket scratch docs / playground scripts" block (`flow.md`, `decisions.md`, `middleware_explained.md`, `middleware_playground.py`, `verifier_playground.py`) that was merged into the shared `.gitignore` by mistake.
- These were personal, ticket-specific ignore entries; `flow.md`/`decisions.md` are already covered locally via `.git/info/exclude` and don't need to live in the shared, tracked `.gitignore`.

## Test plan
- [x] Confirmed none of the removed filenames are actually tracked in the repo (they were ignore-only entries, no leaked scratch content).
- [x] `.git/info/exclude` already ignores `flow.md`/`decisions.md` locally, so removing them from the shared file has no effect on local workflow.


.gitignore
```diff
@@ -51,10 +51,3 @@ packages/railtracks/docs/
 .claude
 CLAUDE.md
 graphify-out/
```
Pooria90: I suggest removing graphify, too. It's not part of our dependencies.

soulFood5632: agreed

Pooria90 (approved):

Pooria90 (approved):
Image

Issue #1457, opened 2026-08-20
OllamaLLM sends no tools: litellm drops `tools` for the `ollama/` prefix


CoronRing:
### Description

`OllamaLLM` prepends the `ollama/` litellm prefix to every model name ([`local/ollama.py`](https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/llm/models/local/ollama.py)):

```python
if not model_name.startswith("ollama/"):
    logger.warning(f"Prepending 'ollama/' to model name '{model_name}' for Ollama")
    model_name = f"ollama/{model_name}"
```

litellm treats `ollama` and `ollama_chat` as two different providers, and only `ollama_chat` supports the `tools` parameter. Because `litellm.drop_params = True` is set in `_litellm_wrapper.py`, the tool definitions are pruned from the request without any error:

```python
>>> import litellm
>>> litellm.drop_params = True
>>> TOOLS = [{"type": "function", "function": {"name": "f", "description": "d",
...           "parameters": {"type": "object", "properties": {}}}}]
>>> "tools" in litellm.utils.get_optional_params(
...     model="llama3.1", custom_llm_provider="ollama", tools=TOOLS, stream=True)
False
>>> "tools" in litellm.utils.get_optional_params(
...     model="llama3.1", custom_llm_provider="ollama_chat", tools=TOOLS, stream=True)
True
```

So an Ollama agent with `tool_nodes` sends no tools at all and the model can never call one — it just answers from the prompt. This is independent of streaming; it affects the buffered path too.

Compounding it, the guard in `OllamaLLM.chat_with_tools` passes, because litellm's capability catalog answers for the underlying model rather than the routing prefix:

```python
>>> litellm.utils.supports_function_calling("ollama/llama3.1")
True
```

so nothing surfaces to the user.

### Suggested fix

Route tool-capable calls through the `ollama_chat/` prefix. This changes buffered behaviour as well as streamed, so it wants validation against a real Ollama server before merging.

### Notes

Found while working #1434 (removing `TOOL_CALLING_STREAMING_BLACKLIST`), which is why Ollama was the one blacklisted provider whose streamed tool calling could not be verified — there is nothing to verify until the tools are actually sent. Not fixed there to keep that change scoped to streaming.

PR #1461, opened 2026-08-20
Replace the tool-calling streaming blacklist with a per-model capability check

### LINKED ISSUE #1434: Evaluate TOOL_CALLING_STREAMING_BLACKLIST and remove supporting models
### Description

models such as Anthropic and gemini natively support streaming, our router and system also support, we need to evaluate the block and remove them to enable proper streaming.

### LINKED ISSUE #1484: No Tracking metadata for openai llm calls
### Description

LLMs streamed with openai seem to not have any cost tracking. 

Image

Any streaming agent with openai will show this error.

CoronRing:
Closes #1434
Closes #1484

## Summary

`TOOL_CALLING_STREAMING_BLACKLIST` blocked streamed tool calling for Anthropic, Gemini, Azure, Ollama and HuggingFace. 

## Static blacklist → dynamic per-model fallback

The hardcoded provider set is replaced by `ModelBase.supports_streamed_tool_calling()`. The `LiteLLMWrapper` override probes litellm's capability catalog, and follows the existing `_model_in_litellm_catalog` convention of attempting the call when litellm has no metadata — uncatalogued Azure Foundry deployment names would otherwise be refused for a perfectly capable deployment.

The graceful buffered fallback is unchanged, just keyed on what the model can actually do instead of which company made it.

## Bugs fixed

**Gemini tool calls arrived with empty arguments.** Gemini packs the whole call (id + name + complete args) into one delta, where OpenAI/Anthropic send the id first and stream args as fragments. `_start_new_tool_call` ignored `call.function.arguments` on the opening delta, so every Gemini tool call reached its tool node as `{}`.

**Only the first call in a delta was read.** `delta.tool_calls[0]` dropped the rest when a provider batched parallel calls into one delta.

**A reused call index overwrote the call already accumulated there.** The old code parsed the previous occupant's arguments and then discarded it.

**A non-`stop`/`tool_calls` finish silently dropped tool calls.** `_is_stream_finished` matched on specific reasons, so a `length`-truncated turn lost the calls it had completed. Any `finish_reason` now ends the turn.

## Cost and token accounting on streamed calls

Streamed calls reported no tokens and no cost at all, on every provider including OpenAI (#1484):

- Anthropic and Gemini emit a usage chunk only when `stream_options={"include_usage": True}` is sent, and we never sent it. litellm exempts `stream_options` from its unsupported-param pruning, so it is safe to send to every provider.
- litellm only fills `_hidden_params["response_cost"]` in on a *buffered* response, so streamed cost was always `None`. `extract_message_info` now prices from the usage chunk when litellm leaves it unset.
- Usage is harvested wherever the chunk turns up rather than assuming it follows the `finish_reason` chunk, and a usage-only chunk with no `choices` no longer needs special-casing.

Without this, removing the blacklist would have *lost* cost tracking for Anthropic/Gemini/Azure, which previously fell back to buffered and so reported cost. `LLMResponseEvent` now carries the same tokens and cost either way.

## Verification

| | plain | tools | parallel | structured | agent loop | tokens + cost |
|---|---|---|---|---|---|---|
| OpenAI (control) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| Anthropic | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| Gemini | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| HuggingFace | ✅ | ✅ | ✅ | — | — | ✅ |
| Ollama | P | P | P | — | — | P |
| Azure | S | S | S | — | — | S |

HuggingFace was checked across several serverless Inference Provider routes, since behaviour there is per-model rather than per-provider:

- `together/openai/gpt-oss-20b` — streams tool calls correctly. The model chose to emit two calls on separate indices; both survive with correct arguments (the reused/parallel index path, confirmed on a third provider family).
- `nscale/Qwen/Qwen3-8B` — streams tool calls correctly.
- `novita/meta-llama/Llama-3.1-8B-Instruct` — the provider rejects tools outright (`"model features function calling not support"`). That surfaces as a clean `BadRequestError` rather than a silent degradation, which is the intended behaviour for HuggingFace: `HuggingFaceLLM._validate_tool_calling_support` deliberately defers to runtime because litellm's catalog can't answer for the whole HF model range.

Azure is skipped: it exposes an OpenAI-compatible interface, and access is currently limited.

## Notes

- Filed #1457: `OllamaLLM` prefixes model names with `ollama/`, which blocks the Ollama test.
- The `asyncio.CancelledError` traceback from `_bridge_sync_stream` noted here earlier was filed as #1464 and fixed on `main` by #1470, which is merged into this branch.


CoronRing:
### Ollama verification needed

Ollama is the one provider on this PR that could not be tested

<details>
<summary>Setup</summary>

```bash
ollama serve                  # in a separate terminal
ollama pull llama3.1          # or any tool-capable local model
```

Then run against this branch with railtracks installed. Override the model with `OLLAMA_MODEL` if you pulled something else.

</details>

```python
"""Probe Ollama's streamed tool calling. Requires a running Ollama server."""

import asyncio
import os

import railtracks as rt
from railtracks.llm.content import ToolCalls
from railtracks.llm.history import MessageHistory
from railtracks.llm.message import UserMessage
from railtracks.llm.tools import Parameter, Tool

MODEL = os.environ.get("OLLAMA_MODEL", "llama3.1")

WEATHER = Tool(
    name="get_weather",
    detail="Get the current weather for a city.",
    parameters=[
        Parameter(
            name="city", param_type="string", description="City name", required=True
        )
    ],
)


def get_weather(city: str) -> str:
    """Get the current weather for a city."""
    return f"{city}: 14 degrees celsius, light rain"


async def plain_chat(llm):
    """Streaming with no tools involved. Should pass regardless of #1457."""
    chunks, final = [], None
    async for item in llm.astream_chat(
        MessageHistory([UserMessage("Write a 60-word paragraph about trains.")])
    ):
        if isinstance(item, str):
            chunks.append(item)
        else:
            final = item
    return {
        "n_chunks": len(chunks),
        "in_tok": final.message_info.input_tokens,
        "out_tok": final.message_info.output_tokens,
        "cost": final.message_info.total_cost,
    }


async def model_level_tools(llm):
    """Does a streamed tool call come back with its arguments intact?"""
    final = None
    async for item in llm.astream_chat_with_tools(
        MessageHistory([UserMessage("What is the weather in Vancouver? Use the tool.")]),
        tools=[WEATHER],
    ):
        if not isinstance(item, str):
            final = item
    content = final.message.content
    calls = list(content) if isinstance(content, ToolCalls) else None
    return {
        "tool_calls": [(c.name, c.arguments) for c in calls] if calls else None,
        "in_tok": final.message_info.input_tokens,
        "cost": final.message_info.total_cost,
    }


async def agent_loop(llm):
    """Full rt.astream agent loop: streamed tool call has to round-trip as a
    ToolMessage and the follow-up turn has to stream."""
    agent = rt.agent_node(
        name="WeatherAgent",
        tool_nodes={rt.function_node(get_weather)},
        system_message="You answer weather questions using the get_weather tool.",
        llm=llm,
    )
    stream = rt.astream(agent, user_input="What is the weather in Vancouver?")
    chunks = [c async for c in stream]
    text = stream.result.text or ""
    return {
        "n_chunks": len(chunks),
        "used_tool": "14 degrees" in text,
        "chunks_match_result": "".join(chunks) == text,
        "result": text[:80],
    }


async def main():
    print(f"model: ollama/{MODEL}\n")
    for label, probe in [
        ("plain chat", plain_chat),
        ("model-level tools", model_level_tools),
        ("agent loop", agent_loop),
    ]:
        try:
            with rt.Session(flow_name="ollama-probe"):
                res = await asyncio.wait_for(probe(rt.llm.OllamaLLM(MODEL)), timeout=300)
            print(f"[OK]   {label:20} {res}")
        except Exception as e:
            print(f"[FAIL] {label:20} {type(e).__name__}: {str(e)[:400]}")


if __name__ == "__main__":
    asyncio.run(main())
```

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
```diff
@@ -564,50 +580,70 @@ def _prepare_response(
...
     def _is_stream_finished(self, choice) -> bool:
-        """Check if the stream has finished."""
-        return choice.finish_reason in ("stop", "tool_calls")
+        """Check if the stream has finished.
+
+        Any `finish_reason` the provider sets ends the turn. 
+        
+        Matching on the reason itself would be provider-specific. 
+        """
+        # On finish Sould drop the accumulated tool calls of a truncated (`length`) or filtered turn.
```
soulFood5632: spelling error

soulFood5632 (commented):
It looks good to me. I want to get a second set of eyes on it before we merge @Amir-R25

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
CoronRing: Wow that is some detailed review, thanks a lot!

Amir-R25 (approved):
Thanks

PR #1467, opened 2026-08-20
Ollama fix

### LINKED ISSUE #1457: OllamaLLM sends no tools: litellm drops `tools` for the `ollama/` prefix

CoronRing:
## Summary

Closes #1457

ollama/ (gpt 3.0 age, chatless) should be deprecated for ollama_chat/, as it does not support tooll call

## Type of change

- [X ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs


soulFood5632 (approved):
LGTM

PR #1470, opened 2026-08-20
1464 stream cancelled callback

### LINKED ISSUE #1464: Streaming: CancelledError traceback on stderr when a stream is abandoned early
### Description

Abandoning a stream early prints an `asyncio.CancelledError` traceback to stderr, on a run that otherwise succeeds.

`_bridge_sync_stream`'s cleanup attaches a done-callback to mark a late worker failure as retrieved ([`_litellm_wrapper.py:446`](https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py#L446)):

```python
finally:
    stop.set()
    if worker.done():
        worker.exception()
    else:
        worker.add_done_callback(lambda t: t.exception())
```

`Task.exception()` **re-raises** on a cancelled task rather than returning the exception. The raise happens inside a `call_soon` callback, where asyncio has nowhere to put it, so it is reported as `Exception in callback`. The same applies to the `worker.done()` branch when the worker was cancelled rather than completed.

Cosmetic — the stream result is unaffected — but a red traceback on a passing run is a bad first impression.

The mechanism, with no railtracks involved:

```python
import asyncio, time

async def main():
    t = asyncio.ensure_future(asyncio.to_thread(lambda: time.sleep(10)))
    await asyncio.sleep(0)
    t.cancel()
    t.add_done_callback(lambda x: x.exception())   # _litellm_wrapper.py:446
    await asyncio.sleep(0.2)

asyncio.run(main())
```

```
Exception in callback main.<locals>.<lambda>(<Task cancell...hreads.py:12>>)
Traceback (most recent call last):
  File ".../asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "<stdin>", line 12, in <lambda>
  File ".../asyncio/threads.py", line 25, in to_thread
    return await loop.run_in_executor(None, func_call)
asyncio.exceptions.CancelledError
```

In practice this shows up when a streamed call is abandoned before its worker thread finishes — observed against a HuggingFace serverless route while working #1434.

### Suggested fix

Treat cancellation as nothing to retrieve:

```python
if worker.done():
    if not worker.cancelled():
        worker.exception()
else:
    worker.add_done_callback(lambda t: t.cancelled() or t.exception())
```

### Notes

Unrelated to #1462, which is a different exception (`RuntimeError('Publisher is not currently running.')`) from a different module and mechanism — a task publishing after session teardown, surfacing as `Task exception was never retrieved`. The two only share the symptom of a stderr traceback on a successful run.

CoronRing:
closes #1464 

Treat cancellation as nothing to retrieve:

if worker.done():
    if not worker.cancelled():
        worker.exception()
else:
    worker.add_done_callback(lambda t: t.cancelled() or t.exception())


soulFood5632 (approved):
LGTM. just make sure the merge bases are correct since this includes some stuff from the other PR

PR #1472, opened 2026-08-20
docs: fix README model IDs, add API key step, tidy first_agent script

CoronRing:
Documentation-only change, not worth their own tickets extends #1114

## README

- **Fix broken Anthropic model ID.** `claude-4-6-sonnet` had the tier and version swapped, so it fails at call time as a provider 404. Now `claude-sonnet-5`.
- **Add a docstring to `get_weather`** in the headline example. `function_node` derives the tool description and parameter descriptions from the docstring, so without one the LLM saw a tool with an empty description — in the first snippet a new user reads.
- **Fix `tool_nodes` in the first example.** `(rt.function_node(get_weather))` was a parenthesized expression, not a tuple, so `agent_node` received a single node class where it expects an iterable. Both examples now use a list, matching the rest of the docs.
- **Update models to `gpt-5.4-mini`**, replacing `gpt-4o`, and align the LLM Support block with the examples.
- **Print the result** in the second example — it claimed to solve "how many r's in strawberry" but discarded the return value.
- **Add a "Set your API key" step** to Quick Start. Nothing previously told a new user where credentials go; the snippet writes `OPENAI_API_KEY` to `.env`, which `railtracks` loads on import, and links to the LLM setup page for other providers.
- Remove an empty `<div align="center">` block.

## docs/scripts/first_agent.py

- Remove `from ast import Str`  unused, and an `ImportError` on Python 3.12+ since `ast.Str` was removed, while the project supports `>=3.10`.
- Replace the deprecated `claude-sonnet-4-20250514` with `claude-sonnet-5` and `gpt-4o` with `gpt-5.4-mini`.

## Not included

`gpt-4o` still appears in roughly 30 other files under `docs/scripts/` and `docs/`, and `claude-sonnet-4-6` remains in `docs/scripts/providers.py`, `llm_setup.md`, and the CLI skills. Those are left for a separate sweep to keep this diff reviewable. The `gpt-4o` references in `docs/documentation/upgrading/1_5_0.md` should probably stay regardless, since that page documents a historical migration.


soulFood5632 (approved):
LGTM

Amir-R25 (approved):

PR #1473, opened 2026-08-20
fix: raise NodeCreationError when agent_node gets both tool_nodes and output_schema

Aryan-Railtown:
## Summary
- `tool_nodes` + `output_schema` together on `rt.agent_node()` behaves differently as of the recent v1.5 update: the model no longer reliably calls the tool in that combined mode and can fabricate a plausible tool result instead of actually invoking it.
- We caught this and are shipping a quick fix: `agent_node()` now raises `NodeCreationError` at construction time when both are provided, instead of silently building a broken agent. No tracked issue for this yet.

## Changes
- New validation check wired into `agent_node()` at creation time (`validation/node_creation/validation.py`).
- Docstring + `agent-builder.md` updated to document the new hard error.
- Test added/updated to assert the error is raised; existing tests for the two valid single-mode cases (tools-only, schema-only) untouched and still passing.

## Test plan
- [x] `pytest packages/railtracks/tests/unit_tests/built_nodes/llm/test_agent.py packages/railtracks/tests/integration_tests/guardrails/test_agent_node_guardrails.py` — 32 passed
- [x] Verified live: combined case raises `NodeCreationError`; tools-only and schema-only construction still work


soulFood5632 (approved):
LGTM

PR #1479, opened 2026-08-22
1478 skills registry and symbol validation

### LINKED ISSUE #1478: Define On-Disk Skill Format and Skill Registry
### Description

Define the on-disk format for a skill and make the `SKILLS` registry a function of the
skills directory rather than a hardcoded dictionary in `cli/__init__.py`.

This issue delivers the format contract and the discovery/validation layer only. It does not change
how any skill is installed; every existing handler keeps writing exactly what it writes today, **fed
from disk instead of from Python**. It is the foundation every other sub-task of #1451 builds on.

## Why
The frontmatter schema, in particular how target-specific metadata is expressed, is the one
decision in #1451 that cannot be cheaply revised. Every install handler and every bundled `SKILL.md`
will depend on its shape. Settling it before any handler is written means the per-tool sub-tasks
become mechanical.

## Proposed schema
A bundled skill is a directory `cli/skills/<name>/` containing `SKILL.md`. Its frontmatter is
Claude-native at the top level, with target-specific keys nested under `tools:`:

```yaml
name: agent-builder
description: Build an agent using the railtracks Python framework. Use when ...
argument-hint: "[describe what the agent should do]"
tools:
  cursor:
    globs: "**/*.py"
    alwaysApply: false
  claude:
    allowed-tools: [Read, Edit, Bash]
```
The top level is already what the workspace drafts in `.claude/skills/`
use, so installing for Claude Code becomes close to a verbatim copy, and every other target derives
its frontmatter by projection.

## Requirements
- A Skill value type carrying at minimum: name, description, argument_hint, body, directory, supporting_files, and the raw per-tool metadata mapping.
- discover_skills() scans cli/skills/*/SKILL.md and returns the discovered skills.
- Validation, raised as a clear error naming the offending directory:
- SKILL.md exists and has parseable YAML frontmatter
- name is present and equals the directory name
- description is present and non-empty
- no unrecognised top-level key
- SKILLS becomes a computed mapping over discovery results. add_skill and list_skills capabilities read from it; their observable behavior is unchanged.
- Discovery is a pure read of the packaged directory; no network, no import of the framework itself, and cheap enough to run on every railtracks add.
- Frontmatter parsing uses PyYAML
- Tests: discovery finds every bundled skill; each validation rule fails with a useful message; SKILLS exposes the same keys and values the current hardcoded dict does.

Pooria90:
## Summary

Makes the bundled-skill registry a function of **the skills directory** instead of a hardcoded dict.

A skill is now a directory `cli/skills/<name>/SKILL.md` whose YAML frontmatter is the single source of its metadata; `discover_skills()` reads and validates them, and `SKILLS` is derived from that. The three bundled skills move to the new layout, and a new symbol-existence check resolves every `rt.*` / `railtracks.*` symbol a skill mentions — prose and code blocks alike — against the installed package, so skills that drift out of date fail a test instead of silently misleading an agent. 

Install behaviour is deliberately unchanged: handlers still write exactly one file, now fed from disk rather than from Python. 

Closes #1478.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

**Known limitation, measured not overlooked.** Only *dotted* mentions are checked; A bare `` `TerminalLLM` `` in a table cell is not. Closing it needs a package-wide symbol index, which is tracked as a follow-up. Recorded as `test_bare_class_names_are_a_known_blind_spot` so it's discoverable rather than silent.


packages/railtracks/src/railtracks/retrieval/loaders/__init__.py
```diff
@@ -1,3 +1,5 @@
+from typing import TYPE_CHECKING
```
Amir-R25, line 1: seems like an unrelated change, but the change itself is correct, so we can keep it :D

Amir-R25 (approved):
Nice, maybe Claude can finally use the framework correctly

PR #1480, opened 2026-08-22
fix: treat docstring continuation lines as description text

### LINKED ISSUE #1476: Google-style docstring parser treats continuation lines starting with `Word:` as new parameters
`parse_args_section` in `packages/railtracks/src/railtracks/llm/tools/docstring_parser.py` matches any indented line of the form `<word>:` as a new parameter definition. Continuation lines of a description that happen to start that way (e.g. `Note:`, `Example:`, `See:`) get picked up as bogus parameters, and the continuation text is dropped from the real parameter's description.

Reproduces on `main`.

```python
from railtracks.llm.tools.docstring_parser import parse_docstring_args

docstring = """
Fetches a URL.

Args:
    headers: Optional dict of extra headers.
        Note: keys are case-insensitive.

Returns:
    The response body.
"""

parse_docstring_args(docstring)
```

Actual:

```python
{'headers': 'Optional dict of extra headers.', 'Note': 'keys are case-insensitive.'}
```

Expected:

```python
{'headers': 'Optional dict of extra headers. Note: keys are case-insensitive.'}
```

Downstream effect: the LLM sees a phantom `Note` parameter on the tool schema, and the intended continuation text is missing from the real parameter's description.

The regex `^\s*(\w+)(?:\s*\([^)]+\))?:\s*(.+)$` is applied line-by-line without tracking indent. Anchoring new-parameter detection to the indent level of the first `name:` line in the section would rule out continuation lines.

lesbass:
## Summary

Fixes #1476.

`parse_args_section` treated any indented line matching `Word:` as a new parameter definition, so continuation lines like `Note: keys are case-insensitive.` minted a phantom `Note` parameter and dropped the text from the real parameter's description.

## Change

Anchor new-parameter detection to the indent level of the first parameter line in the `Args:` section. Only lines indented at that same level start a new parameter; deeper-indented lines that look like parameter definitions are kept as continuation text.

## Verification

- Added regression tests covering the `Note:` continuation-line case at both `parse_args_section` and `parse_docstring_args` levels.
- `uv run pytest packages/railtracks/tests/unit_tests/llm/tools/ -q` → 51 passed.
- `uv run ruff check ...` on the two changed files → All checks passed.

The repro from the issue now returns:

```python
{'headers': 'Optional dict of extra headers. Note: keys are case-insensitive.'}
```

### DIFF
--- a/packages/railtracks/src/railtracks/llm/tools/docstring_parser.py
+++ b/packages/railtracks/src/railtracks/llm/tools/docstring_parser.py
@@ -94,11 +94,12 @@ def parse_args_section(args_section: str) -> Dict[str, str]:
     # This handles both formats:
     # - param_name: Description
     # - param_name (type): Description
-    pattern = re.compile(r"^\s*(\w+)(?:\s*\([^)]+\))?:\s*(.+)$")
+    pattern = re.compile(r"^(\s*)(\w+)(?:\s*\([^)]+\))?:\s*(.+)$")
 
     arg_descriptions = {}
     current_arg = None
     current_description = []
+    param_indent = None
 
     for line in args_section.splitlines():
         # Skip empty lines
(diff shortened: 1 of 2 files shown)


Pooria90 (approved):

Pooria90:
Thanks for your contribution @lesbass. All looks good and I merge!

Issue #1451, opened 2026-08-20
Skill Infrastructure: Directories Support


Pooria90:
### Description

`railtracks add <tool>:<skill>` currently installs a skill as a **single flat file**, with its
frontmatter synthesized in Python from the `SKILLS` dict in `cli/__init__.py`. This feature makes a
bundled skill a **directory on disk**: a `SKILL.md` carrying its own frontmatter, plus whatever
supporting files it needs. It teaches the installer to copy that directory, deriving all metadata
from disk rather than from a hardcoded registry.

This is prerequisite work for the skill catalog proposed in the coding-assistant skills review
(#1343): it must land before the catalog grows in number of skills.

### Why

- **Progressive disclosure.** A flat skill loads in full every time it fires. A directory lets
  `SKILL.md` be a short router that pages in detail on demand, so adding skills stops being a linear
  context tax on every session.
- **Single source of truth for metadata.** Frontmatter lives next to the content it describes.
  Adding a skill becomes "add a directory" instead of "add a file *and* remember to edit Python."
  It also gives per-tool metadata (`allowed-tools`, `model`, `disable-model-invocation`, Cursor's
  `globs`) somewhere to live — today there is nowhere, so bundled skills lose even the
  `argument-hint` the drafts carry.
- **Skills can ship code, not just prose.** A directory can carry scripts, templates and stubs. That
  changes what a skill can do: a debugging skill can ship a working session-graph dump script rather
  than describing one; a scaffold can be a real file. Prose rots silently — a script either runs or
  fails loudly. It is also the only clean route to a skill that verifies itself.
- **Install-time version awareness.** A directory can carry a generated manifest, so a later
  `railtracks add` can detect "your installed skill is from 1.4.0, the package is 1.6.0" and offer
  to update. Not feasible with an opaque single file.

### Requirements
(More like suggestions; sub-tasks are not defined yet.)

- Bundled skills move from `cli/skills/<name>.md` to `cli/skills/<name>/SKILL.md`, with frontmatter
  authored in the file.
- The `SKILLS` registry is **derived from disk** (discovered by scanning the skills directory), not
  hand-maintained in `cli/__init__.py`. `railtracks add --list` reads from the same source.
- The installer copies a skill *directory*, preserving its internal structure and any supporting
  files, rather than writing one file.
- Per-target install behaviour is defined for each supported assistant. Targets that have no
  directory concept need an explicit, documented degradation policy; **filed as sub-tasks**, one
  per tool.
[...]

Issue #1341, opened 2026-08-05
Ability to Add All Skills


soulFood5632:
### Description

Right now, we can add skills through `railtracks add claude:agent-builder`. It would be nice allow us to run a command which adds all of the skills at once through something like `railtracks add claude:all`

of course we should have a similar functionality for all coding agents.

kudala-bharani:
I would like to work on this. I will add `<tool>:all` for all four supported assistants, reuse the existing installers and overwrite behavior, and cover the CLI path with tests plus a short setup-doc example.

PR #1483, opened 2026-08-25
Skills Infra: Directory Support

Pooria90:
## Summary

Skill infrastructure epic: `railtracks add` installs bundled skills as **skill directories** (`SKILL.md` + supporting files) natively for Claude Code, Codex, Copilot, and Cursor, tracked by a per-skill manifest so re-installs sync instead of blindly copying. Part of the work in #1451 (also lands #1479, #1524, #1525, #1536, #1341).

Highlights:
- On-disk skill format, registry, and a CI symbol-existence check (#1479).
- Directory install parameterised by `(root, frontmatter projection)`; one provider file per assistant under `_skillkit/providers/` (#1524, #1536).
- Per-skill `.railtracks.json` manifest + directory sync: dropped files removed, user-edited files preserved, untouched re-installs silent (#1525).
- Legacy Copilot/Cursor installs are **detected and reported, never removed** (#1534 tracks removal).
- `railtracks add <tool>:all` (#1341).

## Type of change

- [x] Feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Docs

## Checklist

## Notes

**User-visible path change:** Copilot skills now install to `.github/skills/<name>/` (was the `copilot-instructions.md` marker block) and Cursor to `.cursor/skills/<name>/` (was `.cursor/rules/<name>.mdc`). Old installs are left in place and reported, not migrated automatically.

**Reviewer guide (suggested order):**
1. `_skillkit/registry.py` — skill format + discovery
2. `_skillkit/install.py` — the install/sync flow and the `(root, projection)` seam
3. `_skillkit/providers/*.py` — the four per-assistant handlers
4. `_skillkit/manifest.py` — manifest + legacy detection (report-only)
5. `cli/__init__.py` — `add_skill` dispatch + `:all`
6. `tests/unit_tests/cli/skillkit/` + `test_cli.py`, then `docs/.../ai_setup.md`

**Merge note:** `main`'s `#1540` add-all tests were re-based onto this branch's manifest-based installer (untouched re-installs are silent; Copilot now prompts like any directory install).

**Next steps (post-merge, before delegating content work):** bare-name symbol index → audit + convert `rag-pipeline` to the folder format as the pilot → snippet compile-check.


Amir-R25:
pasting this here so I won't lose it:

```markdown
## Review: Skills Infra: Directory Support — #1483

**Scope:** 26 files, +2,867 / −217; reviewed commit `8040099`.
Paths below are under `packages/railtracks/src/railtracks/cli/_skillkit/`.

### 🐛 Bugs

- **[P1] `manifest.py:179` — Manifest paths can delete files outside the skill directory.**
  `path = destination / recorded.path` accepts `../` components without checking containment. I added `../../../project-notes.md` and its matching checksum to an installed skill’s manifest; running `railtracks add claude:agent-builder` deleted that project-root file without prompting. Since manifests are intended to be committed, a crafted manifest can trigger this during an ordinary reinstall. Validate paths before hashing or deleting them, including resolved-path containment.

- **[P2] `install.py:104` — Valid source metadata produces invalid installed YAML.**
  `f"{key}: {value}\n"` emits parsed strings without YAML escaping. A valid source description such as `"Build agents: tools and flows"` becomes `description: Build agents: tools and flows`, which raises `ScannerError`. Reproduced on all four providers. Serialize the projected fields through YAML rather than interpolating their values.

- **[P2] `install.py:175` — File-to-directory upgrades fail after partially updating the skill.**
  `path.parent.mkdir(parents=True, exist_ok=True)` runs before stale files are pruned. If a release replaces `references/api` with `references/api/index.md`, the old file blocks directory creation and raises `FileExistsError`, including with `--force`. `SKILL.md` has already been updated while supporting files and the manifest remain old. Resolve tracked file/directory conflicts before writing the new content, preserving the checks for user edits.

### Validation

- All 118 skillkit tests passed.
- All three findings were reproduced against the PR snapshot.
- Broader CLI tests could not collect: the local environment lacks `fastapi` and `duckdb`.

### Verdict

🔴 Blocked on bugs.
```

.github/copilot-instructions.md
```diff
@@ -192,7 +195,7 @@ Steps 3 and 4 are gated on changed paths, so a docs-only PR skips the test suite
   GENERATED — do not edit by hand.
 
   The section below is the bundled `agent-builder` skill
-  (packages/railtracks/src/railtracks/cli/skills/agent-builder.md). Edit that file,
+  (packages/railtracks/src/railtracks/cli/skills/agent-builder/SKILL.md). Edit that file,
   then regenerate from the repository root with:
```
Aryan-Railtown, line 200: also nitpick, middleware and rag-pipeline skills have been added as well, maybe we should add those somewhere in this file (or hopefully merge this file with `AGENTS.md`)

.github/copilot-instructions.md
```diff
@@ -26,10 +26,13 @@ uv pip install -e "packages/railtracks[all]"     # SDK with all optional extras
 
 The SDK is deliberately thin by default; features gate behind extras declared in `packages/railtracks/pyproject.toml` (`visual`, `retrieval`, `chroma`, `portkey`, `stores-*`, …).
```
Aryan-Railtown, line 27: Not sure if we need this file separate from `AGENTS.md`, if i recall correctly, copilot is also able to read and take instructions  from the agents file. We can link these two files together (symlink) so that it is easier to maintain.

packages/railtracks/src/railtracks/cli/__init__.py
```diff
@@ -51,35 +60,15 @@
...
-    "middleware": {
-        "name": "middleware",
-        "description": (
-            "Use middleware as part of your railtracks agent. Use when you want build resilient and effective agents"
-        ),
-        "argument_hint": "[describe the middleware to implement]",
-    },
+# The rich objects; the source of truth for everything about a bundled skill.
+SKILL_REGISTRY: dict[str, Skill] = discover_skills()
+
```
Aryan-Railtown, line 68: one thing to flag: SKILL_REGISTRY = discover_skills() now runs at module import time with nothing catching a bad parse. Since query/connect.py imports from cli.io, that pulls in this whole module, so rt.connect(...) or viz now transitively re-scans and re-validates every bundled skill file, and a malformed SKILL.md would blow up an unrelated code path. Might be worth wrapping this in a try/except (or lazily discovering on first use) so a bad skill file can't take down callers that never touch skill management.

packages/railtracks/src/railtracks/cli/_skillkit/install.py
```diff
@@ -0,0 +1,203 @@
...
+            continue
+        drop_next_blank = False
+
+        if "$ARGUMENTS" not in line:
+            kept.append(line)
+            continue
+
+        # Drop the whole intro line (and its trailing blank); reword inline mentions in place.
+        if line.rstrip().endswith(": $ARGUMENTS"):
+            drop_next_blank = bool(kept) and not kept[-1].strip()
```
Aryan-Railtown, line 59: Minor edge case:
this line only fires if `kept` already has lines in it. If the $ARGUMENTS-intro line is the very first line of the skill body, kept is still empty, so the following blank line doesn't get dropped... leaves a stray leading blank line in the installed file for Codex/Copilot/Cursor. Probably fine to special-case not kept the same way as a "drop" instead of a no-op.

packages/railtracks/src/railtracks/cli/_skillkit/manifest.py
```diff
@@ -0,0 +1,309 @@
...
+        f"'{previous.skill}' was installed from railtracks "
+        f"{previous.package_version}; this is {current}."
+    )
+
+
+# ---------------------------------------------------------------------------
+# Legacy install detection — report only
+# TODO: drop this section once the older install shapes are out of circulation.
+# ---------------------------------------------------------------------------
+
```
Aryan-Railtown, line 222: Just a nit, i am not sure what this is about, but do you think we should make an issue and link this PR to that issue instead of having an inline TODO in the code base?

packages/railtracks/src/railtracks/cli/__init__.py
```diff
@@ -301,148 +290,29 @@ def update_railtracks(beta: bool = False):
...
-            drop_next_blank = bool(kept) and not kept[-1].strip()
-            continue
-
-        kept.append(
-            line.replace("`$ARGUMENTS`", "the request").replace(
-                "$ARGUMENTS", "the request"
-            )
-        )
-    return "\n".join(kept)
+    return install_skill_directory(skill, CLAUDE, force)
```
Aryan-Railtown, line 301: Small cleanup thought, `_add_claude/_add_codex/_add_copilot/_add_cursor` are all one-liners forwarding to install_skill_directory. we could probably collapse _TOOL_HANDLERS to map tool name straight to the InstallTarget constant and call install_skill_directory directly, saves ~24 lines.

Aryan-Railtown (changes requested):
Amazing workk @Pooria90 !! just a few thoughts

Pooria90:
> pasting this here so I won't lose it:
> 
> ```
> ## Review: Skills Infra: Directory Support — #1483
> 
> **Scope:** 26 files, +2,867 / −217; reviewed commit `8040099`.
> Paths below are under `packages/railtracks/src/railtracks/cli/_skillkit/`.
> ...

Thanks Amir @Amir-R25. I made these changes:

`manifest.py:179`: path traversal delete (P1)
- Fixed. Added a `_within()` containment check in `stale_files`, so recorded paths that resolve outside the skill directory (`..` or absolute) are skipped rather than hashed or deleted. Regression test `test_a_traversal_path_is_never_removed` added.

`install.py:104`: invalid installed YAML (P2)
- Fixed. Projected fields now serialize through `yaml.safe_dump` (new `_yaml_line` helper) instead of raw interpolation, so colons/brackets are escaped and round-trip. The Claude provider's `_quote` hack is gone. Test `test_a_description_with_yaml_metacharacters_stays_valid` added.

`install.py:175`: file→directory upgrade crash (P2)
- Fixed. `_directory_conflicts` now detects files sitting where a directory must go, folds them into the overwrite prompt, and clears them before writing; no more `FileExistsError` mid-update. Test `test_a_file_upgraded_to_a_directory_reinstalls_cleanly` added.

.github/copilot-instructions.md
Pooria90: I added some details that were missing from `AGENTS.md` and linked it in the copilot file. Please review!

packages/railtracks/src/railtracks/cli/__init__.py
Pooria90: Good catch. Discovery is now deferred behind `_load_skills()` and reached via accessors + module `__getattr__`, so importing `cli.io` for `rt.connect`/viz no longer scans or can fail on the bundled skills.

packages/railtracks/src/railtracks/cli/_skillkit/install.py
Pooria90: Fixed to` drop_next_blank = not kept or not kept[-1].strip()`, so an intro line at the start of the body no longer leaves a leading blank. Added tests for the leading, mid-paragraph, and after-text cases.

packages/railtracks/src/railtracks/cli/_skillkit/manifest.py
Pooria90: Added this issue: https://github.com/RailtownAI/railtracks/issues/1534

packages/railtracks/src/railtracks/cli/__init__.py
Pooria90: Done!

Aryan-Railtown (dismissed):
Perfect!! Looks good :D

Aryan-Railtown (approved):

PR #1487, opened 2026-08-25
Fix tool schemas degrading to `{"type": "object"}`

### LINKED ISSUE #1458: Tool schemas silently degrade to `{"type": "object"}` under `from __future__ import annotations`
## Summary

`Tool.from_function` reads parameter types from `inspect.signature(func).parameters[n].annotation`
without resolving them. Under PEP 563 (`from __future__ import annotations`) that attribute is a
**string**, not a type, so no handler matches and every parameter falls through to
`ParameterType.OBJECT`.

The result is a valid-looking schema in which every argument is `{"type": "object"}`. No exception,
no warning. The model is told a regex pattern is an object, sends `{}` or malformed arguments, and
the failure surfaces as apparent model-quality degradation rather than a library bug.

`from __future__ import annotations` is common practice and is required for `X | None` syntax on
older interpreters. For reference, 74 modules in standard library use it.

## Root cause

`llm/tools/tool.py`

```python
127:  signature = inspect.signature(func)                 # annotations left unresolved
172:  handler = next(h for h in handlers if h.can_handle(param.annotation))
175:  param_obj = handler.create_parameter(param.name, param.annotation, description, required)
```

With PEP 563, `param.annotation` is `'str'` rather than `str`, so
`ParameterType.from_python_type('str')` returns `OBJECT`, and `DefaultParameterHandler` is the last
handler standing.

## Follow-on: `ToolManifest` cannot be used as a workaround

Supplying an explicit manifest should bypass inference, but
`validate_tool_manifest_against_function` (`validation/node_creation/validation.py:335`) diffs the
manifest against the **broken** inference:

```python
rt.function_node(search_files, manifest=rt.ToolManifest("Search files.", [
    parse_json_schema_to_parameter("pattern", {"type": "string"}, True),
]))
```

```
NodeCreationError: Type mismatch for parameter 'pattern':
function expects 'ParameterType.OBJECT', but manifest specifies 'string'.
```

The bug's own output is treated as the source of truth, so the correct manifest is rejected and the
only accepted manifest is one declaring every parameter an object. The single working workaround is
monkeypatching an undocumented classmethod:

```python
node.node_type.tool_info = classmethod(lambda cls: my_correct_tool)
```

The validator should compare against resolved types, and arguably should treat an explicit manifest
as authoritative over inference rather than as something to cross-check.

## Two related defects, independent of PEP 563

Present even without the `__future__` import:

| Annotation | Derived | Expected |
|---|---|---|
| `Literal["content", "files"]` | `{"type": "object"}` | `{"type": "string", "enum": [...]}` |
| `List[str] \| None` | `{"anyOf": [{"type": "object"}]}` | `{"type": "array", "items": {"type": "string"}}` |

Note the `Tool(parameters=<json schema>)` path handles `enum`, `anyOf` and arrays correctly — the
representation is fine. Only signature inference and the manifest validator are at fault.

## Repro

```python
from __future__ import annotations          # <-- the only line that matters
import railtracks as rt

def search_files(pattern: str, limit: int = 10) -> str:
    """Search files.

    Args:
        pattern: Regex to search for.
        limit: Max results.
    """
    return ""

for p in rt.function_node(search_files).node_type.tool_info().parameters:
    print(p.name, "->", p.param_type)
```

```
pattern -> object      # expected: string
limit   -> object      # expected: integer
```

Delete the first line and both are correct.

CoronRing:
Closes #1458.

`ParameterType.from_python_type` is an identity lookup over 8 types with `OBJECT` as
the default. Anything it doesn't recognise silently becomes `{"type": "object"}`
which under `from __future__ import annotations` is everything. This PR also fix a major bug where, if any unrecognized class is resolved to an object, it cannot be corrected with a manual manifest.

### Fixes

| annotation | before | after |
|---|---|---|
| `pattern: str` under PEP 563 | `{"type": "object"}` | `{"type": "string"}` |
| `Literal["content", "files"]` | `{"type": "object"}` | `{"type": "string", "enum": [...]}` |
| `List[str] \| None` | `{"anyOf": [{"type": "object"}]}` | `{"type": "array", "items": {"type": "string"}}` |
| correct `ToolManifest` under PEP 563 | `NodeCreationError` | accepted |

### Changes

- **`llm/tools/annotations.py`** (new): `resolved_signature()` evaluates string/`ForwardRef`
  annotations. Fast-paths to plain `inspect.signature` when nothing is deferred, so no
  behaviour change for normal code. Unresolvable names warn and pass through.
- **`llm/tools/parameter_handlers.py`**: `UnionParameterHandler` and
  `SequenceParameterHandler` now recurse through the handler chain so nested generics keep their structure. 
- **`llm/tools/tool.py`**: two call-site swaps
- **`validation/node_creation/validation.py`**: resolve annotations before validating.
  Manifest type mismatches now warn instead of raising

### For review

- **One output change for code that already worked:** `Optional[str]` was
  `{"anyOf": [{"type": "string"}]}`, now `{"type": "string"}`. Single-branch `anyOf` is
  noise to a provider and optionality already rides on `required`. Nothing in-repo
  depended on the old shape.
- `_flatten_union_options` is required, not cosmetic: `UnionParameter.__init__` raises
  on a nested union, newly reachable now that union arms recurse (`Union[Tuple[str, int], bool]`).


docs/documentation/agent_design/tools/function_tools.md
```diff
@@ -56,4 +56,48 @@ Agent = rt.agent_node("agent", llm=..., tool_nodes=[fn_a, fn_b])
...
+
+Anything Railtracks cannot recognise falls back to `{"type": "object"}`.
+
+## Overriding the inferred schema
+
+Pass a `ToolManifest` when you want to declare the schema yourself instead of relying
+on inference:
+
+```python
+import railtracks as rt
```
soulFood5632: lets huck this in the scripts folder

packages/railtracks/src/railtracks/llm/tools/annotations.py
```diff
@@ -0,0 +1,123 @@
...
+    if isinstance(annotation, str):
+        expression = annotation
+    elif hasattr(annotation, "__forward_arg__"):
+        expression = annotation.__forward_arg__
+    else:
+        return annotation
+
+    try:
+        # typing.get_type_hints()'s exact behaviour
+        return eval(expression, dict(globalns), dict(localns))  # noqa: S307
```
Amir-R25, line 70: I think this only resolves the outer annotation. I tried `list["Payload"]` the outer string becomes `list["Payload"]`, but the inner `"Payload"` stays unresolved, so the item schema falls back to object. `list[Payload]` works correctly.

```python
from railtracks.llm.tools.tool import Tool
class Payload(BaseModel):
    value: str

def takes_payloads(items: list["Payload"]):
    pass

tool = Tool.from_function(takes_payloads)
print(tool.parameters)
```

packages/railtracks/src/railtracks/llm/tools/annotations.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    if isinstance(annotation, str):
+        expression = annotation
+    elif hasattr(annotation, "__forward_arg__"):
+        expression = annotation.__forward_arg__
+    else:
+        return annotation
+
+    try:
+        # typing.get_type_hints()'s exact behaviour
```
Amir-R25, line 69: I don’t think this is quite the same as get_type_hints(), since that recursively resolves nested forward references, see below

packages/railtracks/src/railtracks/llm/tools/parameter_handlers.py
```diff
@@ -142,17 +197,16 @@ def create_parameter(
...
-                        f"Option {idx} of tuple",
-                        True,
+                    build_parameter(
+                        param_name=f"{param_name}_tuple_option_{idx}",
+                        param_annotation=t,
+                        description=f"Option {idx} of tuple",
                     )
                 )
             # Create UnionParameter to capture all possible tuple element types
             return UnionParameter(
```
Amir-R25, line 207: representing the tuple itself as a union seems to lose the container type. I tried `Union[Tuple[str, int], bool]` and got `str | int | bool,` so the generated schema accepts scalar strings and integers instead of a tuple.

packages/railtracks/src/railtracks/llm/tools/parameter_handlers.py
```diff
@@ -236,3 +333,65 @@ def create_parameter(
             description=description,
             required=required,
         )
+
+
+def _flatten_union_options(options: List[Parameter]) -> List[Parameter]:
```
Amir-R25, line 338: from my comment above, this is where the flattening is happening for a tuple inside the union

packages/railtracks/src/railtracks/validation/node_creation/validation.py
Amir-R25, line 50: Could we use the resolved field annotation from `Pydantic’s model_fields` here? If that’s outside the scope of this PR, I think it’s worth capturing as a follow-up.

packages/railtracks/src/railtracks/validation/node_creation/validation.py
```diff
@@ -355,17 +441,9 @@ def _check_manifest_params_exist_in_function(
...
-            != manifest_params[param_name]
-        ):
-            raise NodeCreationError(
-                message=f"Type mismatch for parameter '{param_name}': function expects '{ParameterType.from_python_type(func_params[param_name].annotation)}', but manifest specifies '{manifest_params[param_name]}'.",
-                notes=[
-                    "Ensure the parameter types in the tool manifest match the function signature.",
-                    "Refer to the ParameterType enum for valid types.",
-                ],
-            )
+        _warn_on_type_mismatch(
```
Amir-R25, line 444: Just checking that this broader behaviour change is intentional: a known mismatch such as a function annotated with int and a manifest declaring string now passes with a warning.

`_infer_param_type()` already distinguishes unreliable inference by returning `None`, so another option would be to skip validation for unknown/object fallbacks while continuing to reject mismatches for types we inferred confidently. Do we want explicit manifests to override even reliable annotations?

packages/railtracks/src/railtracks/llm/tools/parameter_handlers.py
```diff
@@ -77,6 +78,17 @@ def create_parameter(
                 )
             )
 
+        options = _flatten_union_options(options)
+
+        if len(options) == 1:
```
Amir-R25, line 83: This eels like a code smell: it mutates an already-created parameter to repair its name, description, and requiredness. Could we construct the parameter with the correct metadata from the start, or move this normalization into a dedicated helper?

Amir-R25 (changes requested):
Thanks, good overall direction. A few edge cases and code styles need a little work.

A few were happening before but this PR is a good place to address.

I left a few comments:
- Nested forward references such as list["Payload"] are still not fully resolved.
- Tuple handling loses the tuple container and turns its element types into top-level union options.
- I also had a question about changing all manifest type mismatches from errors to warnings, rather than only relaxing validation when inference is unreliable.

One general readability note: there are quite a few inline comments that narrate what the next line is doing. I’d trim those and keep comments for non-obvious constraints or reasoning. A few comments also overstate the behaviour, for example, `“get_type_hints()’s exact behaviour”` and `“nested generics keep their structure” so simplifying them would make the implementation easier to review and maintain.

packages/railtracks/src/railtracks/llm/tools/parameter_handlers.py
```diff
@@ -142,17 +197,16 @@ def create_parameter(
...
-                        True,
+                    build_parameter(
+                        param_name=f"{param_name}_tuple_option_{idx}",
+                        param_annotation=t,
+                        description=f"Option {idx} of tuple",
                     )
                 )
             # Create UnionParameter to capture all possible tuple element types
             return UnionParameter(
                 name=f"{param_name}_tuple_options",
```
Aryan-Railtown, line 208: Not sure why we're appending _tuple_options to the parameter name here. This changes the schema property name away from the real function parameter name, so a tool call built from it won't match the function signature.
```
def locate(coords: Tuple[float, float]) -> None:
    pass


tool = Tool.from_function(locate)
print(tool.parameters[0].name)
# Expected: "coords"
# Actual: "coords_tuple_options" -- a tool call built from this schema would
# pass "coords_tuple_options" as the kwarg, which doesn't exist on `locate`.

```

packages/railtracks/src/railtracks/validation/node_creation/validation.py
```diff
@@ -342,6 +346,88 @@ def validate_tool_params(parameters: Any, param_type) -> bool:
...
+
+    Returns:
+        The inferred schema type, or ``None`` when the annotation carries no
+        reliable type information and therefore constrains nothing.
+    """
+    if annotation is inspect.Parameter.empty or isinstance(annotation, str):
+        # Unannotated, or an annotation we could not resolve
+        return None
+
+    inferred = build_parameter(param_annotation=annotation).param_type
```
Aryan-Railtown, line 386: This builds a throwaway Parameter just to infer a comparison type, but on a schema failure it triggers a warning telling the user to "pass an explicit ToolManifest"... even when the manifest was already supplied.
```
class Payload(BaseModel):
    value: "Undefined"  # noqa: F821 -- forward ref that can never resolve


def process(payload: Payload) -> None:
    pass


manifest = rt.ToolManifest(
    description="Process a payload",
    parameters=[Parameter(name="payload", param_type="object")],
)

rt.function_node(process, manifest=manifest)
# Expected: no warning, since a manifest was already given.
# Actual: prints a UserWarning telling you to "pass an explicit ToolManifest".
```

packages/railtracks/src/railtracks/llm/tools/parameter_handlers.py
```diff
@@ -213,6 +266,50 @@ def create_parameter(
...
+        elif len(value_types) == 1:
+            param_type = value_types[0]
+        else:
+            param_type = value_types
+
+        return Parameter(
+            name=param_name,
+            param_type=param_type,
+            description=description,
+            required=required,
```
Aryan-Railtown, line 308: Should `required` be downgraded here when None is one of the Literal values, the way UnionParameterHandler does for Optional[...]? Right now Literal["a", None] and Optional[Literal["a"]] disagree on required even though they mean the same thing.:
```
def set_mode(mode: Literal["fast", None]) -> None:
    pass


def set_mode_optional(mode: Optional[Literal["fast"]]) -> None:
    pass


literal_tool = Tool.from_function(set_mode)
optional_tool = Tool.from_function(set_mode_optional)

print(literal_tool.parameters[0].required)
print(optional_tool.parameters[0].required)
# Both parameters mean the same thing (mode may be omitted or None), but only
# the Optional[...] form is reported as not required.
```

Aryan-Railtown (changes requested):
Amazing Work Guan!! 
Just 1-2 minor things that i could find

packages/railtracks/src/railtracks/llm/tools/annotations.py
CoronRing: My bad,. `eval("list['Payload']")` returns `list['Payload']` with a bare `str` still sitting in `__args__`. Resolution is now recursive through generics, unions, `Annotated` and `Callable` parameter lists, so `list["Payload"]` and `List[List["Point"]]` both come out fully resolved. 

Thank for noticing,

packages/railtracks/src/railtracks/llm/tools/annotations.py
CoronRing: yeah the comment was wrong, `typing._eval_type` converts nested strings to `ForwardRef`, so the comment claimed the behaviour I had failed to implement. The comment is gone and the code now does the recursive thing.

I kept a hand-rolled walk rather than calling `get_type_hints`, as `get_type_hints` is all or nothing, so a single name that only exists under `TYPE_CHECKING` costs you the schema for every other parameter. 

Added test to be covered by `test_one_unresolvable_parameter_does_not_degrade_the_others` and `test_one_unresolvable_arm_does_not_degrade_the_container`.

packages/railtracks/src/railtracks/llm/tools/parameter_handlers.py
CoronRing: tuple is now an `ArrayParameter` under its own name, with `minItems`/`maxItems` pinned to its length, and `Tuple[X, ...]` left unbounded. `Union[Tuple[str, int], bool]` now gives `anyOf: [array(minItems 2, maxItems 2), boolean]` instead of `str | int | bool`.
 
 bigger than expected. 
 
 Element position is still unconstrained for a heterogeneous tuple, so `["a", "b"]` validates against `Tuple[str, int]`. That needs `prefixItems`, which is draft long time back and wants a provider support check before we emit it, so it is #1582 rather than this PR.

packages/railtracks/src/railtracks/validation/node_creation/validation.py
CoronRing: I added `resolved_signature` two lines above so PEP 563 annotations could not slip past the dict check, then left the model branch walking `__annotations__`, which re-opened the same hole one level down. Measured before the change:

| case | detected |
|---|---|
| `Dict[str, int]` field on a model in a PEP 563 module | no |
| `Dict[str, int]` field inherited from a base model | no |
| `Dict[str, int]` field, plain module, own class | yes |

`model_fields` fixes both.

packages/railtracks/src/railtracks/llm/tools/parameter_handlers.py
CoronRing: Fixed as you suggested. The single-member union path now passes `param_name`, `description` and `required` into `build_parameter` , instead of being built anonymously and then patched. It reads better and it removed the `is_optional` bookkeeping, so the branch is shorter than what it replaced.

nothing was actualy shared or cached, so this was a readability fix rather than a latent bug.

packages/railtracks/src/railtracks/llm/tools/parameter_handlers.py
CoronRing: name and the union modelling both predate this PR, I was only carrying it, but after review I agree it should be updated, the changes should be included in this PR.

packages/railtracks/src/railtracks/llm/tools/parameter_handlers.py
CoronRing: Changed, so the two spellings now agree.

But the fundamental issue with require is not really solved, so I created #1581

`required` in a JSON schema means the key must be present, which is a separate question from whether `None` is an allowed value. `def f(x: Optional[str])` has no default, so `x` must still be passed, and we report it omissible. The `Optional` side is the one that is wrong, and making `Literal` match it spreads the existing behaviour rather than correcting it. Fixing it properly changes every `Optional` parameter without a default,

packages/railtracks/src/railtracks/validation/node_creation/validation.py
CoronRing: `_infer_param_type` now builds inside `degrade_warnings_suppressed()`

packages/railtracks/src/railtracks/validation/node_creation/validation.py
CoronRing: I designed the manifest should always win because it is what reaches the model, but the case splits on hwat `_infer_param_type` already computes: where inference is confident, a contradicting manifest is a user error and silence helps nobody; where inference only reached the `object` fallback, the manifest is the better of the two descriptions and nothing should be compared.

So `int` against a `string` manifest raises again, and a manifest for an annotation we cannot read is taken as given. That keeps the escape hatch open for the bug this PR set out to fix without relaxing check. smo doc update too

CoronRing:
Closing in favour of #1580.

Thanks @Amir-R25 and @Aryan-Railtown for the thorough reviews. Every round found a real edge case, and that is what convinced me this is the wrong shape for the fix.

The core of this PR is a hand-rolled annotation resolver (`annotations.py`) that evaluates strings and rebuilds generic aliases by hand. `typing.get_type_hints(func, include_extras=True)` already handles every case in the table above, including the nested `list["Payload"]` Amir found, and it follows Python's own annotation changes (3.14's `annotationlib`) without us maintaining a parallel implementation. The one thing the custom resolver bought was per-parameter tolerance when a single name can't be resolved, and a clear warning covers that well enough. On top of that the PR grew well past #1458: the tuple rework, recursive union/sequence handling and the `Optional[str]` output change are each their own change.

#1580 takes the stdlib route at a fraction of the size. I've left a review there with what it still needs, mainly the manifest escape hatch and a warning when resolution fails.

Where the rest goes:

- Nested unions (`Optional[List[str]]`, `Optional[Literal[...]]` still degrading to `anyOf: [object]`): #<new issue>
- `required` vs nullable semantics: #1581
- Positional tuple items (`prefixItems`): #1582
- The PEP 563 test module and the manifest / unresolvable-name cases in `test_annotation_resolution.py` are worth carrying over to #1580 or the follow-up.

Issue #1424, opened 2026-08-18
Docs for the pre_verifier/post_verifier middleware


Aryan-Railtown:
### Description

Now that #1485 splits `verifier` into `pre_verifier`/`post_verifier` (both under `railtracks.prebuilt.middleware`, same tier as `Retry`/`Timeout`), write user-facing docs for the pair: how to write an `approve_fn` for each, the `Verdict` shape (accept, accept with comments, decline, decline with comments, and the post-only `result` override), timeout behavior, and how they compose with each other and with other node middleware (placement relative to `Retry` matters).

Follow the existing pattern under `docs/documentation/agent_design/middleware/prebuilt/`: a catalog row in `overview.md` plus one page per middleware (see `list/retry.md` for the format), driven by runnable snippets in `docs/scripts/prebuilt_middleware.py`.

Originally filed as docs for the single `verifier` middleware (#1266, shipped in #1401). Updated here to reflect the #1485 split.

PR #1489, opened 2026-08-25
Split verifier into pre_verifier and post_verifier (#1485)

### LINKED ISSUE #1485: Add post-verifier middleware (gate a call's output, not just its input)
Sub-issue of #1266 (General Verifier).

While exploring #1267 (General HIL structure), we hit a gap: today's `verifier` only gates a call **before** it runs (accept/decline/comment, optionally rewriting the args forwarded to the node). For real LLM control flows you often also need to gate a call's **output** after it has already run, e.g. redacting a tool's returned data before it propagates into an LLM's context, or reviewing the exact text an agent is about to return before it goes further, where declining can't undo the call (it already happened) but can still stop or edit what continues onward.

### Proposal 

- Split into two prebuilt middlewares, same tier as `rt.prebuilt.middleware.Retry`/`.Timeout`:
  - `pre_verifier` - today's `verifier`, unchanged, just renamed/re-homed. No breaking change to behavior.
  - `post_verifier` - new. Runs the approve callable *after* the node's call completes, with access to both the original args and the result. Modeled on the existing `after_node` middleware (`railtracks/middleware/after.py`), with `Verdict` semantics added.
- `Verdict` needs a new field (e.g. `result`) to represent an output override, distinct from `args`/`kwargs` (which are about forwarding into a call, not shaping what came out of one).
- The two compose through the existing `MiddlewareChain` as ordinary middleware list entries, no new chain machinery needed. Where each is placed in `middleware=[...]` relative to other middleware (e.g. `Retry`) controls whether post-review sees every attempt or only the final settled result.

A non-runnable API mockup exists on the `1267-general-hil-structure` branch at `examples/human_in_the_loop/proposal_verifier.py`, and full design notes are on the "General HIL structure (#1267)" Notion page under the "Human in the loop" doc, needs a design conversation with Logan before any of this is built.

Aryan-Railtown:
Closes #1485

Splits `verifier` into `pre_verifier` (unchanged behavior, renamed for pre/post symmetry) and a new `post_verifier` that gates a call's output after it has already run. Both live under `railtracks.prebuilt.middleware`, same tier as `Retry`/`Timeout`. `Verdict`/`VerifierRejectedError` move to `railtracks.middleware`, and `Verdict` is now generic (`Verdict[_R]`) with a `result` field for post-call output overrides, checked against the node's actual return type.

`verifier` was never released publicly, so this is a plain rename with no deprecation shim.

`post_verifier`'s `approve_fn` takes `result` as its first argument, not a trailing keyword, so the whole call is statically checkable via `Concatenate`. Getting that wrong is also validated eagerly, at `post_verifier(...)` call time, raising a clear `TypeError` instead of a confusing crash on first invocation.

## Usage

```python
import railtracks as rt
from railtracks.middleware import Verdict
from railtracks.prebuilt.middleware import post_verifier, pre_verifier


def approve_refund(order_id: str, amount: float) -> Verdict:
    return Verdict(accepted=amount <= 100, comment="over the auto-approve limit")


def redact_ssn(result: dict, order_id: str, amount: float) -> Verdict[dict]:
    scrubbed = {k: v for k, v in result.items() if k != "ssn"}
    return Verdict(accepted=True, result=scrubbed)


@rt.function_node(
    middleware=[
        pre_verifier(approve_refund),  # gate the call BEFORE it runs
        post_verifier(redact_ssn),     # gate the call's OUTPUT after it ran
    ]
)
def process_refund(order_id: str, amount: float) -> dict:
    return {"order_id": order_id, "amount": amount, "ssn": "000-00-0000"}
```

Docs follow-up tracked separately in #1424.

## Test plan
- [x] Unit tests for pre_verifier and post_verifier (accept, decline, timeout, decorator forms)
- [x] Composition test: pre_verifier + post_verifier on one node, and post_verifier placed outside Retry
- [x] Fail-fast validation tests for post_verifier's approve_fn signature
- [x] Full unit + integration suite passing, ruff/format clean, mypy clean on touched files


packages/railtracks/src/railtracks/prebuilt/middleware/post_verifier.py
```diff
@@ -0,0 +1,126 @@
...
+
+def _wrapper(approve_fn: _ApproveFn[_R, _P], timeout: float | None):
+    @functools.wraps(approve_fn)
+    async def wrapped(
+        call: Callable[_P, Awaitable[_R]], *args: _P.args, **kwargs: _P.kwargs
+    ) -> _R:
+        result = await call(*args, **kwargs)
+
+        try:
+            review = unpack_async_sync(approve_fn(result, *args, **kwargs))
```
soulFood5632, line 107: Per our convo, do we really need args and kwargs in the post_verifier?

soulFood5632 (commented):
See comment for more details

packages/railtracks/src/railtracks/prebuilt/middleware/post_verifier.py
Aryan-Railtown: short answer is, i think we do need them, thought about going result-only (matches after_node's shape) but it breaks down for anything side-effecting

two examples:
- send_email(to, subject, body) -> None: result is literally nothing, the whole review has to come from args or there's nothing to review
- delete_file(path) -> bool: result is just True/False, doesn't even say which file

forcing every node to return a "self-describing" object just to be post-reviewable felt like the wrong tradeoff, especially since the highest-value cases for post-review (approving sends/deletes/etc) are usually thin bool/None returns by convention anyway. 
so i suggest we keep the current shape.

packages/railtracks/src/railtracks/prebuilt/middleware/post_verifier.py
```diff
@@ -0,0 +1,126 @@
...
+from railtracks.middleware.core import Middleware, wrap_node
+from railtracks.middleware.verdict import Verdict, VerifierRejectedError
+from railtracks.utils.unpack import unpack_async_sync
+
+logger = logging.getLogger(__name__)
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+_ApproveFn = Callable[Concatenate[_R, _P], Verdict[_R] | Awaitable[Verdict[_R]]]
```
soulFood5632, line 18: Great work with generics. Very clever

packages/railtracks/src/railtracks/prebuilt/middleware/pre_verifier.py
```diff
@@ -3,48 +3,43 @@
...
 @overload
-def verifier(
+def pre_verifier(
     *, timeout: float | None = None, name: str | None = None
-) -> Callable[[_ApproveFn[_P]], Middleware[_P, _R]]: ...
+) -> Callable[[_ApproveFn[_P]], Middleware[_P, Any]]: ...
 
 
-def verifier(
+def pre_verifier(
```
soulFood5632, line 34: I want to cc: @RailtownAI/railtracks-dev, what do we think of the naming convention here for `pre` and `post`. I don't mind it but I think we should have it open to some discussion.

soulFood5632 (commented):
Looks great to me. I have one comment for team discussion and we can merge away.

packages/railtracks/src/railtracks/prebuilt/middleware/pre_verifier.py
Aryan-Railtown: holding the merge, incase a name change is agreed upon.

packages/railtracks/src/railtracks/prebuilt/middleware/post_verifier.py
Aryan-Railtown: a little restrictive, but works :)

packages/railtracks/src/railtracks/prebuilt/middleware/pre_verifier.py
soulFood5632: Lol, it appears it got lost in the shuffle.

packages/railtracks/src/railtracks/prebuilt/middleware/post_verifier.py
```diff
@@ -0,0 +1,126 @@
...
+import functools
+import inspect
+import logging
+from typing import Awaitable, Callable, Concatenate, ParamSpec, TypeVar, overload
+
+from railtracks.middleware.core import Middleware, wrap_node
+from railtracks.middleware.verdict import Verdict, VerifierRejectedError
+from railtracks.utils.unpack import unpack_async_sync
+
+logger = logging.getLogger(__name__)
```
Pooria90: Let's use `logger = get_rt_logger(__name__)` for consistency with code and RT logging format.

packages/railtracks/src/railtracks/prebuilt/middleware/post_verifier.py
```diff
@@ -0,0 +1,126 @@
...
+_R = TypeVar("_R")
+
+_ApproveFn = Callable[Concatenate[_R, _P], Verdict[_R] | Awaitable[Verdict[_R]]]
+
+_POSITIONAL_KINDS = (
+    inspect.Parameter.POSITIONAL_ONLY,
+    inspect.Parameter.POSITIONAL_OR_KEYWORD,
+)
+
+_RED = "\033[91m"
```
Pooria90: I would avoid terminal colors for exception text. This will be purely escape char in log or json.

packages/railtracks/src/railtracks/prebuilt/middleware/pre_verifier.py
```diff
@@ -3,48 +3,43 @@
...
+from typing import Any, Awaitable, Callable, ParamSpec, TypeVar, overload
 
+from railtracks.middleware.core import Middleware, wrap_node
+from railtracks.middleware.verdict import Verdict, VerifierRejectedError
 from railtracks.utils.unpack import unpack_async_sync
 
-from ..middleware.core import Middleware, wrap_node
-from .verdict import Verdict
-
 logger = logging.getLogger(__name__)
```
Pooria90: same here on logging

packages/railtracks/src/railtracks/prebuilt/middleware/post_verifier.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def post_verifier(
+    approve_fn: _ApproveFn[_R, _P] | None = None,
+    /,
+    *,
+    timeout: float | None = None,
+    name: str | None = None,
+) -> Middleware[_P, _R] | Callable[[_ApproveFn[_R, _P]], Middleware[_P, _R]]:
+    """Build a node-verification middleware around ``approve_fn`` that gates a
```
Pooria90, line 64: Just flagging this docstring, because I feel like it's too verbose. No change needed right now; we can rewrite when the docs are settled. Most of the content here is doc-material.

packages/railtracks/src/railtracks/prebuilt/middleware/__init__.py
```diff
@@ -5,11 +5,17 @@
...
 from railtracks.prebuilt.middleware.context_injection import ContextInjection
 from railtracks.prebuilt.middleware.max_calls import MaxCalls, MaxCallsExceededError
+from railtracks.prebuilt.middleware.post_verifier import post_verifier
+from railtracks.prebuilt.middleware.pre_verifier import pre_verifier
 from railtracks.prebuilt.middleware.retry import Retry
 from railtracks.prebuilt.middleware.timeout import Timeout
 
 __all__ = [
     "ContextInjection",
+    "Retry",
```
Pooria90: Duplicate

Pooria90 (dismissed):
Awesome work! I just have minor change requests.

packages/railtracks/src/railtracks/prebuilt/middleware/pre_verifier.py
Aryan-Railtown: chatted with the team, naming looks okay for now. Closing this.

packages/railtracks/src/railtracks/prebuilt/middleware/post_verifier.py
Aryan-Railtown: good catch, i will keep it plain text.

Amir-R25 (approved):
LGTM

PR #1490, opened 2026-08-25
refactor: unify context injection naming

### LINKED ISSUE #1170: [Tech Debt] Clear Naming conventions for Context Injection

In some places `context_injection` is referred to as `prompt_injection`. This should be unified accordingly.,

### Additional Context

_No response_

CoronRing:
Closes #1170

`context_injection` was referred to as `prompt_injection` in several places. Standardizes on "context injection" for the templating feature.

- Rename `llm/prompt_injection_utils.py` → `llm/context_injection_utils.py`
- Rename `utils/prompt_injection.py` → `utils/context_injection.py`
- Rename the matching unit test module and three `test_prompt_injection*` test names
- Rename logger `RT.prompt_injection` → `RT.context_injection`
- Update docs to lead with "context injection"; the concepts page still notes that outside literature uses other terms for it

`SECURITY.md` is left as-is — it uses "prompt injection" in the industry sense of an attack, which is the distinction this rename exists to preserve.

No public API change: `escape_braces` is the only re-export from the package root and keeps its name. Both renamed modules were internal (`railtracks/utils/__init__.py` is empty, so neither was reachable except by direct submodule path).

### DIFF
--- a/docs/documentation/advanced/context.md
+++ b/docs/documentation/advanced/context.md
@@ -37,12 +37,12 @@ In agentic systems, you can use context to store important facts or constraints 
     ```
 
 
-### Prompt Injection
+### Context Injection
 
-One of the most powerful features built on top of the context system is "prompt injection" (also called "context injection"). This allows dynamically inserting values from the global context into prompts for LLMs:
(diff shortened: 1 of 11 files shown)


Amir-R25 (approved):
image

PR #1492, opened 2026-08-25
refactor: isolate the llm package and give LLM failures a catchable taxonomy

### LINKED ISSUE #1096: [Tech Debt] Cross Dependency in Errors module

Currently, the LLM package should never import from anything within the railtracks package as it is its own package. However, the LLMError is currently living on the `railtracks` side and is imported my the `llm` module

```python
class LLMError(RTError):
    """
    Raised when an error occurs during LLM invocation or completion.
    """

    def __init__(
        self,
        reason: str,
        message_history: "MessageHistory" = None,
    ):
        self.reason = reason
        self.message_history = message_history

        message = f"{self._color('LLM Error: ', self.BOLD_RED)}{self._color(reason, self.RED)}"
        super().__init__(message)

    def __str__(self):
        base = super().__str__()
        details = []
        if self.message_history:
            mh_str = str(self.message_history)
            indented_mh = "\n".join(
                "    " + line for line in mh_str.splitlines()
            )  # 2 indents (2-spaces) per indent
            details.append(
                self._color("Message History:\n", self.BOLD_GREEN)
                + self._color(indented_mh, self.GREEN)
            )
        if details:
            notes_str = (
                "\n"
                + self._color("Details:\n", self.BOLD_GREEN)
                + "\n".join(f"  {d}" for d in details)
            )
            return f"\n{self._color(base, self.RED)}{notes_str}"
        return self._color(base, self.RED)
```

This should be refactored into the llm package

### Additional Context

_No response_

CoronRing:
## Summary

`railtracks/llm` is meant to stand on its own, but `_litellm_wrapper` imported `LLMError` and `NodeInvocationError` from `railtracks.exceptions`.

Removing that import surfaced a second problem behind it. `LLMError` was a *sibling* of `NodeInvocationError`, so `except NodeInvocationError` never caught LLM failures, and the boundary raised without `from e` — so `__cause__` was dropped and a timeout was only detectable by substring-matching the message.

This PR fixes both: the `llm` package raises its own errors and never imports upward, and LLM failures become a usable taxonomy on the node side.

## What changed

The `llm` package owns two roots, deliberately unrelated so neither `except` clause can shadow the other:

```
ProviderError                        talking to a model provider
├── ProviderTimeoutError
├── ProviderRateLimitError
├── ProviderAuthenticationError
├── ModelError / ModelNotFoundError
└── RetryError

ToolCreationError                    defining a tool -- a caller bug
```

`railtracks.exceptions` gains the node-facing side, now under `NodeInvocationError`:

```
NodeInvocationError
├── LLMError
│   ├── LLMTimeoutError
│   ├── LLMRateLimitError
│   └── LLMAuthenticationError
└── GuardrailBlockedError
```

litellm exceptions are classified into `Provider*` inside the litellm adapter, and translated once into `LLM*` at `built_nodes/llm/llm_helpers.py`. Every branch chains with `from e`. Users dispatch on two axes with plain `except` clauses:

```python
except LLMTimeoutError:        # back off and retry
except LLMRateLimitError:      # switch tier
except LLMAuthenticationError: # fail fast, fix config
except LLMError:               # any other LLM failure
except NodeInvocationError:    # died for a non-LLM reason
```

Classification looks *through* a `RetryError`, so an exhausted retry of timeouts arrives as `LLMTimeoutError` whether or not a retry approach was configured. `RetryError` keeps its own type for direct llm callers, since that is how they detect exhausted retries.

## Behaviour changes

> **Breaking:** `LLMError` is now a subclass of `NodeInvocationError` -- see the first bullet.

- **`LLMError` is now a `NodeInvocationError`.** `except NodeInvocationError` catches LLM failures where it previously did not. Existing handlers that list `except NodeInvocationError` *above* `except LLMError` make the second clause unreachable — two examples in `docs/scripts/error_handling.py` had exactly this and are reordered here.
- **`__cause__` is populated** on every error the boundary raises.
- **Direct llm-layer callers**: `model.chat_with_tools(...)` with a malformed `Tool` raises `ToolCreationError` instead of `NodeInvocationError`. Through `rt.call` / agent nodes nothing changes. Timeouts, rate limits, and auth failures raised outside a node are now `Provider*` types; anything else still surfaces as litellm's own exception, so existing `except litellm.*` handlers keep working.

## Testing

A parametrized test AST-scans every module under `railtracks/llm/` and fails if any imports a railtracks error module or references `LLMError`, so the layering cannot regress. Boundary tests cover translation, `__cause__` preservation, no-double-wrapping, per-class classification, and the exhausted-retry case.

`2262 passed` (unit), `310 passed` (integration). Remaining errors in both runs are missing optional deps in this environment (`duckdb`, `tavily`, `sqlalchemy`, `mcp`), unrelated to this change.

## Type of change

## Checklist

## Notes

Does not close #1096. Two non-error cross-dependencies remain in the package and are better handled separately:

- `llm/middleware.py` → `railtracks.middleware.core`
- `llm/models/cloud/azureai.py` → `railtracks.utils.deprecation`

The fine axis covers timeout, rate limit, and auth. Anything else — content filtering, context length, malformed responses — still surfaces as plain `LLMError` with the detail on `__cause__`. Adding classes later is non-breaking, so the time to do it is when someone needs to branch on one.


packages/railtracks/src/railtracks/exceptions/_base.py
```diff
@@ -1,16 +1,5 @@
-class RTError(Exception):
-    """
-    A simple base class for all RTExceptions to inherit from.
-    """
+from railtracks.llm._exceptions import RTError
```
soulFood5632: I think I could have been clearer but I think we should have 2 separate exception classes for each section `llm` and `core railtracks`. that being said this may be an ok approach.

soulFood5632 (dismissed):
Overall looks good, wanted to flag a design approach you went with to discuss further

packages/railtracks/src/railtracks/exceptions/errors.py
```diff
@@ -31,48 +47,33 @@ def __str__(self):
...
-    """
-    Raised when an error occurs during LLM invocation or completion.
+    Being a `NodeInvocationError` lets callers branch on both axes at once: catch
+    `NodeInvocationError` for "a node terminated", then test for `LLMError` to ask
+    whether the LLM caused it.
     """
 
     def __init__(
         self,
         reason: str,
```
Amir-R25, line 63: @CoronRing The `LLMError` will have three things now:
- reason
- notes
- message

Amir-R25 (changes requested):
Great overall work, small comment on what we discussed

packages/railtracks/src/railtracks/exceptions/_base.py
CoronRing: Updated so the logic is no longer relavent, resolving for now (feel free to check new design)

packages/railtracks/src/railtracks/exceptions/errors.py
CoronRing: Reason == message, its some some legacy naming where the caller often use the term "message" to describe "reason" which was not really touched in thsi PR. 

One thing we could do is to move notes + reason to RTError as a shared property, so each child do not need to define again in init. Another thing is we should be consistent between reason and message, perhaps replace reason with messages as its more common? Both outscope this ticket tho, let me know your thoughts and i can roll both.

Here message refer to color formatted reason which was better wrapped to be pretty.

CoronRing: After discussion, will resolve item and create a new ticket

CoronRing:
The layering guard has a hole worth closing before this merges — it's a test-only change, so it shouldn't slow the merge down.

`test_llm_package_does_not_import_railtracks_errors` filters escaping imports by `if "exceptions" in imported`. That catches a direct `from railtracks.exceptions import ...`, but not a reach into the framework's errors *through an intermediary*. #1521 is a live example: it adds

```python
from railtracks.validation.node_creation.validation import validate_tool_params
```

to `llm/tools/tool.py`. `validation.py` imports `railtracks.exceptions.errors` and raises `NodeCreationError`. So on a tree with both PRs:

```
tool.py escaping imports found by the helper: ['railtracks.validation.node_creation.validation']
after the "exceptions" filter:                []
test_exceptions.py: 56 passed
```

while at runtime:

```
Tool("t","d",{"type":"object","properties":{...}})
  -> NodeCreationError, mro: (NodeCreationError, RTError, Exception, BaseException)
```

An `RTError` raised from inside `railtracks.llm`, with the guard green. The hierarchy tests don't catch it either — they assert `issubclass` over a fixed list of classes, never what actually gets raised.

CoronRing:
Pushed the guard change plus the merge with `main` — branch is now `MERGEABLE` (`8afa558`).

**Conflict resolution.** One hunk, an import line where two independent changes collided: `main` deprecated `CohereLLM` (#1497) and this branch renamed `RTLLMError` to `ProviderError`. Kept both:

```python
from railtracks.llm import AnthropicLLM, GeminiLLM, HuggingFaceLLM, OpenAILLM
from railtracks.llm._exceptions import ProviderError
```

The two `CohereLLM` parametrize entries #1497 also dropped had already auto-merged out, and `llm/__init__.py` merged clean with no resurrected export. Checked the tree for stray `Cohere` and `RTLLMError` references — none left.

**Guard change**, as described above: `test_llm_package_does_not_import_upward` now asserts on every escaping import rather than filtering for `"exceptions"`, with `KNOWN_ESCAPES` covering the two #1096 cross-deps, and `test_known_escapes_are_all_still_real` so a fixed entry can't sit stale and silently widen the guard. Test-only, no source changes.

**Verified on the merged tree:**

- Unit: 1752 passed, 4 skipped — `test_provider_llm.py` now runs rather than being excluded.
- Integration: 314 passed, 2 skipped. The two `rt_mcp` errors are the fixture failing to `pip install mcp>=1.9.0` in my sandbox, matching the optional-dep noise in your Notes.
- `ruff check` and `ruff format --check` clean.
- Guard checked both ways: green as-is, and fails with `tools/tool.py imports ['railtracks.validation.node_creation.validation']` when #1521's import is applied.

Review is still marked `CHANGES_REQUESTED` from before, so it needs a re-approve to unblock.

packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
```diff
@@ -80,13 +118,38 @@ async def llm_invoke(
                 returned_mess = await model_invoker.invoke(
                     message_history, schema=schema, tools=tools
```
Amir-R25, line 119: a nit on this file would be too many inline comments, but that shouldn't block the merge of this

Amir-R25 (dismissed):
Looks good, thanks Guan!

Amir-R25 (approved):
image

PR #1494, opened 2026-08-25
improve args parsing to handle descriptions on next lines and colons

### LINKED ISSUE #1491: `Args:` section is dropped when a parameter's description starts on the next line
### Description

`extract_args_section` in `packages/railtracks/src/railtracks/llm/tools/docstring_parser.py` terminates the `Args:` section on any line ending in `:`. A parameter whose description starts on the following line looks exactly like a new section header, so the section is cut short and if it's the first parameter, the whole section is lost.

```python
from railtracks.llm.tools.docstring_parser import parse_docstring_args

docstring = """
Fetches a URL.

Args:
    headers:
        Optional dict of extra headers.
    timeout: Optional request timeout.

Returns:
    The response body.
"""

parse_docstring_args(docstring)
```

Actual:

```python
{}
```

Expected:

```python
{'headers': 'Optional dict of extra headers.', 'timeout': 'Optional request timeout.'}
```

Every parameter description on the tool schema is silently empty for docstrings written this way; a valid and reasonably common Google style for long descriptions.

The guard that was meant to prevent this is dead code:

```python
if (
    line.strip()
    and line.strip().endswith(":")
    and not line.strip().startswith(" ")   # always True — the string is already stripped
):
    break
```

The section break should be decided on the *raw* line's indentation: a real section header (`Returns:`, `Raises:`) sits at the same indent as `Args:`, whereas a parameter line is indented deeper. Comparing against the indent of the `Args:` line rules out the false positive without needing a keyword allowlist.

### Related: indent comparison in `parse_args_section`

While in this file: `parse_args_section` uses `indent != param_indent` to decide that a matching line is continuation text rather than a new parameter. A parameter indented *less* than the first one is then swallowed:

```
Args:
        headers: extra headers
    timeout: timeout
```

gives `{'headers': 'extra headers timeout: timeout'}` instead of two parameters.

`indent > param_indent` is the safer rule: silently losing a real parameter's description is worse than keeping a stray `Note:`-style key, which downstream discards anyway (`Tool.from_function` looks descriptions up by signature parameter name).

The two are coupled: fixing `extract_args_section` makes next-line descriptions reach `parse_args_section` for the first time, and that is precisely the shape where the `!=` rule can anchor `param_indent` to a continuation line.

```python
from railtracks.llm.tools.docstring_parser import parse_docstring_args

docstring = """
Fetches a URL.

Args:
    headers:
        Optional dict of extra headers.
    timeout: Optional request timeout.

Returns:
    The response body.
"""

parse_docstring_args(docstring)
```

Pooria90:
## Summary

`extract_args_section` ended the `Args:` block at any line ending in `:`, so two common docstring shapes were silently mis-parsed: a parameter whose description starts on the next line (`headers:` with the text below it) and a description that merely ends in a colon (`mapping: Maps keys to values as follows:`). In both cases the section was truncated at that line — for a first parameter, the entire block was dropped and every parameter description came back empty. The section boundary is now decided by indentation (a following section is indented less than the parameter lines) with a section-name check for docstrings that put their headers at parameter level or leave `Args:` empty. `parse_args_section` accepts an empty inline description so `param:` with the text on the following line parses, and compares indents with `>` rather than `!=` so a parameter indented less than the first one is no longer swallowed as continuation text.

Verified against every docstring in the package: of 290 with an `Args:` section, 9 parse differently — 24 parameter descriptions recovered, none lost, no phantom keys. `SQLLoader` alone goes from 2 of 9 parameters described to all 9; `connect_mcp`, `Message.__init__`, `SemanticChunker`, and the Azure/Ollama/provider wrappers all recover parameters that previously parsed to nothing.

closes #1491

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

Follows #1480, which fixed the inverse case (a `Note:` continuation line minting a phantom parameter). That fix is unchanged and still covered.

Before / after on the reported shape:

```python
docstring = """
Fetches a URL.

Args:
    headers:
        Optional dict of extra headers.
    timeout: Optional request timeout.

Returns:
    The response body.
"""

# before
{}
# after
{'headers': 'Optional dict of extra headers.', 'timeout': 'Optional request timeout.'}

### CODE REVIEW

File: packages/railtracks/src/railtracks/llm/tools/docstring_parser.py [hunk 1]
```diff
@@ -10,6 +10,48 @@
 
 from .parameters import Parameter, ParameterType
 
+# Section headers that can follow an "Args:" block and therefore end it.
+_SECTION_HEADERS = frozenset(
```
Amir-R25, line 14: As we discussed, let's just have these headings: `Args`, `Returns`, `Yields`, `Raises`, `Attributes`

Amir-R25 (changes requested):
small bug where `notes:` is taken as a section heading rather than a parameter:

```
Args:
    notes:
        Notes attached to the record.
    value: The value.
```

Amir-R25 (approved):
Thanks!

Issue #1496, opened 2026-08-26
Let input guards fire once per agent call instead of per model round-trip


Pooria90:
### Description

(Depends on the async guardrail task.)

Guards are middleware on `ModelInvoker`, so they fire once per **model round-trip**, not once per agent call. `OutputGuard` skips intermediate tool turns (`concrete.py:184`); `InputGuard` has no equivalent, so it re-runs every iteration of the tool-calling loop.

That is nearly free for the regex/length rails we ship today, and arguably correct (re-screening tool output catches injection). It stops being free once a guard can make its own LLM call.

Requirements:
- Add `once: bool = False` to `InputGuard.__init__`, threaded through the `input_guard(...)` decorator and `_make_guard`. Default `False` keeps today's re-screening behavior.
- When `once=True` and the guard has already fired this agent invocation, `_middleware_fn` forwards to `call(...)` unguarded.

PR #1499, opened 2026-08-26
1466 async guardrails

### LINKED ISSUE #1466: Guardrail function needs to be async
### Description

We should support sync guardrail function definition. Currently, they become part of the `_middleware_fn` which is async; but users don't have a proper way to define an async guardrail function. So, we technically can't call another agent inside a guardrail.

### LINKED ISSUE #1495: Make the guardrail evaluation path async
### Description

`BaseLLMGuardrail.run` → `_eval_one_rail` → `self(event)` is synchronous, sitting between two async layers. 

Make the eval path async and accept sync-or-async guards, matching the existing `verifier` / `before_llm` pattern (`unpack_async_sync`). All six prebuilt guards and every existing user guard must keep working unchanged.

Pooria90:
## Summary

Guardrails could only be authored as synchronous functions, so a rail could never `await rt.call(...)` and delegate its decision to another agent. The eval path (`BaseLLMGuardrail.run` → `_eval_one_rail` → `self(event)`) was synchronous while sitting between async middleware layers.

This makes the eval path async and accepts sync-or-async guards, following the `unpack_async_sync` pattern already used by `verifier` and `before_llm`/`after_llm`. Every existing synchronous guard, including all six prebuilt rails, keeps working untouched. Adds `adecide()` as the async counterpart to `decide()`, which now raises a clear `TypeError` on an async guard instead of handing back an un-awaited coroutine.

Closes #1495.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Example
```python
class SafetyReport(BaseModel):
    safe_request: bool
    reason: str

Judge = rt.agent_node(
    name="safety-judge",
    output_schema=SafetyReport,
    system_message="You decide whether a user request is safe to answer.",
)

@rt.input_guard(name="llm_judge")
async def llm_judge(event: LLMGuardrailEvent) -> GuardrailDecision:
    verdict = await rt.call(Judge, user_input=str(event.messages[-1].content))
    if not verdict.structured.safe_request:
        return GuardrailDecision.block(
            reason=f"The judge flagged this request: {verdict.structured.reason}",
            user_facing_message="I can't help with that.",
        )
    return GuardrailDecision.allow()
```

## Notes
1. Subclassing works the same way: define `async def __call__(self, event)`.

2. Cost characteristic, documented but not fixed here. Guards are model middleware, so a rail fires once per model round-trip, not once per agent call. OutputGuard skips intermediate tool-call turns; InputGuard has no equivalent. **This will be handled in another PR** (RE: https://github.com/RailtownAI/railtracks/issues/1496). 

3. Fixed a doc and interface misalignment: Re-exported `InputGuard` and `OutputGuard` classes at `rt.guardrails` level to match the docs and make the interface easier (previously users had to import from `rt.guardrails.llm.concrete`

### DIFF
--- a/docs/documentation/agent_design/middleware/guardrails/overview.md
+++ b/docs/documentation/agent_design/middleware/guardrails/overview.md
@@ -58,8 +58,22 @@ Attach them like any model middleware:
 --8<-- "docs/scripts/custom_guardrails.py:decorator_attach"
 ```
 
-!!! note "Guard functions are synchronous"
-    A guard is called synchronously while the rail is evaluated, so the decorated function must be a regular `def`, not `async def`.
+### Async guards
+
+A guard may be a regular `def` or an `async def`. An async rail is awaited while it is evaluated, so it can `await` anything, including `rt.call` into another agent. That is how you build an LLM-judge rail, where the decision is delegated to a second agent:
+
+```python
+--8<-- "docs/scripts/custom_guardrails.py:decorator_async"
+```
+
+Give the judge an `output_schema` rather than parsing its prose: the rail then branches on a typed field instead of substring-matching a reply that can be reworded at any time, and the judge's own `reason` carries straight into the block.
+
+The same holds when subclassing: define `async def __call__(self, event)` and everything downstream of the decision is identical.
+
+!!! warning "An async rail runs once per model round-trip"
+    Guards are model middleware, so a rail fires on every model call, not once per agent call. On a tool-calling agent that means once per iteration of the tool loop: an input rail on an agent that takes three tool turns fires four times, each seeing a longer history. That is free for a regex check but not for a rail that makes its own LLM call. Output rails are exempt from intermediate tool-call turns and fire only on the final reply.
+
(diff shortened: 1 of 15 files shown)


soulFood5632 (approved):
LGTM

docs/documentation/agent_design/middleware/guardrails/overview.md
```diff
@@ -58,8 +58,22 @@ Attach them like any model middleware:
...
+
+```python
+--8<-- "docs/scripts/custom_guardrails.py:decorator_async"
+```
+
+Give the judge an `output_schema` rather than parsing its prose: the rail then branches on a typed field instead of substring-matching a reply that can be reworded at any time, and the judge's own `reason` carries straight into the block.
+
+The same holds when subclassing: define `async def __call__(self, event)` and everything downstream of the decision is identical.
+
+!!! warning "An async rail runs once per model round-trip"
```
Amir-R25, line 73: Nice distinction!

Amir-R25 (dismissed):
Le' GTM

soulFood5632 (dismissed):
LGTM

soulFood5632 (approved):
LGTM

PR #1511, opened 2026-08-28
feat(middleware): rename before/after hooks to pre/post with deprecation shims

### LINKED ISSUE #1510: Naming consistency: rename before/after middleware to pre/post (before_llm, after_llm, after_node)
## Context

PR #1489 (following the #1485 discussion) split `verifier` into `pre_verifier` / `post_verifier` under `railtracks.prebuilt.middleware`. During review, a naming-convention mismatch came up: the rest of the middleware system uses `before`/`after`, not `pre`/`post`:

- `packages/railtracks/src/railtracks/built_nodes/llm/middleware/before_llm.py:50` — `before_llm(...)`, a `model_middleware=` hook that runs before the raw model call.
- `packages/railtracks/src/railtracks/built_nodes/llm/middleware/after_llm.py:36` — `after_llm(...)`, the symmetric post-call hook at the LLM-call level.
- `packages/railtracks/src/railtracks/middleware/after.py:30` — `after_node(...)`, a node-level (`middleware=`) post-call hook.

Note: there's no symmetric `before_node` — the generic node-level pre-hook role is already filled by `wrap_node` itself, so no rename target is needed there.

`pre_verifier`/`post_verifier` is a bit of a special case (it's a ready-made `prebuilt` HIL/governance middleware, not a generic hook-builder primitive like `before_llm`/`after_llm`/`after_node`), but the `pre`/`post` *naming pattern* it introduced arguably reads better and should extend to the rest of middleware for consistency, regardless of that distinction.

## Proposed direction (starting point, not a locked spec)

- `before_llm` -> `pre_llm`
- `after_llm` -> `post_llm`
- `after_node` -> `post_node`

## Backward compatibility

Unlike `verifier` (never released before #1489), `before_llm`/`after_llm`/`after_node` have presumably been available for a while, so a hard rename would be breaking. The old names should probably keep working via a thin deprecation shim (e.g. delegate to the new name and emit a `DeprecationWarning`) rather than being removed outright. Whoever picks this up should first check actual release/usage status to decide how strict the deprecation needs to be (e.g. whether a hard cutover is acceptable, what timeline, one PR vs. staged).

## Docs

`before_llm`/`after_llm`/`after_node` are currently referenced in:
- `docs/documentation/agent_design/middleware/overview.md`
- `docs/documentation/agent_design/middleware/custom.md`
- `docs/scripts/middleware.py`

These should be updated to the new names once the rename lands.

## Scope note

This is intentionally exploratory/minimal — a good-first-issue-shaped ticket, not a fully-designed spec. The exact deprecation mechanics, timeline, and whether to do the rename in one PR or stage it are left to whoever picks it up.

ceasermikes002:
## Summary

Renames the middleware hook primitives to use the consistent `pre_*` / `post_*` naming convention:

* `before_llm` → `pre_llm`
* `after_llm` → `post_llm`
* `after_node` → `post_node`

The existing names remain available as deprecated shims for backward compatibility and emit a `FutureWarning` pointing users to the new names.

Closes #1510.

## Type of change

* [ ] Bug fix
* [x] Feature
* [ ] Breaking change
* [ ] Docs
* [ ] Refactor / chore / build / tests

## Checklist

* [x] Lint & format pass (`ruff check . && ruff format .`)
* [x] Tests added/updated and pass locally (`pytest tests`)
* [x] Docs updated if user-facing behavior changed
* [ ] Breaking changes include migration notes

## Notes

The new `pre_llm`, `post_llm`, and `post_node` APIs are re-exported from the relevant public modules.

Deprecated `before_llm`, `after_llm`, and `after_node` continue to work through lightweight delegation shims, allowing existing users to migrate without an immediate breaking change.

Internal framework usage has been migrated to the new names so normal framework execution does not emit deprecation warnings.

Added/updated tests cover the new APIs, deprecated API warnings, and parameterized decorator forms.

### DIFF
--- a/packages/railtracks/src/railtracks/llm/tools/tool.py
+++ b/packages/railtracks/src/railtracks/llm/tools/tool.py
@@ -49,12 +49,23 @@ class Tool:
         if (
             isinstance(parameters, dict) and len(parameters) > 0
         ):  # if parameters is a JSON-output_schema, convert into Parameter objects (Checks should be done in validate_tool_params)
-            props = parameters.get("properties")
+            props = parameters.get("properties") or {}
             required_fields = list(parameters.get("required", []))
+            if not props and required_fields:
+                raise ToolCreationError(
+                    f"Tool {name!r}: schema declares required fields "
+                    f"{required_fields} but has no 'properties' block.",
+                    notes=[
+                        "Add a 'properties' entry describing each required field.",
+                        "A schema with no parameters should omit 'required' entirely.",
+                    ],
+                )
             param_objs: List[Parameter] = []
-            for name, prop in props.items():
+            for param_name, prop in props.items():
                 param_objs.append(
-                    parse_json_schema_to_parameter(name, prop, name in required_fields)
+                    parse_json_schema_to_parameter(
+                        param_name, prop, param_name in required_fields
+                    )
                 )
             parameters = param_objs
 
--- a/packages/railtracks/tests/unit_tests/llm/tools/test_tool.py
+++ b/packages/railtracks/tests/unit_tests/llm/tools/test_tool.py
@@ -1,0 +1,45 @@
+"""
+Tests for the Tool class.
+
+This module contains tests for railtracks.llm.tools.tool.Tool.
+"""
+
+import pytest
+from railtracks.llm.tools import Tool
+from railtracks.llm.tools.tool import ToolCreationError
+
+
+class TestToolFromSchemaDict:
+    def test_keeps_the_given_name(self):
+        schema = {
+            "type": "object",
+            "properties": {
+                "city": {"type": "string"},
+                "date": {"type": "string"},
+            },
+            "required": ["city"],
+        }
+
+        tool = Tool(
+            name="Weather_1_GetWeather", detail="Get the weather.", parameters=schema
+        )
+
+        assert tool.name == "Weather_1_GetWeather"
+        assert {p.name for p in tool.parameters} == {"city", "date"}
+        assert {p.name for p in tool.parameters if p.required} == {"city"}
+
+    def test_schema_without_properties(self):
+        tool = Tool(
+            name="no_args", detail="Takes nothing.", parameters={"type": "object"}
+        )
+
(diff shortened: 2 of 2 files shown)


Amir-R25 (changes requested):
Thanks @ceasermikes002 for the quick turnaround on this. Looks good from my side, seems like a small conflict with a PR that landed that needs addressing when you have a chance

packages/railtracks/src/railtracks/built_nodes/llm/middleware/before_llm.py
```diff
@@ -57,64 +52,15 @@ def before_llm(
...
-        return wrapper
-
+) -> Any:
+    """Deprecated: Use ``rt.pre_llm`` instead."""
+    warn_pending_change(
+        "rt.before_llm",
+        change="is renamed",
+        instead="rt.pre_llm",
+        detail="The function itself is unchanged.",
+        stacklevel=2,
```
Aryan-Railtown: stacklevel=2 in before_llm/after_llm/after_node's deprecation call is off by one, should be the default (3, like rt.interactive uses). Right now every warning gets attributed to the internal call site inside before_llm.py/etc, not the user's actual line. can you double check this once?

Aryan-Railtown (changes requested):
just this one small comment. other than that looks amazing. Good job :)

Aryan-Railtown:
one more minor thing noticed while poking at this -- if you have 2+ `before_llm`/`after_llm`/`after_node` calls in a script, only the first one ever shows a warning (same root cause as the stacklevel comment above -- they all collapse onto the same location so python's own dedup swallows the rest). not blocking imo, users still get at least one warning telling them to migrate -- just flagging in case it's not expected.

packages/railtracks/src/railtracks/built_nodes/llm/middleware/before_llm.py
ceasermikes002: Updated before_llm, after_llm, and after_node to use the default stacklevel=3 so warnings point directly to the user's line :)

PR #1513, opened 2026-08-28
fix: keep tool name when parameters come from a JSON-schema dict

### LINKED ISSUE #1512: Tool built from a JSON-schema dict takes the name of its last property
`Tool.__init__` uses `name` as the loop variable when converting a JSON-schema dict, shadowing the `name` argument. Every tool built from a schema dict ends up named after the last property in `properties`.

https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/llm/tools/tool.py#L49-L65

```python
for name, prop in props.items():   # shadows the `name` argument
    ...
self._name = name                  # last property key, not the tool name
```

Repro:

```python
schema = {
    "type": "object",
    "properties": {"city": {"type": "string"}, "date": {"type": "string"}},
    "required": ["city"],
}
Tool(name="Weather_1_GetWeather", detail="...", parameters=schema).name
# -> 'date'   (expected 'Weather_1_GetWeather')
```

Tool calling cannot work under this: the name sent to the provider does not match any registered tool.

Second defect in the same block: `props = parameters.get("properties")` returns `None` when the key is absent, so `Tool(name="X", detail="...", parameters={"type": "object"})` raises `AttributeError: 'NoneType' object has no attribute 'items'`.

Fix: rename the loop variable and default `props` to `{}`.

CoronRing:
Closes #1512

`Tool.__init__` used `name` as the loop variable when converting a schema dict, so the tool took the name of its last property. Renamed the loop variable, and defaulted a missing `properties` to `{}` instead of raising `AttributeError`.

Adds a regression test for both cases.


packages/railtracks/src/railtracks/llm/tools/tool.py
```diff
@@ -49,12 +49,16 @@ def __init__(
         if (
             isinstance(parameters, dict) and len(parameters) > 0
         ):  # if parameters is a JSON-output_schema, convert into Parameter objects (Checks should be done in validate_tool_params)
-            props = parameters.get("properties")
+            props = parameters.get("properties") or {}
```
Amir-R25, line 52: `validate_tool_params` at `validation/node_creation/validation.py:299` already checks for this (asserts `"properties" in parameters`), but grepping the repo it's never actually called. The comment on line 51 says "Checks should be done in validate_tool_params" that function is not used in our framework.

So the `AttributeError` this PR removes is the only line of defense on the direct-`Tool(...)` path.

Two options:
1. Raise here when `properties` is missing/None but `required` isn't empty this way we will keep the no-arg case working.
2. Actually wire up `validate_tool_params` on the `Tool.__init__` dict path, which is what the existing comment already promises.

```python
props = parameters.get("properties")
required_fields = list(parameters.get("required", []))
if not props:
    if required_fields:
        raise ValueError(
            f"Tool {name!r}: schema declares required fields "
            f"{required_fields} but has no 'properties' block"
        )
    props = {}
```

We don't expose `Tool` to users directly, thought I'd make a comment to track as a separate issue.

Amir-R25 (dismissed):
Looks good, if this unblocks you it's good to go, I made a separate issue for something related #1513

CoronRing:
@Amir-R25 Great point, I was trying to get the fix asap did not notice the item. Updated with your suggestion!

Amir-R25 (approved):

PR #1518, opened 2026-08-31
fix: remove dead validate_tool_params, unblock stale comment

### LINKED ISSUE #1514: Dead code: `validate_tool_params` is defined but never called
### What
Came up in reviewing #1512

[`validate_tool_params`](https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/validation/node_creation/validation.py#L299) in `validation/node_creation/validation.py` is defined but never invoked anywhere in the codebase.

The only reference outside the definition is a comment in [`Tool.__init__`](https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/llm/tools/tool.py#L49-L51) claiming these checks _should be_ done in `validate_tool_params` but nothing imports or calls it.

### Problem

`validate_tool_params` performs real checks on JSON-schema dicts passed to `Tool(parameters=<dict>)`:

- `"type"` key exists and equals `"object"`
- `"additionalProperties"` is present and set to `False`
- `"properties"` key exists

Because it's never called, the direct `Tool(name=..., parameters=<schema-dict>)` path has zero validation. Malformed schemas go through silently and produce a zero-arg tool, and the failure surfaces much later as a confusing provider error at LLM-call time. `additionalProperties: False` and a present `"properties"` key are hard requirements for OpenAI structured outputs, so this isn't hypothetical.

Note: the one check called out in the original review (a `required` list naming a field not in `properties`) was already inlined at [`tool.py:52-59`](https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/llm/tools/tool.py#L52-L59) in #1513. The other three checks above still have no enforcement anywhere.

### Suggested fix

Wire `validate_tool_params` into `Tool.__init__` so the checks the comment already promises actually run, with two small adjustments so it matches `Tool`'s current signature:

1. **Accept lists / iterables in the type guard.** `Tool.parameters` is typed `Iterable[Parameter] | Dict[str, Any] | None`, so the outer `isinstance(parameters, (set, dict))` check must also allow lists (or any iterable of `Parameter`). Existing callers pass `parameters=[]` in `test_llm_helpers.py`, `test_model_invoker.py`, and `test_node_builder.py` — these should keep working.

2. **Update the tests that pass minimal / non-compliant schemas.** `test_schema_without_properties` and similar cases exist today only because no validation runs; once the checks are wired up, those tests should be updated to pass schemas that satisfy the intended contract (`type: object`, `additionalProperties: False`, `properties: {…}`).

Also delete the stale comment at `tool.py:51` — with the function wired in, it's no longer aspirational.

### Alternative considered

An earlier attempt (#1518) went with the alternative — delete `validate_tool_params` (and its `ExceptionMessageKey` entries) entirely. Smaller diff, but it leaves the three schema checks unenforced: `Tool(parameters={"type": "object"})` still silently produces a zero-arg tool, and `additionalProperties`-less schemas still blow up at LLM-call time instead of at construction. Going with Option 1 here.

cc: @CoronRing

AmirF194:
## Summary

`validate_tool_params` (`validation/node_creation/validation.py`) has no callers
anywhere in the tree. The only reference to it is a comment in `Tool.__init__`
(`tool.py:51`) claiming its checks happen there, but nothing imports or calls the
function.

The issue offers two fixes: wire it up, or delete it and inline the essential
checks. I reproduced what wiring it up (option 1) would actually do against the
existing test suite, and it is not a safe change:

- Its first check accepts only `set`, `dict`, or `None`. `Tool.parameters` is typed
  `Iterable[Parameter] | Dict[str, Any] | None`, and a plain list is already accepted
  and tested (`parameters=[]` in `test_llm_helpers.py`, `test_model_invoker.py`,
  `test_node_builder.py`). Wiring the function in would reject every one of those.
- Its schema branch requires `"additionalProperties"` and `"properties"` keys on any
  non-empty dict schema. `Tool`'s own tests construct schemas without them
  (`test_keeps_the_given_name`, `test_schema_without_properties`), and both are
  accepted today.

So `validate_tool_params` does not describe the contract `Tool` actually has; it
predates the class's current signature. Went with option 2: delete the function and
its now-unused `ExceptionMessageKey` entries, and drop the stale comment. The one
check it duplicated that `Tool.__init__` does not already have (rejecting a schema
whose `required` list names a field not in `properties`, see
`test_required_without_properties_raises`) was already inlined at `tool.py:52-59`
before this change, so nothing needs to move.

Fixes #1514

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [ ] Docs updated if user-facing behavior changed (none needed, no public API changed)
- [ ] Breaking changes include migration notes (none, no behavior changed)

## Notes

New regression test `test_validate_tool_params_no_longer_defined` fails on `main`
(the function still exists) and passes on this branch; ran both ways in the same
Docker image (`python:3.10`, `pip install -e packages/railtracks[all]`).

Full suite: `pytest --ignore=packages/railtracks/tests/unit_tests/retrieval
packages/railtracks/tests/unit_tests/ packages/railtracks/tests/integration_tests/`
passes (2252 passed, 8 skipped), same image.

Not checked: the `retrieval` test suite (unaffected, no retrieval code touched) and
CodeQL/dependency-license scanning, which need the GitHub-hosted actions to run.


Amir-R25:
Thanks for picking this up so quickly, @AmirF194 and especially for actually running Option 1 against the test suite before committing to a direction. That reproduction work was the most valuable part of the review: it surfaced that validate_tool_params doesn't match Tool's current signature (the list-parameters path, the minimal-schema tests), which is exactly the gap Option 1 needs to close rather than a reason to walk away from it.

After talking it through, we'd like to go with Option 1 after all and wire the function in and adjust it (and the affected tests) so the schema checks actually run at construction time. I've updated the issue body with the plan and the two adjustments needed. Going to close this PR in favour of that direction, but the diff here is still a useful reference for what to touch. Really appreciate the careful writeup and the reproducible test run, made the decision easy to reason about.

Happy to assign the issue to you if you'd like to continue working on it but no pressure!

AmirF194:
Makes sense, thanks for weighing it and updating the issue with the plan. Happy to take a shot at Option 1 against those two adjustments.

Amir-R25:
Awesome, thanks, issue assigned

PR #1519, opened 2026-08-31
test: verify cache-aware LLM request costs

### LINKED ISSUE #1433: Precise request cost with cache hit info
### Description

Most modern providers charge a significnatly (1/100) reduced cost for cache hit tokens. Currently we exclude these information from our calculation, dispite it actually being returned. 

Including this feature will sinificantly improe our token count accuracy (right now a long token can be off by a factor of 50)

YaoSong808:
## Summary

- verify that provider-normalized cached input token details reach LiteLLM cost calculation
- cover buffered and streamed responses
- cover OpenAI-compatible prompt token details and Anthropic cache-read token fields
- clarify that the usage-based fallback preserves discounted cache pricing

The usage-based pricing path added on main now makes cache-aware pricing work without duplicating LiteLLM pricing tables. These tests lock in that behavior for both supported response paths.

Closes #1433

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs
- [x] Tests

## Checklist

- [x] Lint & format pass (ruff check and ruff format)
- [x] Docs updated if user-facing behavior changed (not applicable)
- [x] Breaking changes include migration notes (not applicable)

## Notes

Validation: 2255 core and integration tests passed; 8 skipped. Dependency ordering and the basic framework smoke test also passed.

### DIFF
--- a/packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
+++ b/packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
@@ -992,7 +992,9 @@ class LiteLLMWrapper(ModelBase, ABC):
         system_fingerprint = _return_none_on_error(lambda: raw.system_fingerprint)
         total_cost = _return_none_on_error(lambda: raw._hidden_params["response_cost"])
         if total_cost is None:
-            # A streamed response carries no cost. Price it from the usage chunk.
+            # A streamed response carries no cost. Price it from the complete usage
+            # object so provider-normalized details such as cached input tokens retain
+            # their discounted rate.
             total_cost = _return_none_on_error(
                 lambda: litellm.completion_cost(
                     completion_response=raw, model=requested_model or raw.model
--- a/packages/railtracks/tests/unit_tests/llm/models/test_litellm_wrapper.py
+++ b/packages/railtracks/tests/unit_tests/llm/models/test_litellm_wrapper.py
@@ -936,6 +936,59 @@ class TestStreamedUsage:
         assert info.total_cost is not None
         assert info.total_cost > 0
 
+    @pytest.mark.parametrize(
+        "response_type",
+        [ModelResponse, ModelResponseStream],
+        ids=["buffered", "streamed"],
+    )
+    @pytest.mark.parametrize(
+        "cache_usage",
+        [
+            {"prompt_tokens_details": {"cached_tokens": 900}},
+            {"cache_read_input_tokens": 900},
+        ],
+        ids=["openai-compatible", "anthropic"],
+    )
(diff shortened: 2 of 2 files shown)


packages/railtracks/tests/unit_tests/llm/models/test_litellm_wrapper.py
```diff
@@ -936,6 +936,59 @@ def test_cost_is_priced_from_usage_when_litellm_leaves_it_unset(self):
         assert info.total_cost is not None
         assert info.total_cost > 0
 
+    @pytest.mark.parametrize(
+        "response_type",
+        [ModelResponse, ModelResponseStream],
+        ids=["buffered", "streamed"],
```
Aryan-Railtown: the buffered/streamed split here isn't actually testing two paths... both ModelResponse and ModelResponseStream are built by hand with empty _hidden_params, so both fall through to the same completion_cost fallback. can we split this into two tests: one for the fallback path (streamed, what's here now) and one that populates _hidden_params["response_cost"] directly to actually cover the primary/buffered path?

Aryan-Railtown (changes requested):

packages/railtracks/tests/unit_tests/llm/models/test_litellm_wrapper.py
YaoSong808: Addressed in 36a62fe. I split the parameterized test into two genuinely distinct paths: the streamed cache tests keep an empty `_hidden_params` and verify usage-based fallback pricing for both supported cache usage shapes, while the buffered test sets `_hidden_params["response_cost"]` directly and asserts `litellm.completion_cost` is not called. The full wrapper test file passes (73 tests), and Ruff check/format pass.

Aryan-Railtown (changes requested):
just one minor thing and we should be good to merge

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
```diff
@@ -992,7 +992,9 @@ def extract_message_info(
         system_fingerprint = _return_none_on_error(lambda: raw.system_fingerprint)
         total_cost = _return_none_on_error(lambda: raw._hidden_params["response_cost"])
         if total_cost is None:
-            # A streamed response carries no cost. Price it from the usage chunk.
+            # A streamed response carries no cost. Price it from the complete usage
+            # object so provider-normalized details such as cached input tokens retain
+            # their discounted rate.
```
Aryan-Railtown, line 997: could you shorten this inline comment please

Aryan-Railtown (changes requested):

PR #1521, opened 2026-08-31
fix: wire validate_tool_params into Tool.__init__ (Option 1)

### LINKED ISSUE #1514: Dead code: `validate_tool_params` is defined but never called

AmirF194:
## Summary

`validate_tool_params` (`validation/node_creation/validation.py`) checks that a
`Tool`'s parameters are a well-formed JSON schema (`type: object`,
`additionalProperties: False`, `properties` present) or a set of `Parameter`
objects, but nothing calls it. The only reference is an aspirational comment in
`Tool.__init__`. A schema missing `additionalProperties` or `properties` builds a
`Tool` with no complaint and only fails later, as a provider error at LLM-call
time.

Wired `validate_tool_params` into `Tool.__init__`, with the two adjustments this
needs against `Tool`'s current signature:

- The type guard only accepted `set`, `dict`, or `None`, but `Tool.parameters` is
  typed `Iterable[Parameter] | Dict[str, Any] | None` and the two real production
  call sites (`Tool.from_function`, `node_builder.py`'s `_prepare_llm_tool`) both
  pass a `list`. Added `list` alongside `set`.
- `test_keeps_the_given_name` and `test_schema_without_properties` predate any
  enforcement and used schemas without `additionalProperties`; updated both (and
  `test_required_without_properties_raises`, which needed a `properties: {}` key
  to still reach `Tool`'s own required-fields check instead of failing the new
  guard first) to the contract the issue describes.

Also removed the stale comment at `tool.py:51` now that the checks it pointed at
actually run.

Fixes #1514

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [ ] Docs updated if user-facing behavior changed (none needed, no public API changed)
- [ ] Breaking changes include migration notes (none: this only rejects schemas that
      were already broken, and the two real callers that pass a list are now the
      supported list branch of the same guard)

## Notes

Verified in a clean `python:3.10` container (matching CI's `PYTHON_VERSION`),
`pip install -e packages/railtracks[all]`:

- Two new tests toggle the exact regression: `test_schema_missing_additional_properties_now_raises`
  and `test_list_with_non_parameter_element_raises` both fail on unmodified `main`
  (`DID NOT RAISE`) and pass on this branch, run both ways in the same container.
  `test_list_of_parameter_objects_is_accepted` pins the list branch used by
  `Tool.from_function`.
- Full suite green: `pytest --ignore=.../tests/unit_tests/retrieval tests/unit_tests/
  tests/integration_tests/` -> 2264 passed, 8 skipped (retrieval and the two
  live-server contract suites excluded, matching this repo's own PR gate).
- `ruff check` and `ruff format --check` clean; `scripts/check_dependencies_sorted.py`
  clean.
- Not run: `scripts/docs_validation.sh` (documentation-validation job) and the
  retrieval suite, neither of which this diff touches.


AmirF194:
This implements Option 1 as specced on #1518: validate_tool_params now accepts list/iterable parameters (not just set/dict), wired into Tool.__init__, and the minimal-schema tests updated to the now-enforced contract. Also removed the stale aspirational comment at tool.py:51 now that it's actually wired in.

CoronRing (changes requested):
Good catch on the root cause here. `validate_tool_params` really was dead code, so this is a genuine gap worth closing, and the new tests in `test_tool.py` are a real improvement over zero coverage.

That same fact is what makes the change riskier than it looks, though: wiring the helper in for the first time activates pre-existing bugs inside it and promotes its assertions into hard requirements on the public `Tool` API. A few `Tool(...)` calls that work on `main` now raise. I applied the patch on top of `origin/main` and exercised the constructor on both sides for each case below.

### Blocking

**1. The import escapes the llm package, and CI fails** (`tool.py:50`)

`from railtracks.validation...` is an upward import out of `railtracks.llm`. `test_llm_package_does_not_import_upward[tools/tool.py]` fails on this branch. The guard added in 56060add forbids exactly this shape, its assertion message calls out "errors raised on its behalf by a helper it calls." Please don't add a `KNOWN_ESCAPES` entry; that allowlist is meant to shrink, and `test_known_escapes_are_all_still_real` exists to keep it honest. The check should live inside `railtracks.llm` and raise `ToolCreationError`.

**2. Enum members passed where strings are expected, so the error can't be printed** (`validation.py:334`)

The set/list branch passes `ExceptionMessageKey` members directly into `message=`/`notes=` instead of going through `get_message`/`get_notes`:

```python
Tool(name="bad", detail="d", parameters=["city"])
# NodeCreationError is raised, but then:
str(e)  # TypeError: 'ExceptionMessageKey' object is not iterable
```

The message added at `exception_messages.yaml:78` therefore never renders. `test_list_with_non_parameter_element_raises` passes only because it has no `match=` argument, adding one would have caught this.

**3. Tuples and frozensets are now rejected** (`validation.py:300`)

The whitelist covers `set`/`list`/`dict`/`None`, but `tool.py:38` still annotates the parameter as `Iterable[Parameter]`. On `main`, a tuple or frozenset of `Parameter` produces a correct litellm payload; on this branch both raise `NodeCreationError`, and a type checker still accepts the call because the signature wasn't narrowed. Worth either widening the runtime check to `Iterable` or narrowing the annotation so the two agree.

**4. `additionalProperties: False` is required, then thrown away** (`validation.py:314`)

Two things here. It rejects raw dict schemas that worked on `main`, and the key has no effect either way: `_handle_set_of_parameters` (`llm/models/_litellm_wrapper.py:120`) regenerates the schema and never emits it, and `schema_parser.py:37` / `parameter_handlers.py:112` already default it to `False` when absent. Relatedly, the new test's docstring says such a schema "used to blow up later at LLM-call time", as far as I can tell that isn't the case, nothing downstream ever read the key.

**5. One constructor, two disjoint exception roots** (`tool.py:54`)

`Tool()` can now raise `NodeCreationError` for type/schema problems or `ToolCreationError` for required-without-properties. Those roots were kept separate deliberately, and the taxonomy's own guidance is to wrap tool definition in `except ToolCreationError`, which now silently misses half the failure modes. Folding this into (1) resolves it.

### Minor

- With `"properties" in parameters` now asserted, the `parameters.get("properties") or {}` fallback in `Tool.__init__` is unreachable.
- `PARAMETER_SET_CONTAINS_INVALID_TYPE_NOTES` still says "pass in an empty set" now that lists are accepted too.

The direction is right, it's mainly that the validation body needs a pass now that it's actually reachable, and the check needs to sit on the `llm` side of the package boundary. A `match=` on each of the new raising tests would pin down (2) and anything similar.

I concatenated my feedbacks in a singel chunk so its easier to copy/AI. Please let meknow if you prefer them in line!

AmirF194:
All five addressed at 7cfa09a9. Moved the check into tool.py itself so it never leaves the llm package, raising ToolCreationError instead of NodeCreationError, which folds in the disjoint-roots point too. Widened the type guard to any Iterable (excluding str/bytes), so tuple and frozenset go through now. Dropped the additionalProperties requirement, since litellm_wrapper.py and schema_parser.py never read it back and it was rejecting schemas that build fine on main.

Also fixed the enum-into-message bug: `str(e)` on the old NodeCreationError raised TypeError because the message/notes were ExceptionMessageKey members, never resolved through get_message/get_notes.

Cleaned up the unreachable `.get("properties") or {}` fallback, and added a match= to the new raising tests.

CoronRing (changes requested):
his is a much better shape than the first pass, great work, and moving the check into `llm/tools/tool.py` as `_validate_tool_params` was the right call rather than allowlisting the import. All five blocking items and both minor notes are addressed:

- The layering guard passes now: `pytest packages/railtracks/tests/unit_tests/llm/test_exceptions.py` is 56/56, and `KNOWN_ESCAPES` was left alone.
- The enum-into-`message=` bug is gone with the old function, and the new errors actually render: `str(e)` on each raising path returns real text.
- `tuple`, `frozenset`, and `dict_values` of `Parameter` all construct again, matching `main` and the `Iterable[Parameter]` annotation.
- The `additionalProperties` requirement is dropped, and `test_schema_missing_additional_properties_is_accepted` pins it.
- Everything raises `ToolCreationError` from one root, so `except ToolCreationError` around tool definition is sufficient again.
- The `parameters.get("properties") or {}` fallback is now a direct `parameters["properties"]`, and the message text was updated.

One new thing to fix before merge, then three small tidy-ups.

### Blocking: generators are silently emptied

`tool.py:58` consumes the iterable inside `all(isinstance(x, param_type) for x in parameters)`, so anything single-pass is exhausted before the constructor stores it:

```python
Tool("t", "d", (mk(n) for n in ["a", "b"]))
# main:   tool.parameters -> ['a', 'b']
# branch: tool.parameters -> []
```

A generator is legal under the declared `Iterable[Parameter]`, nothing raises, and the tool reaches the provider with no parameters at all. That is the same silent-wrong-schema failure this PR exists to prevent, so it would be a shame to ship it in the validator itself. Materializing once fixes it and makes the stored container predictable:

```python
if isinstance(parameters, ABCIterable) and not isinstance(parameters, (str, bytes)):
    parameters = list(parameters)
    ...
```

That needs the materialized list handed back to `__init__` (return it, or do the `list()` in `__init__` before validating). A test alongside the tuple/frozenset ones would cover it.

### Minor

The four `TOOL_PARAMETERS_*` / `PARAMETER_SET_CONTAINS_INVALID_TYPE_*` keys now have no consumer anywhere in `src` beyond the enum declaration itself, since the new messages are inline literals. Worth deleting the enum members and their `exception_messages.yaml` entries in this PR: as it stands the yaml still says "a set or list of Parameter objects" while the live message says "an iterable", so the dead copy now contradicts the real one. (The `llm` package can't read those keys without re-introducing the upward import, so inline literals are the right choice, just finish the removal.)

`validation.py:299` leaves a comment explaining where `validate_tool_params` went. Per AGENTS.md, comments shouldn't reference previous versions of the code. The whole block including the START/END banners can go; `git log` covers the history. Same for the "before `validate_tool_params` was wired in" phrasing in the new test docstring, which reads fine as just a statement of current behavior.

The `parameters` property is annotated `List[Parameter] | None` but returns whatever container it was given (a `set` on the common path). Pre-existing, though this PR widens the accepted set enough that it's more visible now. `Iterable[Parameter] | None`, or the `list()` from the fix above, would settle it.

AmirF194:
Fixed at f970416b. Materialized any non-dict iterable inside `_validate_tool_params` before running the type check, and return it so `__init__` stores the same list it just validated (a generator, tuple, or frozenset are all now `list`, matching the `parameters` property's own annotation). Added `test_generator_of_parameter_objects_is_not_exhausted`, confirmed failing on the unpatched code (`assert [] == ['city', 'country']`) and passing on the branch.

Also took the three tidy-ups: deleted `PARAMETER_SET_CONTAINS_INVALID_TYPE_*` and `TOOL_PARAMETERS_*` from exception_messages (all five entries, not just the four named, since this PR's own diff is what made every one of them unreferenced), removed the stale comment banner they left behind in validation.py, and reworded the test docstring.

`ruff check`/`ruff format --check` clean, full `test_tool.py` + `test_exceptions.py` suite 67 passed.

CoronRing (approved):
Wonderful changes, looks really clean on my side! You should be able to click that merge button (for the pleasure and honour!) ! If not let me know, I can squash and merge.

AmirF194:
No merge button on my end, I don't have push access here. Go ahead and squash whenever suits you.

The generator catch on tool.py:58 was a good one, that's exactly the kind of silent-wrong-schema case this PR set out to close.

PR #1527, opened 2026-09-01
fix: serialize structured-output BaseModel content in _to_litellm_message

### LINKED ISSUE #1493: Structured-output AssistantMessage content not serialized for a second/later completion call
When a structured-output call parses its JSON response into a pydantic model, the resulting `AssistantMessage(content=<BaseModel instance>)` is passed unchanged through `_to_litellm_message()`'s fallback branch, so a later `litellm.completion(...)` call receives a message dict whose `content` is a raw pydantic `BaseModel` instance instead of a string. If that message (e.g. from `resp.message_history`) is fed back into a second/later completion call, litellm forwards this non-string `content` straight to the OpenAI SDK, which rejects it with a 400 `BadRequestError`.

## Repro

No network access or API key is required — only the pure Python conversion method `_to_litellm_message` is exercised, not any live LLM call.

```python
"""
Repro: structured-output AssistantMessage.content (a pydantic BaseModel) is
passed through _to_litellm_message()'s fallback branch verbatim, producing a
litellm/OpenAI-SDK message dict whose "content" is a BaseModel instance
instead of a string. This is exactly what a manually-threaded multi-turn chat
loop (turn 1 structured response fed back in as message_history for turn 2)
sends to litellm.completion(messages=[...]), which the OpenAI SDK rejects
with:

    openai.BadRequestError: Error code: 400 - {'error': {'message':
    "Invalid type for 'messages[2].content': expected one of a string or
    array of objects, but got an object instead.", ...}}

No network access or API key is required: only the pure Python conversion
method _to_litellm_message is exercised, not any live LLM call.

Run with: uv run python repro_structured_content_bug.py
(from a railtracks checkout so `railtracks` and `pydantic` are importable)
"""

from railtracks.llm.message import AssistantMessage
from railtracks.llm.models.api_providers.openai import OpenAILLM


class Reply(BaseModel):
    text: str
    wants_to_end: bool


def main() -> None:
    # Simulates what _structured_handle_base() does after turn 1 parses the
    # model's structured JSON response:
    #   parsed = schema(**json.loads(content_str))
    #   return Response(message=AssistantMessage(content=parsed), ...)
    turn1_reply = Reply(text="hi", wants_to_end=False)
    assistant_msg = AssistantMessage(content=turn1_reply)

    print("AssistantMessage.content type:", type(assistant_msg.content))
    print("AssistantMessage.content repr:", repr(assistant_msg.content))

    # Construct a minimal LiteLLMWrapper subclass instance. This performs no
    # network call and requires no real API key -- __init__ just stores
    # config fields.
    wrapper = OpenAILLM(model_name="gpt-4o-mini", api_key="sk-not-a-real-key")

    # This is the exact per-message conversion litellm-backed chat/structured
    # calls run before calling litellm.completion(), e.g.:
    #   litellm_messages = [self._to_litellm_message(m) for m in messages]
    litellm_msg = wrapper._to_litellm_message(assistant_msg)

    print("\nResulting litellm message dict:")
    print(litellm_msg)

    content = litellm_msg["content"]
    print("\ncontent type:", type(content))
    print("content repr:", repr(content))

    if isinstance(content, str):
        print(
            "\nRESULT: content is a string -- bug NOT reproduced "
            "(fallback branch is producing valid content)."
        )
        raise SystemExit(1)
    else:
        print(
            "\nRESULT: content is NOT a string (it's a raw "
            f"{type(content).__name__} instance) -- BUG REPRODUCED.\n"
            "This is exactly the shape litellm.completion(messages=[...]) "
            "would receive, and it is what the OpenAI SDK rejects with:\n"
            "  openai.BadRequestError: Error code: 400 - {'error': "
            "{'message': \"Invalid type for 'messages[N].content': expected "
            "one of a string or array of objects, but got an object "
            "instead.\", ...}}"
        )
        assert not isinstance(content, str), "content must not be a string here"


if __name__ == "__main__":
    main()
```

Run with `uv run python repro_structured_content_bug.py` from a railtracks checkout so `railtracks` and `pydantic` are importable. Expected output ends with:

```
RESULT: content is NOT a string (it's a raw Reply instance) -- BUG REPRODUCED.
```

## Root cause

`_structured_handle_base` (`packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py:740-748`) parses a structured-output completion into a pydantic model instance and wraps it directly as `AssistantMessage(content=parsed)` (line 748), where `parsed = schema(**json.loads(content_str))` (line 747). This is by-design storage for a structured turn's parsed content, per `Content = Union[str, ToolCalls, List[ToolCall], ToolResponse, BaseModel]` in `llm/content.py:140`.

The break is one layer down: `_to_litellm_message` (`_litellm_wrapper.py:863`) branches on message shape (UserMessage+attachment, ToolMessage, list-of-ToolCall) starting at line 870, but has no case for a bare `BaseModel` content, so it falls into the generic `else: base["content"] = msg.content` at line 962 and hands the raw pydantic model instance straight through unconverted.

When a caller manually re-feeds turn 1's `resp.message_history` (which now contains that `AssistantMessage`) into a fresh structured call for turn 2, `_invoke` (line 368: `litellm_messages = [self._to_litellm_message(m) for m in messages]`) maps every message through `_to_litellm_message` and passes the resulting dicts to `litellm.completion(messages=...)`. litellm then forwards this non-string, non-array `content` field to the OpenAI SDK, which rejects it with a 400 `BadRequestError`:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid type for 'messages[N].content': expected one of a string or array of objects, but got an object instead.", ...}}
```

## Context

This was originally found while running an example script exploring a continuously-running human-in-the-loop chat pattern (a structured-output agent used across multiple turns, with `message_history` manually threaded from one call into the next -- a real, documented usage pattern, not a misuse). It is NOT a human-in-the-loop/verifier bug -- it's in core LLM message serialization (`packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py`). It was previously flagged informally (not as a GitHub issue) on a Notion page ("Known bug: structured-output content breaks multi-turn history", linked from the "Human in the loop" Notion doc) -- this GitHub issue is the first formal tracking of it.

Aryan-Railtown:
## Summary
- Fixes #1493: `_to_litellm_message` had no case for `AssistantMessage.content` being a parsed pydantic `BaseModel` (the shape a structured-output turn stores), so it passed the raw `BaseModel` straight through to litellm/OpenAI, which rejects non-string `content` with a 400 when that message is later re-fed into a completion call (e.g. manually-threaded `message_history` in a multi-turn loop).
- Serializes it via `model_dump_json()` instead.
- Extracted the attachment-handling block into `_to_litellm_attachment_content` to keep the dispatcher under the cyclomatic-complexity lint threshold after adding the new branch.

## Test plan
- [x] Added `test_to_litellm_message_structured_output_content`, confirmed it fails before the fix and passes after
- [x] Full `_to_litellm_message` test suite passes (71 tests)
- [x] `ruff check` / `ruff format --check` pass


Amir-R25 (approved):
Nice find!

PR #1528, opened 2026-09-01
Opt in stacktrace

Amir-R25:
## Summary

`LLMError.__str__` used to render the full input `MessageHistory` into the exception text, so any `logger.exception(...)`, `str(err)`, `repr(err)` contained the message history.

Default now emits a redacted summary (`"N message(s) redacted"`). The `message_history` attribute is still there for structured callers. When a specific log line genuinely needs the full render, call `err.format_verbose()` at that site:

```python
except LLMError as err:
    logger.error("LLM call failed: %s", err.format_verbose())
```

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

## Notes

No global toggle. The choice to render the history is made at the call site that logs it, which keeps `str(err)` predictable regardless of process state.

Docs: new `### Inspecting LLMError message history` subsection in `error_handling.md`

### DIFF
--- a/packages/railtracks/src/railtracks/llm/models/_model_exception_base.py
+++ b/packages/railtracks/src/railtracks/llm/models/_model_exception_base.py
@@ -20,24 +20,40 @@ class ModelError(RTLLMError):
 
     def __str__(self):
         base = super().__str__()
-        details = []
-        if self.message_history:
-            mh_str = str(self.message_history)
-            indented_mh = "\n".join(
-                "    " + line for line in mh_str.splitlines()
-            )  # 2 indents (2-spaces) per indent
-            details.append(
-                self._color("Message History:\n", self.BOLD_GREEN)
-                + self._color(indented_mh, self.GREEN)
-            )
-        if details:
-            notes_str = (
-                "\n"
-                + self._color("Details:\n", self.BOLD_GREEN)
-                + "\n".join(f"  {d}" for d in details)
-            )
-            return f"\n{self._color(base, self.RED)}{notes_str}"
-        return self._color(base, self.RED)
+        if not self.message_history:
+            return self._color(base, self.RED)
+
+        try:
+            count = len(self.message_history)
+        except TypeError:
+            count = None
+
+        summary = (
+            f"{count} message(s) redacted; "
+            "call err.format_verbose() to render or read err.message_history"
+            if count is not None
+            else "message history redacted; "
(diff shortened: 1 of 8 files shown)


packages/railtracks/src/railtracks/llm/models/_model_exception_base.py
```diff
@@ -20,24 +20,40 @@ def __init__(
...
+            else "message history redacted; "
+            "call err.format_verbose() to render or read err.message_history"
+        )
+        detail = self._color("Message History: ", self.BOLD_GREEN) + self._color(
+            summary, self.GREEN
+        )
+        notes_str = "\n" + self._color("Details:\n", self.BOLD_GREEN) + f"  {detail}"
+        return f"\n{self._color(base, self.RED)}{notes_str}"
+
+    def format_verbose(self) -> str:
```
Aryan-Railtown, line 44: Minor thing that was bugging me when this was first designed as well:
There is code duplication between the 2 modules. I think it goes out of scope of this ticket so just mentioning it here. I believe the Exception refact is already in progress rn, so we can look into that separately.

Aryan-Railtown (commented):
Minor comment, other than that looks good 🚀

Aryan-Railtown (approved):

CoronRing (approved):
1. Structured output bypasses the whole new classification:  _litellm_wrapper.py:852

_structured's blanket except Exception re-wraps the freshly classified provider errors back into a generic ModelError. Reproduced directly against the branch: with litellm.completion raising Timeout, chat gives ProviderTimeoutError but structured gives ModelError; with FixedRetry, chat gives RetryError but structured gives ModelError. So an agent using structured output can never catch LLMTimeoutError/LLMRateLimitError/LLMAuthenticationError.

2. Streaming's "no final Response" path now loses message_history: model_invoker.py:81

Since LLMError is now a NodeInvocationError, the LLMError raised by _drain_to_queue matches the new except NodeInvocationError: raise at llm_helpers.py:124 and propagates with message_history=None. Before this PR it fell to the catch-all branch, which attached the live history. Users following the new docs' if e.message_history: pattern get nothing on that path. Best fixed at the raise site by passing the history in.

Will kind break the feature branch for exception rework.
Both is not super tied and can be done as another ticket, depending on how speedy this PR is, feel free to just merge in and solve after.

PR #1541, opened 2026-09-06
fix: prevent type collapse when combining prebuilt middleware in a list

### LINKED ISSUE #1538: Type hints collapse when you use a list of pre-built middlewares
### Affected component

Agents / Nodes

### Description

# Root cause

### Reproduction

```python
import railtracks as rt

def a(x: str) -> str:
    return x
```

### Error trace / logs

```shell


```

### System info

```shell

```

Christian-Sidak:
## Summary

- Retry, Timeout, MaxCalls, and Lock subclassed the bare Middleware generic, which mypy resolved as Middleware[Never, Never] under invariant ParamSpec generics. Combining any two in a single middleware=[] list caused _P to collapse to Never, producing a (*Never, **Never) mismatch against the decorated function's real signature.
- Two changes fix this: (1) parameterize all four classes as Middleware[Any, Any], and (2) use Iterable[Middleware[Any, Any]] in the func=None overload of function_node so _P is inferred from the decorated function rather than constrained by the middleware list.
- Adds regression tests combining multiple prebuilt middleware in one list, and a CI-mypy-checked docs/scripts example.

Fixes #1538

## Test plan

- [x] mypy on docs/scripts/prebuilt_middleware.py passes (including the new combining example)
- [x] Exact reproduction from issue passes mypy
- [x] All existing middleware unit tests pass
- [x] New regression tests for combining two and three prebuilt middleware pass

### DIFF
--- a/packages/railtracks/src/railtracks/built_nodes/function/node.py
+++ b/packages/railtracks/src/railtracks/built_nodes/function/node.py
@@ -6,6 +6,7 @@ import inspect
 import warnings
 from types import BuiltinFunctionType
 from typing import (
+    Any,
     Callable,
     Coroutine,
     Iterable,
@@ -78,7 +79,7 @@ def function_node(
     *,
     name: str | None = None,
     manifest: ToolManifest | None = None,
-    middleware: Iterable[Middleware[_P, _TOutput]] | None = None,
+    middleware: Iterable[Middleware[Any, Any]] | None = None,
 ) -> Callable[
     [Callable[_P, Coroutine[None, None, _TOutput]] | Callable[_P, _TOutput]],
     RTFunction[_P, _TOutput],
--- a/packages/railtracks/src/railtracks/built_nodes/function/node.py
+++ b/packages/railtracks/src/railtracks/built_nodes/function/node.py
@@ -33,6 +33,14 @@ from .node_builder import FunctionNodeBuilder
 _TOutput = TypeVar("_TOutput")
 _P = ParamSpec("_P")
 
+# Separate TypeVars for the decorator-factory overload so the middleware parameter
+# constrains the decorated function rather than accepting everything via Any.
+# When slot-agnostic middleware (Middleware[Any, Any]) is provided the inferred
+# _MW_P / _MW_Out will be Any, which is the correct result: slot-agnostic middleware
+# imposes no constraint and lets any function through.
+_MW_P = ParamSpec("_MW_P")
+_MW_Out = TypeVar("_MW_Out")
+
 
 # note there is an intentional overlap in overloads
 # by running the first overload check it will pick up all `async` functions
@@ -79,10 +87,10 @@ def function_node(
     *,
     name: str | None = None,
     manifest: ToolManifest | None = None,
-    middleware: Iterable[Middleware[Any, Any]] | None = None,
+    middleware: Iterable[Middleware[_MW_P, _MW_Out]] | None = None,
 ) -> Callable[
-    [Callable[_P, Coroutine[None, None, _TOutput]] | Callable[_P, _TOutput]],
-    RTFunction[_P, _TOutput],
+    [Callable[_MW_P, Coroutine[None, None, _MW_Out]] | Callable[_MW_P, _MW_Out]],
+    RTFunction[_MW_P, _MW_Out],
 ]:
     pass
 
--- a/packages/railtracks/src/railtracks/built_nodes/function/node.py
+++ b/packages/railtracks/src/railtracks/built_nodes/function/node.py
@@ -33,14 +33,6 @@ from .node_builder import FunctionNodeBuilder
 _TOutput = TypeVar("_TOutput")
 _P = ParamSpec("_P")
 
-# Separate TypeVars for the decorator-factory overload so the middleware parameter
-# constrains the decorated function rather than accepting everything via Any.
-# When slot-agnostic middleware (Middleware[Any, Any]) is provided the inferred
-# _MW_P / _MW_Out will be Any, which is the correct result: slot-agnostic middleware
-# imposes no constraint and lets any function through.
-_MW_P = ParamSpec("_MW_P")
-_MW_Out = TypeVar("_MW_Out")
-
 
 # note there is an intentional overlap in overloads
 # by running the first overload check it will pick up all `async` functions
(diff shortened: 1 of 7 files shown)


packages/railtracks/src/railtracks/built_nodes/function/node.py
```diff
@@ -78,7 +79,7 @@ def function_node(
     *,
     name: str | None = None,
     manifest: ToolManifest | None = None,
-    middleware: Iterable[Middleware[_P, _TOutput]] | None = None,
+    middleware: Iterable[Middleware[Any, Any]] | None = None,
```
Aryan-Railtown: This is the one that actually breaks things, not just an inconsistency. Widening this to Any, Any means @rt.function_node(middleware=[...]) (the decorator form, used in your own tests) no longer checks that a middleware's signature matches the function it's attached to... verified with mypy, a middleware typed for a totally different function passes silently now.

Also note: just swapping this back to Middleware[_P, _TOutput] to match the other overloads (lines 47/58/186/265) isn't the fix either, that's what caused #1538 in the first place, since a heterogeneous middleware=[...] list can't unify under one shared _P/_TOutput without collapsing to Never. The real fix needs the generics to handle a mixed-middleware list without losing the per-middleware signature check, not just picking one of the two existing options

packages/railtracks/tests/unit_tests/middlewares/test_node_middleware.py
```diff
@@ -145,3 +146,38 @@ async def top_level():
             return await rt.call(five)
 
     assert asyncio.run(top_level()) == 50
+
+
+def test_multiple_prebuilt_middleware_in_one_list():
```
Aryan-Railtown, line 151: Also the new tests are runtime only, so they can't actually catch this since generics get erased at runtime.
But I am okay with keeping tests!!

Aryan-Railtown (changes requested):
Really solid direction on this one... parameterizing Retry/Timeout/MaxCalls/Lock as Middleware[Any, Any] at the class level is the right call since they're genuinely slot-agnostic, and that's the core of #1538 handled correctly. One thing on the `function_node` side worth a closer look before merging, left as an inline comment.

Christian-Sidak:
Addressed in 5f3c807. The decorator-factory overload now uses dedicated ParamSpec/TypeVar (_MW_P, _MW_Out) for the middleware parameter and links them to the return type, so a middleware typed for a different function signature is caught by mypy. For slot-agnostic prebuilt middleware (Retry, Timeout, MaxCalls, all Middleware[Any, Any]) the inferred _MW_P/_MW_Out resolve to Any -- no constraint is added, which is correct. The trade-off vs the current Any,Any approach: when slot-agnostic middleware is provided, the result type is RTFunction[Any, Any] instead of RTFunction[_P, _TOutput], so callers that need precision would need an explicit annotation. Happy to take a different approach if there is a preferred solution -- e.g. a higher-kinded bound, a separate covariant TypeVar, or keeping Any,Any with a runtime check.

Aryan-Railtown:
> Addressed in [5f3c807](https://github.com/RailtownAI/railtracks/commit/5f3c807e64672517cca6a9809b0b36c525d107e7). The decorator-factory overload now uses dedicated ParamSpec/TypeVar (_MW_P, _MW_Out) for the middleware parameter and links them to the return type, so a middleware typed for a different function signature is caught by mypy. For slot-agnostic prebuilt middleware (Retry, Timeout, MaxCalls, all Middleware[Any, Any]) the inferred _MW_P/_MW_Out resolve to Any -- no constraint is added, which is correct. The trade-off vs the current Any,Any approach: when slot-agnostic middleware is provided, the result type is RTFunction[Any, Any] instead of RTFunction[_P, _TOutput], so callers that need precision would need an explicit annotation. Happy to take a different approach if there is a preferred solution -- e.g. a higher-kinded bound, a separate covariant TypeVar, or keeping Any,Any with a runtime check.

This reintroduces #1538 rather than fixing it. Verified with mypy: the new _MW_P/_MW_Out split on the func=None overload brings back the Never-collapse error, e.g. @rt.function_node(middleware=[Retry(3), Timeout(seconds=30)]) fails again on a plain typed function. Also fails CI as pushed (unused Any import trips ruff-lint).

Simplest path forward: drop the _MW_P/_MW_Out change entirely. The four Middleware[Any, Any] class edits alone already make the direct-call form work cleanly, no overload changes needed. Then add the negative type-checking coverage we talked about, since neither push has that yet.

Christian-Sidak:
Fixed in c77ef7b.

Dropped the _MW_P/_MW_Out TypeVars entirely and reverted the func=None overload to use Middleware[Any, Any]. The slot-agnostic overload leaves _P unconstrained so it is inferred from the decorated function rather than from the middleware list -- which is the same behavior as a5c9ad5 and does not reintroduce the Never-collapse.

Also added three tests as the negative type-coverage:
- test_decorator_factory_form_preserves_signature: applies the decorator form and then checks that the resulting node exposes the decorated function's real parameter names in tool_info(). Collapsed _P would lose them at runtime, so this test would fail if the Never error returns.
- test_prebuilt_middleware_does_not_constrain_function_signature: applies the same [Retry, Timeout] list to two functions with structurally different signatures and verifies both calls produce correct results. A constrained list would make one call fail.

The ruff-lint unused import is gone since Any is now back in use in the overload.

Lpk124:
Thanks for updating this. I checked `c77ef7b` with mypy 1.19.1 and 2.3.1 against the remaining factory cases. An async function with `[Retry, Timeout]` still collapses to `Never`, factory-decorated functions are exposed as non-callable `RTFunction` values, and mismatched pre/post verifier signatures leave the expected ignores unused. The runtime tests pass, but they do not exercise those static failures.

I opened #1558 with those cases covered. It keeps your four class changes as the first commit with your authorship intact, then carries the middleware signature constraint separately for factory inference. Both mypy versions pass the positive fixture and report the same six errors when the expected ignores are removed.

Christian-Sidak:
Thanks for the detailed analysis, @Lpk124. Your point about the async Never-collapse on factory-decorated functions with `[Retry, Timeout]` is correct, and I see why reverting the overload change was not enough on its own.

Since #1558 builds on the four class changes from this PR (with authorship preserved) and adds the factory inference fix on top, I'm happy to close this one once #1558 is reviewed and merged, rather than continuing to iterate here in parallel. Happy to answer any questions about the original changes if it helps your review.

Aryan-Railtown:
@Christian-Sidak @Lpk124 Perfect, that PR #1558 fixes the issue correctly, i just had one minor nitpick for that PR then we are good to merge. Closing this for now! Thank you for the amazing work @Christian-Sidak @Lpk124

Christian-Sidak:
Thanks for the update, Aryan. Understood, happy to yield to #1558 once it lands. Appreciate you reviewing both PRs.

PR #1542, opened 2026-09-06
docs: explain message history handoff

### LINKED ISSUE #1133: [Docs] We need docs on message history between session and calls.
### Issue Categorization

- [x] Missing documentation
- [ ] Incorrect or outdated information
- [ ] Overly verbose or complex explanation
- [ ] Broken links or formatting
- [ ] Missing code examples
- [ ] Typos or grammar issues

### Documentation Problem

Currently, we never explain how to pass message history around. We show this in a couple of places, such as structured_extraction, but it was never explained. Tutorials and examples were not check since they are not part of docs site. This should be further highlighted for clarity:

this may include passing message_history object in same function or saving and passing them in global context across methods.

Image

### Additional Context

_No response_

be-student:
## Summary

Closes #1133.

This PR adds a canonical Message History guide that explains how to continue a conversation across `rt.call(...)` invocations, share history between nodes through run-scoped context, and persist history explicitly across separate runs. It adds type-checked examples, registers the page in the Agent Invocation navigation, and links to it from Direct Invocation, Context, and Structured Extraction.

The documentation now states the current system-message behavior precisely: caller-supplied system messages remain in `response.message_history`, while the target agent's configured system message is attached fresh on each call and does not accumulate in returned history.

Review feedback addressed:

- Include the imports and agent definition in the rendered example setup.
- Use the current `gpt-5.4-mini` example model.
- Move the material to a canonical Message History page with navigation and focused cross-links.
- Document system-message round-trip behavior.
- Warn that the context get → copy → append → put pattern assumes sequential turns; concurrent turns need separate keys or synchronized access.
- Keep new documentation paragraphs unwrapped and record that convention in the repository code-style skill.

Verified head: `50bc297836daaabe109756a7323ab78129c554fd`, rebased onto upstream `main` at `9db14df07713f2e4c64d3abc31fda515670a7128`.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

- [x] Lint & format pass (`ruff check . --no-fix` and `ruff format --check .`)
- [x] Breaking changes include migration notes (not applicable; no breaking change)

## Validation

- `ruff check . --no-fix`: passed.
- `ruff format --check .`: passed; 623 files already formatted.
- `mkdocs build --strict`: passed. Local Python 3.14 emitted existing pdoc annotation warnings while the build exited successfully.
- Focused mypy check for `docs/scripts/documentation/message_history.py`: passed.
- Core unit and integration run excluding retrieval, matching the core CI split: 2,468 passed and 8 skipped. Two MCP setup errors caused by the local `uv` environment lacking `pip` passed when rerun after installing local `pip`. One Python 3.14 behavior difference in `test_builtin_function_raises_runtime_error` passed when rerun under the repository's CI Python 3.10.

Local limitations:

- The repository-wide documentation mypy script reports three unrelated existing errors in `docs/scripts/tools_mcp_guides.py`, `docs/scripts/tools.py`, and `docs/scripts/evaluations/conductr.py`; the new Message History example passes its focused check.
- Full retrieval-test collection is blocked on this macOS host because system policy rejects the third-party `pypdfium2_raw/libpdfium.dylib` signature before tests run.

## Notes

AI assistance disclosure: OpenAI Codex assisted with the documentation changes, upstream reconciliation, and validation. The exact head and all known validation limitations are reported above.


docs/scripts/documentation/message_history.py
```diff
@@ -0,0 +1,52 @@
...
+
+ChatAgent = rt.agent_node(
+    name="ChatAgent",
+    system_message="Answer clearly and remember details from the conversation.",
+    llm=rt.llm.OpenAILLM("gpt-5.1"),
+)
+
+
+# --8<-- [start: direct_handoff]
+async def direct_conversation() -> str:
```
CoronRing, line 13: The snippet regions exclude the imports and the agent definition. import railtracks as rt and ChatAgent sit above # --8<-- [start: direct_handoff] in message_history.py, so both rendered blocks use rt and ChatAgent with no definition on the page. Every other doc script starts its region at the import. Either move the start marker up, or add a setup region shown once at the top of the section.

docs/documentation/invocation/call.md
```diff
@@ -18,4 +18,24 @@ resp = await rt.call(AgentName, "user message to the agent")
 
 The `call` API is also useful when you want to use agents as tools by having them wrapped within another a function (see [Agents as Tools](../agent_design/tools/agents_as_tools.md)). 
 
-For configuration management such as context, observability through invocations, and other settings we recommend using [Flows](../invocation/flows.md).
\ No newline at end of file
+For configuration management such as context, observability through invocations, and other settings we recommend using [Flows](../invocation/flows.md).
+
+## Continue a conversation across calls
+
+Each call is stateless: Railtracks does not automatically send an earlier conversation to the next call. An agent response includes the complete conversation in `response.message_history`. To continue that conversation, copy the returned history, append the next user message, and pass it to the agent again:
```
CoronRing: "An agent response includes the complete conversation" is not quite true: Responses are built with message_history.removed_system_messages() (llm_helpers.py:382), so the returned history has no system message. this is toensure its round-trip is safe: append_system_message prepends the agent's own system_message fresh on every call, so you don't accumulate a duplicate per turn. This logic may be updated later to return system message and detect duplication (Maybe your next ticket? ; D  ), but for now. Say it explicitly is easier so we do not block.

docs/scripts/documentation/message_history.py
```diff
@@ -0,0 +1,52 @@
+import railtracks as rt
+
+ChatAgent = rt.agent_node(
+    name="ChatAgent",
+    system_message="Answer clearly and remember details from the conversation.",
+    llm=rt.llm.OpenAILLM("gpt-5.1"),
```
CoronRing: Not error, but maybe we update to gpt-5.4-mini so it's consistent, will pass if one one, but since there are other changes request, might as well.

docs/documentation/invocation/call.md
```diff
@@ -18,4 +18,24 @@ resp = await rt.call(AgentName, "user message to the agent")
 
 The `call` API is also useful when you want to use agents as tools by having them wrapped within another a function (see [Agents as Tools](../agent_design/tools/agents_as_tools.md)). 
 
-For configuration management such as context, observability through invocations, and other settings we recommend using [Flows](../invocation/flows.md).
\ No newline at end of file
+For configuration management such as context, observability through invocations, and other settings we recommend using [Flows](../invocation/flows.md).
```
CoronRing, line 21: The "Share history between nodes in one run" section isn't about rt.call, more on context recipe. Filing it under a page the nav calls Direct Invocation means context.md still never mentions message history, so someone reading the context page won't find it. And call.md is currently 20 lines whose whole job is "here's the async syntax, use Flows if you need config" this addition triples it and changes what the page is about.

More importantly, #1133 says the problem is that message history is shown in a few places but never explained. That's a request for a canonical destination other pages can point at. structured_extraction.md:28 passes first_step.message_history with zero explanation; that line wants a link, and it can't link to a subsection of "Direct Invocation" without looking odd.

A smarter move will be:

- docs/documentation/invocation/message_history.md, nav entry Message History under Agent Invocation after Flow Invocation (must be added to nav: or --strict warns).
- The script is already named message_history.py, so no churn there.
- One-line pointers from call.md, context.md (under Real-World Examples), and structured_extraction.md:28 (any maybe other places you see fit? but not everywhere!)

CoronRing (changes requested):
Lovely PR, and apprciate the contribution, wish I can give you a cert for helping out!

Most request are structual/cleaness. Content wise its pretty good!

be-student:
Addressed the structural review in 88e73fb8. Message history now has a canonical nav page, complete setup snippets, explicit system-message behavior, the current model ID, and focused pointers from call, context, and structured extraction. Ruff and the strict docs build pass.

CoronRing (changes requested):
Two additional small item

1. Concurrency caveat on the context pattern. The get → copy → append → put sequence in conversation_turn is only correct because the two turns run sequentially. Two of those nodes in flight at once both read the same saved_history and the later put silently drops the other turn. This is super tiny and mot worth a rewrite, one sentence under that section, e.g. that the pattern assumes one turn at a time and concurrent turns need their own keys will be great!

2. The new markdown is hard-wrapped at ~95 chars. message_history.md and both cross-link paragraphs wrap mid-sentence. The rest of the docs don't — call.md and context.md use one long line per paragraph. Worth unwrapping for consistency; it also makes future diffs on those paragraphs reflow noisily. (claude often do this, for your benifit, I guess adding a rule in memory file telling it not to do that for doc pages)

GREEEEAAAAT work!

be-student:
Updated and rebased for re-review at `50bc297836daaabe109756a7323ab78129c554fd`.

All review requests are addressed: the examples include their setup, the guide is now the canonical nav destination with focused cross-links, system-message behavior matches current `main`, the context pattern has an explicit concurrency caveat, the new prose is unwrapped, and the style rule records that convention.

Ruff lint/format, the strict MkDocs build, and the focused example mypy check pass. The core unit/integration run completed with 2,468 passed and 8 skipped; its two local MCP setup errors pass after adding `pip`, and the remaining Python 3.14-specific failure passes under CI Python 3.10. The PR body records the unrelated repository-wide docs mypy errors and the macOS `libpdfium` retrieval-collection limitation.

Please re-review when convenient.

CoronRing (approved):
great work! Really high quality work! thanks for helping the project and community!

PR #1543, opened 2026-09-06
feat(retrieval): add structured content extractors

### LINKED ISSUE #1100: [Feature] Pluggable content extractor for nested dataset columns
### Problem

`HuggingFaceDatasetLoader` (and `JSONLoader`) currently render a column's value into `Document.content` via `str(row[col])` (see [`huggingface_loader.py:115`](packages/railtracks/src/railtracks/retrieval/loaders/huggingface_loader.py#L115)). 

For columns holding dicts or lists (e.g. `narrativeqa.question` = `{"text": ..., "tokens": [...]}`, `ms_marco.passages`), this produces a **Python repr**, single quotes, `None`/`True`, escaped apostrophes, which is neither valid JSON nor embedding-friendly prose. The string then flows through the chunker and embedder, so retrieval matches against Python syntax rather than text.

The discussion in #1098 surfaced this; the loader was merged with the existing default behaviour since it works for scalar text columns (most of the HF Hub).

### Proposal

Introduce a pluggable **content extractor**: a callable that turns a column value into a string and accept it as a constructor argument on loaders that ingest structured rows (`HuggingFaceDatasetLoader`, `JSONLoader`, future loaders with the same shape).

```python
ContentExtractor = Callable[[Any], str]
```

Ship a small set of built-ins:

| Extractor | Behavior |
|---|---|
| `StrExtractor` | Current behavior: `str(value)`. Kept for backward-compatibility. |
| `JsonExtractor` | `json.dumps(value, ensure_ascii=False)`, valid JSON, structure preserved. |
| `ProseExtractor` | Flattens nested dicts/lists into readable text (e.g. `"text: ...; tokens: [...]"`). Embedding-friendly. |

Users can also pass a custom callable.

### Open design questions

- **Default extractor.** `StrExtractor` is the backward-compatible choice

Candidates:
  - Keep `StrExtractor` as default; document the constraint clearly.
  - Switch default to `JsonExtractor` (well-formed, structure-preserving, no Python-only artifacts).
  - Default raises on non-scalar input, forcing an explicit choice.
- **Scoping.** Single extractor per loader, or per-column mapping (`content_extractors={"passages": ProseExtractor}`)? Per-column is more flexible but adds API surface.
- **`metadata_columns` interaction.** Metadata stores raw values today, which is correct — extractors should not apply there.

### Acceptance criteria

- [ ] `ContentExtractor` type / protocol defined in a shared module under `railtracks.retrieval`.
- [ ] Three built-in extractors implemented and unit-tested.
- [ ] `HuggingFaceDatasetLoader` accepts a `content_extractor` (and/or per-column map) constructor argument.
- [ ] `JSONLoader` accepts the same.
- [ ] Default behavior decided and documented.
- [ ] Docs updated: ingestion overview, HF page, JSON page, with at least one nested-dataset example (`narrativeqa` or `ms_marco`).
- [ ] Existing tests still pass; back-compat path covered.

### Out of scope

- Per-row transformation hooks (a different abstraction).
- Async extractors — synchronous is enough; loaders already use `to_thread`.
- Auto-detection of "best" extractor based on column type.

### Related

- PR #1098 (HF loader; original discussion thread).
- Touches `JSONLoader` and any future structured-row loader.

### Additional Context

_No response_

be-student:
Closes #1100

- add shared string, JSON, prose, and custom callable extraction support
- apply extractors to Hugging Face and JSON structured content while preserving existing defaults and raw metadata
- document nested `ms_marco` and JSON usage

Validation: 41 focused tests and 147 non-cloud loader tests passed; Ruff, focused mypy, and strict MkDocs passed.

### DIFF
--- a/docs/retrieval/components/ingestion/methods.md
+++ b/docs/retrieval/components/ingestion/methods.md
@@ -125,12 +125,12 @@ For `.jsonl`:
 | Parameter | Type | Default | Description |
 |---|---|---|---|
 | `file_path` | `str` | required | Path to a `.json` / `.jsonl` file, or a directory containing them |
-| `content_keys` | `list[str] | "*"` | `"*"` | Keys whose values form `content`. `"*"` serialises the whole object. |
-| `id_key` | `str | None` | `None` | Top-level key whose value is the per-object id in `Document.source`. Falls back to position (array index or JSONL line). |
-| `ignore_keys` | `list[str] | None` | `None` | Keys dropped entirely |
+| `content_keys` | `list[str] \| "*"` | `"*"` | Keys whose values form `content`. `"*"` serialises the whole object. |
+| `id_key` | `str \| None` | `None` | Top-level key whose value is the per-object id in `Document.source`. Falls back to position (array index or JSONL line). |
+| `ignore_keys` | `list[str] \| None` | `None` | Keys dropped entirely |
 | `content_separator` | `str` | `"\n"` | Used to join content-key values |
 | `encoding` | `str` | `"utf-8-sig"` | File encoding |
-| `content_extractor` | `ContentExtractor | None` | `None` | Converts selected values to text. Omitted preserves existing serialization. |
+| `content_extractor` | `ContentExtractor \| None` | `None` | Converts selected values to text. Omitted preserves existing serialization. |
 
 ---
 
@@ -307,7 +307,7 @@ For gated datasets set `HF_TOKEN` in your environment, or pass
 **Document metadata**: `row_index` plus any column listed in
 `metadata_columns`. `Document.source` is `"{dataset_name}/{split}"`.
 
-### Structured content extractors
+## Structured content extractors
 
 Both `JSONLoader` and `HuggingFaceDatasetLoader` accept any callable matching
 `ContentExtractor`: it receives one selected value and returns a string. The
@@ -315,12 +315,14 @@ built-ins are importable from `railtracks.retrieval`:
 
 - `StrExtractor` uses `str(value)` and is the Hugging Face loader default.
 - `JsonExtractor` produces valid JSON and preserves Unicode text.
-- `ProseExtractor` renders nested dictionaries and lists as labelled prose.
+- `ProseExtractor` renders nested dictionaries and lists as labelled prose, with braces around nested dictionaries and brackets around lists to show their boundaries.
 
 For example, the `passages` column in `ms_marco` is structured. The
 `hf_kwargs` example above uses `ProseExtractor` so the nested labels and values
-become useful embedding input. Extractors apply only to content fields;
+remain visible in the extracted text, including `is_selected` flags and raw URLs. For retrieval focused on passage content, use a custom extractor that selects only `passage_text`. Extractors apply only to content fields;
 `metadata_columns` stay as their original Python values.
+
+Extraction failures stop loading and raise a `ValueError` naming the content column or key, zero-based row or object index, and source; the original exception is preserved as the cause. They do not currently produce per-document recovery events. For `JSONLoader(content_keys="*")`, the reported key is `"*"` because the extractor receives the entire filtered object.
 
 ---
 # LangChain Loaders
--- a/docs/retrieval/components/ingestion/methods.md
+++ b/docs/retrieval/components/ingestion/methods.md
@@ -108,6 +108,12 @@ Each object — array element or JSONL line — becomes one `Document`.
 --8<-- "docs/scripts/retrieval/ingestion_example.py:json_loader"
 ```
 
+Nested values can be normalized before embedding with a content extractor:
+
+```python
+--8<-- "docs/scripts/retrieval/ingestion_example.py:json_structured_content"
+```
+
 For `.jsonl`:
 
 ```python
@@ -124,6 +130,7 @@ For `.jsonl`:
 | `ignore_keys` | `list[str] | None` | `None` | Keys dropped entirely |
 | `content_separator` | `str` | `"\n"` | Used to join content-key values |
 | `encoding` | `str` | `"utf-8-sig"` | File encoding |
+| `content_extractor` | `ContentExtractor | None` | `None` | Converts selected values to text. Omitted preserves existing serialization. |
 
 ---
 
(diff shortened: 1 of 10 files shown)


CoronRing (changes requested):
@be-student Thank you for the contribution, and for the thorough tests and docs that came with it. The overall shape is good: the protocol plus three built-ins is the right abstraction, and I appreciate that both loaders keep the new kwarg last so existing positional callers are unaffected, and that the defaults reproduce the previous output exactly. Metadata staying raw is also the right call.

I ran the branch locally. Ruff is clean, mypy is clean on the changed files aside from the pre-existing missing `datasets` stub, and the retrieval unit tests pass. One note: the four new HuggingFace tests were skipped on my machine because `datasets` was not installed, so CI's `retrieval_tests` job is what will actually exercise those.

I have collected everything in one comment so it is easier to work through. Items 1 to 3 are the ones I would like resolved before merge, the rest are cleanups you can fold in or leave for a follow up, whichever you prefer.

## Requested changes

**1. Make the protocol's parameter positional-only.** `content_extraction.py:9-12` declares `__call__(self, value: Any) -> str`, so a type checker requires every conforming callable to name its parameter literally `value`. That contradicts the docs promising "any callable matching `ContentExtractor`". `content_extractor=str` and `def extract(v): ...` both fail mypy while working fine at runtime. Your tests use a lambda and a class with `__call__(self, value)`, which is why it does not show up. Adding `/` fixes it:

```python
def __call__(self, value: Any, /) -> str: ...
```

**2. Disambiguate nesting in `ProseExtractor`.** `content_extraction.py:34-38` uses the same `; ` and `: ` separators at every depth, so distinct inputs collapse to identical text:

```
{'a': {'b': 1, 'c': 2}, 'd': 3}   ->  'a: b: 1; c: 2; d: 3'
{'a': 'b: 1; c: 2', 'd': 3}       ->  'a: b: 1; c: 2; d: 3'
[{'x': 1}, {'y': 2}]              ->  '[x: 1, y: 2]'
```

Your own test shows it too: `"metadata: reviewed: true; score: null"` gives no signal that those keys belong to `metadata`, and the list case makes two dicts look like one. Lists already get bracket delimiters, so could dicts get something equivalent, `(...)` or `{...}`? It costs nothing at depth one, which is what the documented `ms_marco` example actually produces.

**3. Guard the extractor call so a bad value does not kill the whole ingest.** `runtime.py:248` iterates `loader.astream()` with no exception handling, so unlike chunking and embedding failures, which degrade into per-document `DocumentFailed` events, one bad value tears down the entire run. `JsonExtractor` raises on anything non-serializable, and HuggingFace datasets do produce those: `Image` and `Audio` features decode to PIL and numpy objects, and pandas timestamps are common.

```
TypeError('Object of type datetime is not JSON serializable')
```

There is no row index, column name, or document in that message. Could you wrap the extractor call in both loaders and re-raise with the offending column or key and the row index attached? Full per-document recovery would be a larger change and is fine to leave out of scope, I mainly want the error to say where it happened.

**4. Fold the two copies of the JSONLoader default handling into one.** `json_loader.py:109-113` inlines an `is not None` ternary in the `"*"` branch while `json_loader.py:146-149` has `_extract_content` doing the same check for the explicit-keys branch. One helper taking the per-mode fallback would cover both. Related, the HF loader stores a concrete `StrExtractor()` default while JSONLoader keeps a `None` sentinel. The asymmetry is justified since JSONLoader's default differs per mode, but a one line comment saying so would help, we try to keep sibling components consistent in shape.

**5. Match the docstring parameter order to the signatures.** Both loaders document `content_extractor` before the parameter it actually follows, before `encoding` in JSONLoader and before `dataset_kwargs` in the HF loader. Keeping the argument last in the signature is correct for positional compatibility, so it is the docstring entries that should move.

**6. Documentation placement and a table escape.** The new "Structured content extractors" section is an `###` under `## HuggingFaceDatasetLoader` but documents `JSONLoader` too, and `ingestion.md` deep links into it from prose that is not HuggingFace specific. It would read better as a sibling `##` section covering both. Separately, the new JSONLoader table row writes `` `ContentExtractor | None` `` with an unescaped pipe, which splits the cell. It matches the already broken rows above it while the HuggingFace table escapes as `\|`, so if you are willing, fixing all of them in one pass would be welcome.

**7. Optional, the ms_marco example.** The `hf_kwargs` snippet embeds `is_selected` flags and raw URLs into the same text as the passage content. It demonstrates the feature, but it models a retrieval pipeline most people would not want. A custom extractor pulling just `passage_text` would make a stronger example, or a sentence acknowledging the tradeoff would do.

Happy to discuss any of these if you disagree, particularly 2 and 3 where there is more than one reasonable answer. Thanks again.

Ciao =D

CoronRing:
@be-student Thanks for turning this around. I checked against each point from the last round and all seven are addressed. Three small optional notes more for your knowledge/record keeping. From my side this is good to merge once CI is green.

## Optional notes, none blocking

**A. The new protocol test does not guard the regression it is aimed at.** `test_protocol_accepts_builtin_and_differently_named_callable` asserts `str(42) == "42"` at runtime, which was never in question. The thing that broke was type checking, and the `extractors: tuple[ContentExtractor, ...]` annotation is a local variable annotation, so Python never evaluates it. Our CI only runs mypy over `docs/scripts/`, never over the package or tests, so nothing would catch a reintroduced regression here. That is our gap rather than yours, so please leave the test as is. I mention it only so you are not counting on coverage that is not there.

**B. `JsonExtractor()` is constructed per object.** In `json_loader.py`, the `"*"` branch passes `JsonExtractor()` as the fallback on every call to `_object_to_document`, so it allocates once per document even when a custom extractor is supplied and the fallback is discarded. It is stateless and the cost is negligible, but a module-level constant would read better in a streaming loop.

**C. One long line mid-paragraph.** In `methods.md`, the sentence added to the ms_marco paragraph runs to about 220 characters between two lines wrapped at around 70. The file is genuinely mixed on this so it is not a rule, it just looks odd inside that one paragraph.

One small ask for next time: a short note on the PR saying what you changed saves a reviewer from diffing commits to find out. Not a complaint, the commit itself was easy to follow.

Nice work on this.

Ciao =D

CoronRing (approved):
Great work, all is clear as day on my side! If you enjoyed the framework, please tell your friends and star the repo so it gets a bit of attention. 

- thank & love from maintenance team

PR #1546, opened 2026-09-08
chore(deps): bump pypdf, pymdown-extensions and setuptools

### LINKED ISSUE #1531: Dependency Update
### Affected component

Agents / Nodes

### Description

1. pypdf 6.14.2 → 6.16.2: 
Blocked by a hard cap in the package manifest, not by the lock: pyproject.toml:72 and pyproject.toml:77 both pin pypdf >= 4.0.0, <= 6.14.2. Raise the floor to >= 6.16.1, < 7 in both the pdf and ocr extras, then relock. 

2. pymdown-extensions 10.21.3 → 11.0.2
Transitive only, and only in the docs dependency group (via mkdocs-material==9.6.19 and mkdocs-mermaid2-plugin==1.2.2). mkdocs-material only requires pymdown-extensions>=10.2, so 11.x is permitted.

3. setuptools 80.9.0 → 84.0.0 

### Acceptance criteria

_No response_

CoronRing:
## Summary

Closes #1531. Bumps the three flagged dependencies.

| Package | Before | After | Where the cap was |
| --- | --- | --- | --- |
| `pypdf` | 6.14.2 | 6.18.0 | `packages/railtracks/pyproject.toml`, `pdf` + `ocr` extras |
| `pymdown-extensions` | 10.21.3 | 11.0.2 | transitive, blocked by `mkdocs-material` |
| `setuptools` | 80.9.0 | 84.0.0 | uncapped, lock only |

**`pypdf`** — the floor in both the `pdf` and `ocr` extras goes from `>= 4.0.0, <= 6.14.2` to `>= 6.16.1, < 7`. The loaders only use `PdfReader` / `.pages` / `extract_text()`, which are stable across 4.x-6.x, so no source change was needed.

**`pymdown-extensions`** — one correction to the issue description: it says mkdocs-material "only requires `pymdown-extensions>=10.2`, so 11.x is permitted", but that is only true from **9.7.0** onward. The pinned 9.6.19 requires `pymdown-extensions~=10.2`, which caps at `== 10.*`, so 11.x could not be reached without also moving mkdocs-material. Hence two changes in the `docs` group:

- `mkdocs-material>=9.6.15` -> `>=9.7.0`
- new explicit `pymdown-extensions>=11.0.2,<12`

The explicit entry is deliberate. Left purely transitive, nothing in the manifest stops a future resolution from landing back on 10.x and silently undoing the fix.

**`setuptools`** — nothing constrains it anywhere in the tree, so the relock is the entire fix; no manifest entry added.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [x] Lint & format pass (`ruff check . && ruff format .`) - no Python changed
- [ ] Docs updated if user-facing behavior changed - n/a
- [ ] Breaking changes include migration notes - see note below

## Notes

**Verification**

- `uv lock --check` - lock in sync with the manifests
- `mkdocs build --strict` - passes on mkdocs-material 9.7.7 + pymdown-extensions 11.0.2
- `pytest packages/railtracks/tests/unit_tests` - 2338 passed, 8 skipped

**Two things worth a reviewer's eye**

1. The relock also pulls in `apple-fm-sdk 0.2.1` and moves `mkdocs-mermaid2-plugin` 1.2.2 -> 1.2.3. The `apple` extra was added to the manifest without a relock, so the lock was already stale against `main`; any relock has to pick it up.

2. Raising the `pypdf` floor to 6.16.1 drops 4.x/5.x support for anyone installing `railtracks[pdf]` or `railtracks[ocr]` with an existing older pin. That is the intent of the fix rather than an accident, but it is a narrowing of the supported range.

### DIFF
--- a/packages/railtracks/pyproject.toml
+++ b/packages/railtracks/pyproject.toml
@@ -74,12 +74,12 @@ websearch = [
 # `railtracks[retrieval]` to opt in.
 tiktok = ["tiktoken>=0.10.0, <=0.12.0"]
 semantic = ["scikit-learn>=1.3.0"]
-pdf = ["pypdf >= 4.0.0, <= 6.14.2"]
+pdf = ["pypdf >= 6.16.1, < 7"]
 ocr = [
     "pypdfium2 >= 4.30.0",
     "pytesseract >= 0.3.10",
     "pillow >= 10.0.0",
-    "pypdf >= 4.0.0, <= 6.14.2",
+    "pypdf >= 6.16.1, < 7",
 ]
 glm = [
     "glmocr ~= 0.1.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,9 +22,13 @@ members = [
 [dependency-groups]
 docs = [
     "mkdocs>=1.6.1,<2.0.0",
-    "mkdocs-material>=9.6.15",
+    # mkdocs-material <9.7.0 caps pymdown-extensions at ~=10.2.
+    "mkdocs-material>=9.7.0",
     "mkdocs-material-extensions>=1.3.1",
     "mkdocs-mermaid2-plugin>=1.2.1",
+    # Pulled in transitively by mkdocs-material and mkdocs-mermaid2-plugin;
+    # constrained here so the resolver cannot fall back to a 10.x release.
+    "pymdown-extensions>=11.0.2,<12",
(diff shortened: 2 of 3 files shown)


Aryan-Railtown (approved):
LGTM 🚀

PR #1550, opened 2026-09-08
ci: scope the e2e workflow token to contents: read

### LINKED ISSUE #1549: e2e.yaml is the only workflow without a permissions block
## Problem

`.github/workflows/e2e.yaml` declares no `permissions:` block, so its `GITHUB_TOKEN` inherits whatever the repository default grants rather than the least privilege the job actually needs.

Every other workflow already scopes its token:

- `pr_tests.yaml` — `contents: read`, `pull-requests: read`
- `push_main.yaml` — `contents: read`
- `release_docs.yaml` — `contents: write`
- `release_package.yaml` — `contents: read`, `id-token: write`

## Impact

Low in practice: the workflow is `workflow_dispatch` only and takes no untrusted input. It does hold five provider API keys in its environment though, and it is the one inconsistency in an otherwise uniform set, so the broad token scope is unnecessary risk for no benefit.

## Fix

Add `permissions: contents: read` at the top of the workflow. The job only checks out the repo and runs pytest, so nothing else is needed.

CoronRing:
## Summary

`.github/workflows/e2e.yaml` was the only workflow without a `permissions:` block, so its `GITHUB_TOKEN` inherited the repository default instead of the least privilege the job needs. The job only checks out the repo and runs pytest, so `contents: read` covers it.

Brings it in line with `pr_tests.yaml`, `push_main.yaml`, `release_docs.yaml` and `release_package.yaml`, which all scope their token already.

closes #1549

## Type of change

## Checklist

## Notes

No Python touched. The workflow is `workflow_dispatch` only, so verification is a manual run.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -3,6 +3,9 @@ name: Run E2E Tests (Manual)
 on:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 env:
   PYTHON_VERSION: "3.10"
 


Aryan-Railtown (commented):
Amazing. Lets ship it 🚀

Aryan-Railtown (approved):

CoronRing:
Thanks for the review @Aryan-Railtown

PR #1558, opened 2026-09-10
fix: preserve function node middleware type inference

### LINKED ISSUE #1538: Type hints collapse when you use a list of pre-built middlewares

Lpk124:
## Summary

Fixes #1538.

`Retry`, `Timeout`, `MaxCalls`, and `Lock` can wrap any node signature, so this keeps their `Middleware[Any, Any]` parameterization from #1541. The decorator-factory overload still needs to defer inference until it receives the decorated function without losing the signature checks from verifier and custom middleware.

This patch carries that signature constraint separately from the middleware's runtime parameter and return slots. The prebuilt middleware uses an unconstrained carrier; signature-aware middleware keeps its parameter and output constraints. `function_node(...)` returns a private overloaded callable protocol, which preserves the exact sync or async function signature.

The regression fixture covers empty factories, reusable prebuilt lists, mixed verifier lists, async functions, direct calls, and deliberately invalid verifier signatures. `docs_validation.sh` now uses `--warn-unused-ignores`, so widening the factory back to `Any` makes the negative cases fail CI.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [x] Lint and format pass (`ruff check .` and `ruff format --check .`)
- [x] Tests added and pass locally
- [x] Docs updated with a mypy-checked regression fixture
- [x] No breaking changes

## Notes

The first commit is Christian-Sidak's work from #1541 with the original authorship preserved. The follow-up commit replaces the factory-wide `Any` annotation with the constraint carrier and adds the static coverage.

Local checks:

- mypy 1.19.1: all 48 files under `docs/scripts`
- mypy 2.3.1: `docs/scripts/function_node_typing.py`
- fixture with the expected ignores removed: six errors under both mypy versions
- middleware tests: 104 passed
- Ruff, dependency ordering, the executable fixture, and `git diff --check`

### DIFF
--- a/docs/scripts/function_node_typing.py
+++ b/docs/scripts/function_node_typing.py
@@ -1,0 +1,161 @@
+"""Static regression cases for ``function_node`` middleware inference."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Awaitable, Callable
+
+import railtracks as rt
+from railtracks.middleware import Verdict
+from railtracks.prebuilt.middleware import (
+    Lock,
+    MaxCalls,
+    Retry,
+    Timeout,
+    post_verifier,
+    pre_verifier,
+)
+from typing_extensions import assert_type
+
+neutral_middleware = [
+    Retry(max_tries=1),
+    Timeout(seconds=5),
+    MaxCalls(max_calls=3),
+    Lock(),
+]
+
+
+@rt.function_node()
+def empty_factory(value: str) -> int:
+    return len(value)
+
+
+@rt.function_node(middleware=[])
+def empty_list(value: str) -> int:
+    return len(value)
+
+
+@rt.function_node(middleware=[Retry(max_tries=1)])
+def one_prebuilt(value: str) -> int:
+    return len(value)
+
+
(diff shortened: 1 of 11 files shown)


Aryan-Railtown (changes requested):
Great work overall, this fixes the issue correctly.
One ask: `function_node_typing.py` in docs/scripts/ is really a test file wearing a docs-script costume. Can you move that logic into tests/ instead? If it can't fit cleanly there, fine to just drop it, don't need static-typing test coverage badly enough to keep it in the wrong place.

Aryan-Railtown (approved):
Thanks for sticking with this one, @Lpk124 
Appreciate the depth you put into it. Merging this in.

If you enjoyed working on this, a star on the repo would mean a lot 🙂

Amir-R25 (approved):

PR #1561, opened 2026-09-10
fix(middleware): treat middleware as shared instances so MaxCalls and Lock work in the node slot (#1560)

### LINKED ISSUE #1560: bug: MaxCalls in the `middleware=` slot never limits anything (counter resets each invocation)
## Summary

`MaxCalls` never limits anything when passed to the node-level `middleware=` slot (for any `max_calls > 0`). The middleware *is* wired in and does run, but its counter is reset on every node invocation, so the limit is never reached. The `model_middleware=` slot works correctly within a single run.

This makes `middleware=[MaxCalls(n)]` a silent no-op: it looks like a budget, reports no error, and enforces nothing.

## Repro

Four calls to a tool budgeted at two, inside a single run:

```python
import railtracks as rt
from railtracks.prebuilt.middleware import MaxCalls

@rt.function_node(middleware=[MaxCalls(2, custom_message="tool budget hit")])
def limited(x: str) -> str:
    """A tool with its own budget."""
    return x

@rt.function_node
async def driver(n: int) -> list[str]:
    """Call the limited tool n times in one run."""
    out = []
    for i in range(n):
        try:
            out.append(await rt.call(limited, x=str(i)))
        except Exception as e:
            out.append(type(e).__name__)
    return out

print(rt.Flow("node-slot", entry_point=driver).invoke(n=4))
```

Actual: `['0', '1', '2', '3']` — all four ran.
Expected: the third and fourth raise `MaxCallsExceededError`.

## The middleware is running, the counter is not persisting

`MaxCalls(0)` blocks every call, which rules out "the middleware was never applied":

```python
@rt.function_node(middleware=[MaxCalls(0, custom_message="zero budget")])
def blocked(x: str) -> str: ...
# -> ['MaxCallsExceededError: zero budget', 'MaxCallsExceededError: zero budget']
```

And `MaxCalls` is correct in isolation:

```python
mc = MaxCalls(1)
wrapped = mc.wrap(target)
await wrapped()   # ok
await wrapped()   # MaxCallsExceededError
```

## Also resets between runs, in both slots

`MaxCalls(1)` allows three consecutive `flow.invoke` calls in every slot:

| Slot | 3 sequential runs |
|---|---|
| `agent_node(middleware=[MaxCalls(1)])` | `['ok', 'ok', 'ok']` |
| `agent_node(model_middleware=[MaxCalls(1)])` | `['ok', 'ok', 'ok']` |
| `function_node(middleware=[MaxCalls(1)])` | `['ok', 'ok', 'ok']` |

`model_middleware` does enforce the cap *within* one run: an agent whose tool loop needs two model calls, given `model_middleware=[MaxCalls(1)]`, fails on the second (surfaced as `LLMError`). So the per-run behaviour of that slot is fine; only the node slot is inert and only cross-run persistence is absent.

## Likely cause

Middleware is deep-copied at several layers, so the object that counts is not the object that persists:

- `built_nodes/llm/node_builder.py:91-95` deep-copies both `middleware` and `model_middleware` when the node class is built.
- `nodes/nodes.py:187,197` deep-copy again in `extend_middleware` / `extend_model_middleware`.
- `nodes/nodes.py:251` (`safe_copy`) deep-copies the whole instance `__dict__`.
- `nodes/nodes.py:130` builds the `MiddlewareChain` per node instance, and a fresh node instance is created per invocation.

Any stateful middleware has the same problem, not just `MaxCalls`; `MaxCalls` is where it is visible because its whole purpose is to accumulate.

## The docstring promises behaviour that does not hold

`prebuilt/middleware/max_calls.py` states:

> The count is tracked per `MaxCalls` instance, so a single instance shared across nodes enforces a combined budget, while a fresh instance per node gives each its own limit.

Neither half holds today. Sharing one instance across two agents gives two independent counters, because each `agent_node()` deep-copies it:

```python
budget = MaxCalls(5)
A = rt.agent_node("A", llm=..., model_middleware=[budget])
B = rt.agent_node("B", llm=..., model_middleware=[budget])

A._user_model_middleware[0] is budget                      # False
A._user_model_middleware[0] is B._user_model_middleware[0] # False
```

The caller's own instance also never observes any spend, so a budget cannot be inspected from outside.

## Deciding what to fix

The behaviour and the documentation disagree, so this needs a product call before a code change:

1. **Per-run budgets** (probably the common expectation): stateful middleware needs one instance for the life of a run, so the deep-copy needs a carve-out, and the docstring's cross-node sharing claim should go.
2. **Explicitly shared budgets**: keep the caller's instance by reference so `MaxCalls` can span nodes and runs as documented, which then needs a thread/async-safety story for the counter.
3. **Neither**: if per-invocation reset is intended, `MaxCalls` in the `middleware=` slot should be rejected at construction rather than silently doing nothing, and the docstring needs rewriting.

Whichever way it goes, a regression test asserting the third call of `MaxCalls(2)` raises would have caught this. The current tests in `tests/unit_tests/middlewares/test_prebuilt_max_calls.py` only drive the middleware directly, never through a node.

## Docs affected

The Agent Harness page (`docs/documentation/harness/overview.md`, section 4) describes the two slots' budget semantics and will need updating to match whatever is decided here.

CoronRing:
## Summary

Fixes #1560.

`MaxCalls` in the node-level `middleware=` slot never enforced its cap, for any `max_calls > 0`. `Node.safe_copy()` deep-copies a node on every invocation, so each invocation got a fresh `MaxCalls` clone with a zeroed counter.

The root cause is that middleware instances were treated as per-node payload state rather than as shared policy objects. Fixing that one assumption fixes `MaxCalls`, `Lock`, and any future stateful middleware at the same time.

### 1. Preserve middleware identity across copies

`MiddlewareChain.__deepcopy__` shallow-copies the middleware list, so `Node.safe_copy()` keeps the same middleware instances instead of cloning them. Every other attribute is still deep-copied.

`agent_node` additionally deep-copied the `middleware=` and `model_middleware=` lists at build time (`node_builder.py`). That silently defeated the fix for agents: two agents sharing one `MaxCalls` each got their own clone, `Lock` never serialized between them, and `call_count`/`reset()` on the caller's handle acted on an instance nothing was incrementing. Only the list is copied now, matching `function_node`, whose builder has always used `list(middleware)` with no `deepcopy`.

### 2. Scope the budget per run

Counts are tracked per session id, so a new run starts fresh while all calls within one run — across every node sharing the instance — spend one budget. Overlapping runs keep independent counters.

### 3. Inspection and control

`call_count` and `reset()` act on the live session inside a run. Outside one they fall back to the most recently counted session, so the budget a flow spent is readable after `invoke()` returns rather than reporting zero. Counting itself always uses the live session and never falls back, so a call made outside a run cannot spend a finished run's budget. `reset_all()` clears every session; retained counters are capped at the 64 most recent so a long-lived instance cannot grow without bound.

### 4. Tests and docs

Regression coverage for node-level enforcement, fresh budgets on subsequent runs, genuinely overlapping runs via `asyncio.gather`, post-run inspection, counter eviction, shared budgets across nodes, and instance preservation through `safe_copy()` and the agent builder for both `MaxCalls` and `Lock`. `max_calls.md` documents run scoping, shared budgets, post-run inspection and the retention cap.

## Behavior change

Passing one middleware instance to two `agent_node`s now shares it rather than giving each a private clone. This is the documented intent, and it aligns `agent_node` with `function_node`, which already behaved this way. Passing a fresh instance per node still gives each its own budget, and mutating your list after building a node still has no effect on it.

`Node.extend_middleware` (reachable only through `rt.couple`) still deep-copies, so coupling a node that carries a shared instance produces a clone. Left alone here as a separate, narrower path.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] Breaking change
- [x] Docs

## Checklist

- [x] Lint & format pass (`ruff check packages/railtracks && ruff format --check packages/railtracks`)
- [x] Docs updated

## Verification

- Reproduction from #1560: 4 calls to a node budgeted at 2 return `['0', '1', 'MaxCallsExceededError', 'MaxCallsExceededError']`; the next run starts fresh.
- Unit + integration suites green (2132 passed, 6 skipped), excluding the retrieval/cli/tools/mcp dirs that need optional deps or network.

### DIFF
--- a/docs/documentation/agent_design/middleware/prebuilt/list/max_calls.md
+++ b/docs/documentation/agent_design/middleware/prebuilt/list/max_calls.md
@@ -7,5 +7,42 @@
 ```
 
 Because it only controls the wrapped call, `Max Calls` works in both
-`middleware=` and `model_middleware=`. The limit is enforced on the complete call in
-the selected slot.
+`middleware=` (capping whole-node invocations) and `model_middleware=` (capping raw model calls inside the agent tool loop). The limit is enforced on the complete call in the selected slot.
+
+### Budget Scoping (`per_run`)
+
+By default (`per_run=True`), the budget is scoped to each workflow run / session. When a new run begins (e.g. via `flow.invoke()` or `rt.call()`), the call counter starts fresh.
+
+If you want a cumulative budget across multiple runs for the lifetime of the Python instance, set `per_run=False`:
+
+```python
+# Lifetime budget: allows 10 calls total across all runs
+lifetime_budget = MaxCalls(max_calls=10, per_run=False)
+```
+
+### Shared Budgets Across Nodes
+
+A single `MaxCalls` instance shared across multiple nodes enforces a combined budget:
+
+```python
+shared_budget = MaxCalls(max_calls=5)
+
+ToolA = rt.function_node(func_a, middleware=[shared_budget])
+ToolB = rt.function_node(func_b, middleware=[shared_budget])
+```
+
+Calls to either `ToolA` or `ToolB` accumulate against the same budget of 5.
+
+### Inspecting and Resetting
+
+You can inspect the current call count and reset the counter programmatically:
+
+```python
+budget = MaxCalls(max_calls=3)
+
+# Inspect current spend
+print(budget.call_count)
+
+# Reset counter
+budget.reset()
+```
--- a/docs/documentation/agent_design/middleware/prebuilt/list/max_calls.md
+++ b/docs/documentation/agent_design/middleware/prebuilt/list/max_calls.md
@@ -9,16 +9,10 @@
 Because it only controls the wrapped call, `Max Calls` works in both
 `middleware=` (capping whole-node invocations) and `model_middleware=` (capping raw model calls inside the agent tool loop). The limit is enforced on the complete call in the selected slot.
 
-### Budget Scoping (`per_run`)
+### Workflow Run Scoping
 
-By default (`per_run=True`), the budget is scoped to each workflow run / session. When a new run begins (e.g. via `flow.invoke()` or `rt.call()`), the call counter starts fresh.
+The budget is scoped to each workflow run / session. When a new run begins (e.g. via `flow.invoke()` or `rt.call()`), the call counter starts fresh.
 
-If you want a cumulative budget across multiple runs for the lifetime of the Python instance, set `per_run=False`:
-
-```python
-# Lifetime budget: allows 10 calls total across all runs
-lifetime_budget = MaxCalls(max_calls=10, per_run=False)
-```
 
 ### Shared Budgets Across Nodes
 
--- a/docs/documentation/agent_design/middleware/prebuilt/list/max_calls.md
+++ b/docs/documentation/agent_design/middleware/prebuilt/list/max_calls.md
@@ -11,8 +11,7 @@ Because it only controls the wrapped call, `Max Calls` works in both
 
 ### Workflow Run Scoping
 
-The budget is scoped to each workflow run / session. When a new run begins (e.g. via `flow.invoke()` or `rt.call()`), the call counter starts fresh.
-
+The budget is scoped to each workflow run / session. When a new run begins (e.g. via `flow.invoke()` or `rt.call()`), the call counter starts fresh automatically.
 
 ### Shared Budgets Across Nodes
 
@@ -34,9 +33,12 @@ You can inspect the current call count and reset the counter programmatically:
 ```python
 budget = MaxCalls(max_calls=3)
 
-# Inspect current spend
+# Inspect current spend (scoped to the active session)
 print(budget.call_count)
 
-# Reset counter
+# Reset counter for the current session
 budget.reset()
-```
+
+# Reset counters for all sessions
+budget.reset_all()
+```
--- a/packages/railtracks/src/railtracks/prebuilt/middleware/max_calls.py
+++ b/packages/railtracks/src/railtracks/prebuilt/middleware/max_calls.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
-from typing import Any
+import uuid
+from collections import defaultdict
+from typing import ClassVar
 
 from railtracks.context.central import get_session_identity, is_context_present
 from railtracks.middleware.core import Middleware
@@ -45,6 +47,9 @@ class MaxCalls(Middleware):
             ``max_calls`` times.
     """
 
+    _session_counts: ClassVar[dict[tuple[str, str], int]] = defaultdict(int)
+    _lifetime_counts: ClassVar[dict[str, int]] = defaultdict(int)
+
     def __init__(
         self,
         max_calls: int,
@@ -53,15 +58,10 @@ class MaxCalls(Middleware):
         per_run: bool = True,
     ):
         self._max_calls = max_calls
-        self._call_count = 0
+        self._budget_id = str(uuid.uuid4())
         self._custom_message = custom_message
         self._per_run = per_run
-        self._session_counts: dict[str, int] = {}
         super().__init__(self._middleware_fn)
-
-    def __deepcopy__(self, memo: dict[int, Any]) -> MaxCalls:
-        memo[id(self)] = self
-        return self
 
     @property
     def max_calls(self) -> int:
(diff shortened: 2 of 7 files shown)


packages/railtracks/src/railtracks/prebuilt/middleware/max_calls.py
```diff
@@ -32,18 +37,59 @@ class MaxCalls(Middleware):
...
+            try:
+                return get_session_identity().session_id
+            except ContextError:
+                return None
+        return None
+
+    @property
+    def call_count(self) -> int:
+        """The current number of calls made against this budget in the active session."""
+        return self._session_counts.get(self._current_session_id(), 0)
```
Pooria90, line 69: This will break if the flow is over and we want to inspect the number of calls. I'm not sure if 0 or None is the proper output in this scenario.

packages/railtracks/src/railtracks/prebuilt/middleware/max_calls.py
```diff
@@ -18,9 +22,10 @@ class MaxCalls(Middleware):
             model_middleware=[middleware.MaxCalls(5)],  # cap raw model calls
         )
 
-    The count is tracked per ``MaxCalls`` instance, so a single instance shared
-    across nodes enforces a combined budget, while a fresh instance per node
-    gives each its own limit.
+    The count is tracked per workflow run / session, resetting to zero when a new
```
Pooria90, line 25: Can you make sure that this also works for `agent_node`? Because I think this line will mess up with it:
https://github.com/RailtownAI/railtracks/blob/a2e0dad7f0a045ef37677baac62a1253a1e81bba/packages/railtracks/src/railtracks/built_nodes/llm/node_builder.py#L94

packages/railtracks/tests/unit_tests/middlewares/test_prebuilt_max_calls.py
```diff
@@ -77,4 +77,174 @@ async def broken():
     with pytest.raises(ValueError, match="broken"):
         await max_calls.wrap(broken)()
 
-    assert max_calls._call_count == 1
+    assert max_calls.call_count == 1
+
+
+def test_node_middleware_slot_enforces_limit_inside_single_run():
+    import railtracks as rt
```
Pooria90, line 84: Let's move these local imports to the top.

packages/railtracks/tests/unit_tests/middlewares/test_prebuilt_max_calls.py
```diff
@@ -77,4 +77,174 @@ async def broken():
...
+            try:
+                out.append(await rt.call(tool, x=str(i)))
+            except Exception as e:
+                out.append(type(e).__name__)
+        return out
+
+    flow_a = rt.Flow("session-a", entry_point=driver)
+    flow_b = rt.Flow("session-b", entry_point=driver)
+
+    # Each flow gets its own session, so each gets a fresh budget of 2
```
Pooria90, line 248: These are not concurrent calls; unless the concurrent in docstring means something different. 🤔

packages/railtracks/src/railtracks/prebuilt/middleware/max_calls.py
```diff
@@ -32,18 +37,59 @@ class MaxCalls(Middleware):
...
+        self,
+        max_calls: int,
+        custom_message: str | None = None,
+    ):
         self._max_calls = max_calls
-        self._call_count = 0
         self._custom_message = custom_message
+        # Keyed by session_id (str) when inside a run; None when outside.
+        # Each session gets its own independent counter.
+        self._session_counts: dict[str | None, int] = {}
```
Pooria90, line 49: This will leak the `session_id: count` outside the session context. Should we implement a tear down mechanism for that or we expect the user to reset it manually?

Pooria90 (changes requested):
Thanks for the PR @CoronRing. Very clever idea on the chain!
I have a few change requests.

packages/railtracks/src/railtracks/prebuilt/middleware/max_calls.py
CoronRing: thanks for the detailed review, Dropped deepcopy from the middleware and model_middleware lists (left it on connected_nodes).

packages/railtracks/src/railtracks/prebuilt/middleware/max_calls.py
CoronRing: It should work now/

packages/railtracks/tests/unit_tests/middlewares/test_prebuilt_max_calls.py
CoronRing: Done

packages/railtracks/src/railtracks/prebuilt/middleware/max_calls.py
CoronRing: I don;t think it matters that much as its sort of a optional thing anyways. are both private, and no session id reaches the public API, but it does hold in memory, which I added a teardown for

docs/documentation/agent_design/middleware/prebuilt/list/max_calls.md
```diff
@@ -41,4 +43,6 @@ budget.reset()
 
 # Reset counters for all sessions
 budget.reset_all()
-```
\ No newline at end of file
+```
+
+Read `call_count` outside a run and it reports the run that just finished, so you can check what a flow spent once `invoke()` returns. Counters for finished runs are retained for this, capped at the 64 most recent sessions.
```
Pooria90: This is non-blocker; maybe we can put the 64 in a warning or info box so people can see it clearly.

Pooria90 (dismissed):
Thanks @CoronRing . All good!

docs/documentation/agent_design/middleware/prebuilt/list/max_calls.md
CoronRing: Closed

CoronRing:
@Pooria90 Addressed 64 placing, added a small e2e example.

Pooria90 (approved):

PR #995, opened 2026-03-09
994 gemini llm fix

Amir-R25:
## What does this add?
* Added a `_raw_litellm_message` attribute to the `AssistantMessage` class to optionally store the raw provider message, enabling round-trip of provider-specific metadata (e.g. Gemini's `thought_signature`). (`packages/railtracks/src/railtracks/llm/message.py`)
* Modified `_chat_with_tools_handler_base` to attach the raw provider message to `AssistantMessage`, ensuring metadata is preserved for tool call responses. (`packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py`)
* Updated `_handle_response` in `_tool_call_base.py` to copy provider-specific metadata from the original message to the assistant message before appending to history. (`packages/railtracks/src/railtracks/built_nodes/concrete/_tool_call_base.py`)
* Changed `_to_litellm_message` to return the raw provider message directly if present, allowing automatic round-tripping of metadata when tool calls are serialized. (`packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py`)
## Type of changes

Closes #994 

Please check the type of change your PR introduces:

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update (improvements or corrections to documentation)
- [ ] 🎨 Code style/formatting (changes that do not affect the meaning of the code)
- [ ] ♻️ Refactoring (code change that neither fixes a bug nor adds a feature)
- [ ] ⚡ Performance improvement (code change that improves performance)
- [ ] ✅ Test update (adding missing tests or correcting existing tests)
- [ ] 🔧 Build/CI changes (changes to build process or continuous integration)
- [ ] 🗑️ Chore (other changes that don't modify src or test files)

### Code Quality
- [x] Code follows the project's style guidelines (run `ruff check .` and `ruff format .`)
- [x] Code is commented, particularly in hard-to-understand areas

### Testing
- [x] Tests added/updated and pass locally (`pytest tests`)
- [x] Test coverage maintained

### Documentation
- [x] Documentation updated if needed (bot will verify)

### Git & PR Management
- [x] PR title clearly describes the change

### Breaking Changes
- [x] Breaking changes are documented
- [x] Migration guide provided in documentation (step-by-step instructions for users to update their code/config)

PR #1563, opened 2026-09-10
1431 surface thinking token in llm response

### LINKED ISSUE #1431: Surface thinking token in LLM response.
### Description

Many providers (ie gemini, deepseek, openai in part) support displaying think parameter/tokens, this might be demanded my the user to be returned/accessed. 

Some provider (deepseek, anthropic) expelicitly request thinking token to be returned on the following agent invoke as thinking contains critical decision that otherwise may be lost: https://platform.claude.com/docs/en/build-with-claude/thinking#:~:text=Mid%2Dturn%20conflicts%20degrade%20gracefully.%20If%20you%20toggle,history%20is%20incompatible%20with%20thinking%20being%20enabled.

To do so, wwe much first surface these tokens in llm response, in response to user, and in history objects.

This ticket describe the requirement to SURFACE the tokens

Pooria90:
## Summary

Surface reasoning/"thinking" tokens that providers (Anthropic, DeepSeek, Gemini,
OpenAI in part) return, instead of dropping them. Reasoning now lives on the
`AssistantMessage`, so it rides along in history and serialization, and is reachable via
`Response.reasoning`. Covers chat, structured, tool-calling, and streaming paths.

closes #1431

## Type of change

- [x] Feature
- [x] Docs

## Checklist

- [ ] Breaking changes include migration notes (n/a — additive, defaults to `None`)

## Notes

New surface: `AssistantMessage.reasoning_content: str | None`,
`AssistantMessage.thinking_blocks: list[dict] | None` (signed blocks), and a
`Response.reasoning` convenience accessor. All default to `None` when the provider
returns no reasoning. Round-tripping thinking back to the provider is a follow-up.

Providers differ in what they surface, which is why both fields exist. Anthropic returns
structured `thinking_blocks` carrying a cryptographic `signature` (in addition to the
`reasoning_content` text) — the signed blocks are what a later turn must send back for
thinking to stay valid, so we keep them verbatim. DeepSeek returns `reasoning_content`
text only (no signed blocks); Gemini surfaces its thinking similarly; OpenAI exposes only
partial/summarized reasoning. So `reasoning_content` (and `Response.reasoning`) is the
portable, human-readable view that works everywhere, while `thinking_blocks` is populated
mainly for Anthropic and is the piece the round-trip follow-up will rely on.

Buffered:

```python
model = rt.llm.AnthropicLLM("claude-sonnet-4-5", reasoning_effort="low")
resp = await model.achat(rt.llm.MessageHistory([rt.llm.UserMessage("Why is the sky blue?")]))
print(resp.reasoning)                    # reasoning text, or None
print(resp.message.thinking_blocks)      # signed blocks, or None
```

Streaming (reasoning is attached to the final `Response`, not emitted as chunks):

```python
async for item in model.astream_chat(history):
    if isinstance(item, str):
        print(item, end="")              # answer chunks
    else:
        print(item.reasoning)            # final Response carries the reasoning
```

### DIFF
--- a/packages/railtracks/src/railtracks/llm/message.py
+++ b/packages/railtracks/src/railtracks/llm/message.py
@@ -382,6 +382,11 @@ class AssistantMessage(Message[_T, Role.assistant], Generic[_T]):
         # it back without any manual reconstruction.
         self.raw_litellm_message: Any | None = None
 
+        # Reasoning/"thinking" the model surfaced alongside its answer, when the
+        # provider returns it (Anthropic, DeepSeek, Gemini, OpenAI in part).
+        self.reasoning_content: str | None = None  # human-readable text
+        self.thinking_blocks: list[dict[str, Any]] | None = None  # structured blocks of reasoning, e.g. Gemini's thought_signature
+
     def encode(self):
         encoded = super().encode()
 
@@ -389,6 +394,13 @@ class AssistantMessage(Message[_T, Role.assistant], Generic[_T]):
         # spoke with its calls; surface it the way providers put it on the wire.
         if isinstance(self.content, ToolCalls) and self.content.text is not None:
             encoded["text"] = self.content.text
+
+        # Surface reasoning in the serialized message so it is visible in the run
+        # graph/session data, not just on the live object.
+        if self.reasoning_content is not None:
+            encoded["reasoning_content"] = self.reasoning_content
+        if self.thinking_blocks is not None:
+            encoded["thinking_blocks"] = self.thinking_blocks
 
         return encoded
 
--- a/packages/railtracks/src/railtracks/llm/message.py
+++ b/packages/railtracks/src/railtracks/llm/message.py
@@ -385,7 +385,9 @@ class AssistantMessage(Message[_T, Role.assistant], Generic[_T]):
         # Reasoning/"thinking" the model surfaced alongside its answer, when the
         # provider returns it (Anthropic, DeepSeek, Gemini, OpenAI in part).
         self.reasoning_content: str | None = None  # human-readable text
-        self.thinking_blocks: list[dict[str, Any]] | None = None  # structured blocks of reasoning, e.g. Gemini's thought_signature
+        self.thinking_blocks: list[dict[str, Any]] | None = (
+            None  # structured blocks of reasoning, e.g. Gemini's thought_signature
+        )
 
     def encode(self):
         encoded = super().encode()
--- a/packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
+++ b/packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
@@ -207,6 +207,19 @@ def _to_litellm_tool(tool: Tool) -> Dict[str, Any]:
     return litellm_tool
 
 
+def _attach_reasoning(assistant_msg: AssistantMessage, source: Any) -> AssistantMessage:
+    """Copy any reasoning/thinking the provider returned from a litellm message or
+    delta onto the assistant message.
+
+    `reasoning_content` (text) and `thinking_blocks` (signed structured blocks) are
+    only present on `source` when the provider surfaced reasoning, so both are read
+    with `getattr(..., None)`. Returns the same message for convenient chaining.
+    """
+    assistant_msg.reasoning_content = getattr(source, "reasoning_content", None)
+    assistant_msg.thinking_blocks = getattr(source, "thinking_blocks", None)
+    return assistant_msg
+
+
 class StreamedToolCall(BaseModel):
     tool: ToolCall
     args: str | None = Field(default=None)  # accumulating string of arguments (in json)
@@ -548,6 +561,11 @@ class LiteLLMWrapper(ModelBase, ABC):
 
         """
         accumulated_content = ""
+        # Reasoning/"thinking" arrives in its own deltas (before the content deltas)
+        # on providers that surface it; accumulate the text and collect the signed
+        # blocks so the final Response carries them like a buffered call would.
+        accumulated_reasoning = ""
+        accumulated_thinking_blocks: List[dict] = []
 
         # fall back on empty message info if we don't get one from the stream.
         message_info = MessageInfo()
@@ -575,6 +593,13 @@ class LiteLLMWrapper(ModelBase, ABC):
                 stream_finished = True
                 continue
 
+            reasoning_delta = getattr(choice.delta, "reasoning_content", None)
+            if reasoning_delta:
+                accumulated_reasoning += reasoning_delta
+            thinking_delta = getattr(choice.delta, "thinking_blocks", None)
+            if thinking_delta:
+                accumulated_thinking_blocks.extend(thinking_delta)
+
             if choice.delta.tool_calls:
                 # Some providers batch several calls into a single delta, so every entry
                 # has to be consumed rather than just the first.
@@ -599,6 +624,8 @@ class LiteLLMWrapper(ModelBase, ABC):
             tools=tools,
             output_schema=output_schema,
             message_info=message_info,
+            reasoning_content=accumulated_reasoning or None,
+            thinking_blocks=accumulated_thinking_blocks or None,
         )
 
         yield r
@@ -611,11 +638,15 @@ class LiteLLMWrapper(ModelBase, ABC):
         tools: list[ToolCall],
         output_schema: type[BaseModel] | None,
         message_info: MessageInfo,
+        reasoning_content: str | None = None,
+        thinking_blocks: list[dict] | None = None,
     ):
         """
         From the provided content, creates a completes a response object dyanmically.
 
         This function handles the normalization of the different response `content` types.
+        Any `reasoning_content`/`thinking_blocks` accumulated from the stream are attached
+        to the built assistant message.
         """
         structured_response: BaseModel | None = None
 
@@ -623,24 +654,17 @@ class LiteLLMWrapper(ModelBase, ABC):
             structured_response = output_schema(**json.loads(accumulated_content))
 
         if structured_response is not None:
-            r = Response(
-                message=AssistantMessage(content=structured_response),
-                message_info=message_info,
-            )
+            message = AssistantMessage(content=structured_response)
         elif len(tools) > 0:
-            r = Response(
-                message=AssistantMessage(
-                    content=ToolCalls(tools, text=accumulated_content or None)
-                ),
-                message_info=message_info,
+            message = AssistantMessage(
+                content=ToolCalls(tools, text=accumulated_content or None)
             )
         else:
-            r = Response(
-                message=AssistantMessage(content=accumulated_content),
-                message_info=message_info,
-            )
+            message = AssistantMessage(content=accumulated_content)
 
-        return r
+        message.reasoning_content = reasoning_content
+        message.thinking_blocks = thinking_blocks
+        return Response(message=message, message_info=message_info)
 
     def _is_stream_finished(self, choice) -> bool:
         """Check if the stream has finished.
@@ -781,8 +805,11 @@ class LiteLLMWrapper(ModelBase, ABC):
     # ================ START Base Handlers ==================
 
     def _chat_handle_base(self, raw: ModelResponse, info: MessageInfo):
-        content = raw["choices"][0]["message"]["content"]
-        return Response(message=AssistantMessage(content=content), message_info=info)
+        message = raw["choices"][0]["message"]
+        assistant_msg = _attach_reasoning(
+            AssistantMessage(content=message["content"]), message
+        )
+        return Response(message=assistant_msg, message_info=info)
 
     def _structured_handle_base(
         self,
@@ -790,9 +817,10 @@ class LiteLLMWrapper(ModelBase, ABC):
         info: MessageInfo,
         schema: Type[BaseModel],
     ) -> Response:
-        content_str = raw["choices"][0]["message"]["content"]
-        parsed = schema(**json.loads(content_str))
-        return Response(message=AssistantMessage(content=parsed), message_info=info)
+        message = raw["choices"][0]["message"]
+        parsed = schema(**json.loads(message["content"]))
+        assistant_msg = _attach_reasoning(AssistantMessage(content=parsed), message)
+        return Response(message=assistant_msg, message_info=info)
 
     def _chat_with_tools_handler_base(
         self, raw: ModelResponse, info: MessageInfo
@@ -805,10 +833,11 @@ class LiteLLMWrapper(ModelBase, ABC):
         if choice.finish_reason == "stop" and not choice.message.tool_calls:
             # litellm types content as str | None, but a plain "stop" completion always
             # carries (possibly empty) text content.
-            return Response(
-                message=AssistantMessage(content=cast(str, choice.message.content)),
-                message_info=info,
+            assistant_msg = _attach_reasoning(
+                AssistantMessage(content=cast(str, choice.message.content)),
+                choice.message,
             )
+            return Response(message=assistant_msg, message_info=info)
 
         calls: List[ToolCall] = []
         for tc in choice.message.tool_calls or []:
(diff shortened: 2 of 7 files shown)


packages/railtracks/tests/unit_tests/llm/models/test_litellm_wrapper.py
```diff
@@ -663,14 +663,27 @@ def _gen():
...
-                delta=Delta(content=content, tool_calls=tool_calls),
+                delta=Delta(
+                    content=content,
+                    tool_calls=tool_calls,
+                    reasoning_content=reasoning_content,
+                    thinking_blocks=thinking_blocks,
+                ),
                 finish_reason=finish_reason,
             )
         ],
```
CoronRing, line 689: _to_litellm_message has no branch emitting reasoning_content/thinking_blocks; I verified that a plain assistant turn goes out as {'role': 'assistant', 'content': '4'} and a tool-call turn as role/content/tool_calls only. The sole provider-extras hook is the raw_litellm_message copy in the tool_calls branch (_litellm_wrapper.py:1000) and that is never populated, because llm_helpers.py:187 reads getattr(response.message, "raw_content", None) while the attribute is raw_litellm_message. raw_content appears nowhere else in src.

packages/railtracks/src/railtracks/llm/message.py
```diff
@@ -382,6 +382,13 @@ def __init__(self, content: _T):
         # it back without any manual reconstruction.
         self.raw_litellm_message: Any | None = None
 
+        # Reasoning/"thinking" the model surfaced alongside its answer, when the
+        # provider returns it (Anthropic, DeepSeek, Gemini, OpenAI in part).
+        self.reasoning_content: str | None = None  # human-readable text
+        self.thinking_blocks: list[dict[str, Any]] | None = (
+            None  # structured blocks of reasoning, e.g. Gemini's thought_signature
```
CoronRing, line 389: Claude believes Gemini puts it in provider_specific_fields["thought_signatures"] (vertex_ai/gemini/transformation.py:929-948), not thinking_blocks. A Gemini round-trip needs provider_specific_fields. 

I feel litellm will handle this, but good to add a test/experiment to confirm.

packages/railtracks/src/railtracks/llm/message.py
```diff
@@ -390,6 +397,13 @@ def encode(self):
         if isinstance(self.content, ToolCalls) and self.content.text is not None:
             encoded["text"] = self.content.text
 
+        # Surface reasoning in the serialized message so it is visible in the run
+        # graph/session data, not just on the live object.
+        if self.reasoning_content is not None:
```
CoronRing, line 402: Claude belive that sinc litellm 1.89's Message and Delta both carry reasoning_items alongside reasoning_content/thinking_blocks, reasoning_items is ignored, respecting that's the OpenAI Responses-API round-trip field, need to confirm/a bit of rework

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
```diff
@@ -575,6 +593,13 @@ def _stream_handler_base(
                 stream_finished = True
                 continue
 
+            reasoning_delta = getattr(choice.delta, "reasoning_content", None)
+            if reasoning_delta:
+                accumulated_reasoning += reasoning_delta
+            thinking_delta = getattr(choice.delta, "thinking_blocks", None)
+            if thinking_delta:
+                accumulated_thinking_blocks.extend(thinking_delta)
```
CoronRing, line 601: ### Streamed thinking_blocks accumulate as unusable fragments.
litellm's Anthropic streaming handler (chat/handler.py:667-679) emits one single-element thinking_blocks list per delta: each a text fragment with signature="", and the real signature arriving in a final block whose thinking is "". This function yields N partial blocks, not one. I am not sure how litellm handles this, we can handle or get some a test/check so we confirm.  The test feeds a single complete block, so it passes while masking this. 

A accumulated_thinking_blocks.extend(...) yields, for a four-delta stream:


[{...'thinking': 'Let ',   'signature': ''},
 {...'thinking': 'me ',    'signature': ''},
 {...'thinking': 'count.', 'signature': ''},
 {...'thinking': '',       'signature': 'ErUBCkYIBx'}]
Anthropic won't accept that back. 

Possible: a merge step per content block: join thinking fragments, take the last non-empty signature, and pass redacted_thinking blocks through untouched.

Claude propose this:
```

Suggested fix, with no reassembly logic on our side. litellm already has the merge but keeps it: complete_streaming_response at streaming_handler.py:1977 feeds only usage, the cache and the logger, and only usage is copied onto what the consumer receives. But CustomStreamWrapper retains every chunk in self.chunks (:178, :1055, :1939) — and raw in _stream_handler_base is that wrapper. So after draining, the public litellm.stream_chunk_builder(chunks=raw.chunks) gives you the assembled message; take thinking_blocks off it. I confirmed it returns exactly [{'type':'thinking','thinking':'Let me count.','signature':'ErUBCkYIBx'}] for the fragmented stream above. Worth wrapping so a litellm change degrades to "no blocks" rather than raising mid-stream.

That also settles the unsigned-provider question (Mistral magistral never signs) by deferring to litellm rather than encoding our own policy — it drops unsigned blocks, and reasoning_content still carries the text either way.
```

Which is fair, depend on where its done as we don;t want litellm to "spread"

CoronRing (changes requested):
Nice, careful PR — the getattr reads are exactly right (litellm deletes reasoning_content/thinking_blocks from Message rather than leaving them None, so attribute-probing is the only correct way), thinking_blocks really are TypedDicts so list[dict] is accurate and JSON-safe, and the _prepare_response refactor is a genuine simplification. The buffered path is exact pass-through and I'd merge that as-is.


Do note goal of the ticket is not only to return, but also include thinking args in the outgoing meessage. 


llm_helpers.py:158, :161, and :181 each build a fresh AssistantMessage(returned_mess.message.content). The new fields live on the response message, which is then discarded. So under rt.agent_node including the multi-turn tool loop, which is exactly where Anthropic requires thinking back, the reasoning is gone on the first hop. It only survives if the caller hand-appends response.message themselves. 

.

packages/railtracks/tests/unit_tests/llm/models/test_litellm_wrapper.py
Pooria90: Both confirmed. The attribute was `raw_litellm_message` and the reader said `raw_content`, so the raw copy never fired; fixed. And there was no reasoning branch at all for plain assistant turns; added `_echo_reasoning()` which runs for every `AssistantMessage`, emitting `thinking_blocks`/`reasoning_content`. Confirmed litellm's Anthropic transform converts an incoming `thinking_blocks` into a signed thinking content block (and ignores `reasoning_content`), so the round-trip actually works now, not just for the tool-call branch.

packages/railtracks/src/railtracks/llm/message.py
Pooria90: The Gemini round-trip works, and litellm handles it as you expected. I ran a real Gemini tool-calling loop (gemini-2.5-flash, reasoning_effort=low) and captured what we actually send back on the follow-up request: Gemini accepted the echoed assistant turn with no provider error, so the round-trip is valid. The echoed turn carried the thought metadata three ways at once: `provider_specific_fields`, `thinking_blocks`/`reasoning_content`, and Gemini's own thought signature embedded in the tool-call id (call_…__thought__CvoB…).
We don't encode any Gemini-specific policy for this: the `raw_litellm_message` copy in `_to_litellm_message` forwards `provider_specific_fields` verbatim, so litellm does the reassembly.

packages/railtracks/src/railtracks/llm/message.py
Pooria90: Confirmed, and it's a no-op for us right now. `reasoning_items` is real on Message/Delta in litellm 1.89, but it's defined in litellm's Responses-API transformation (`completion_extras/litellm_responses_transformation/`), not Chat Completions. railtracks calls litellm.completion (Chat Completions), and I verified live with gpt-5-mini: across a plain reasoning call and a tool loop, the raw OpenAI messages carried no reasoning_items (nor reasoning_content/thinking_blocks). OpenAI doesn't expose reasoning on the Chat Completions path at all, regardless of effort.

Our tool-call branch would already forward it via the `raw_litellm_message` passthrough if it were ever populated. If we add first-class `reasoning_items` support (a field on `AssistantMessage` mirroring `thinking_blocks`, echoed in `_to_litellm_message`), it should ride in alongside Responses-API routing as a follow-up rather than this PR. 

Issue: https://github.com/RailtownAI/railtracks/issues/1576

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
Pooria90: Reproduced exactly: `accumulated_thinking_blocks.extend(...)` yields four unsigned fragments Anthropic rejects. Fixed by deferring to litellm's own merge: I collect the raw chunks during the drain and pass them to `litellm.stream_chunk_builder(chunks=...)`, then take `thinking_blocks` off the assembled message. Wrapped in try/except so a litellm change degrades to "no blocks" rather than raising mid-stream, and it settles the unsigned-provider case (litellm drops unsigned blocks; `reasoning_content` still carries the text). I collect chunks locally rather than reading `raw.chunks` so it doesn't couple to the wrapper internals. Updated the test to feed fragmented deltas; the old single-block test was masking this.

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
```diff
@@ -594,53 +630,77 @@ def _stream_handler_base(
...
+        """
+        if not raw_chunks:
+            return None
+        try:
+            built = litellm.stream_chunk_builder(chunks=raw_chunks)
+            if built is None:
+                return None
+            blocks = getattr(built.choices[0].message, "thinking_blocks", None)
+            return blocks or None
+        except Exception:
```
CoronRing, line 667: - Is this safe to include? would it not swallow actual errors? Maybe we should at least log it?

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
```diff
@@ -594,53 +630,77 @@ def _stream_handler_base(
             active_tool_calls, completed_tool_calls
         )
 
+        # Anthropic streams thinking as unsigned per-delta fragments with the signature
+        # in a final block, so we let litellm's merge reassemble them from the raw chunks
+        # rather than concatenating ourselves (see `_assemble_thinking_blocks`).
+        thinking_blocks = self._assemble_thinking_blocks(raw_chunks)
```
CoronRing, line 636: _assemble_thinking_blocks runs unconditionally on every streamed call, this might not be nessasory, a  flag should makes it free for non-reasoning streams

packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
```diff
@@ -154,10 +154,10 @@ async def llm_invoke(
             path = process_message(returned_mess, schema)
 
             if path == "Content":
-                message_history.append(AssistantMessage(returned_mess.message.content))
+                message_history.append(returned_mess.message)
```
CoronRing, line 157: `AssistantMessage(returned_mess.message.content)` built a
fresh message from the content alone and dropped every other attribute on the response
message. That was lossless when `AssistantMessage` was just a content wrapper, which it
was when the line was written. It became silently lossy as attributes accrued. first
`raw_litellm_message` (#995), now `reasoning_content`/`thinking_blocks`, so the rebuild
was discarding exactly what this PR adds. Appending the original is the right call.

**1. Two strategies for one problem, ten lines apart.** `run_tools` below can't append
the original, because `Message.tool_calls` deep-copies so the loop can hand calls to
`invoke_tools` without history sharing mutable argument dicts. So it rebuilds and
hand-copies three fields instead:

    hist_msg.raw_litellm_message = raw
    hist_msg.reasoning_content = response.message.reasoning_content
    hist_msg.thinking_blocks = response.message.thinking_blocks

That copy list is the fragile half. Add a fourth attribute to `AssistantMessage` and the
Content/Structured paths pick it up for free while the tool path silently drops it, 
the same failure mode this PR just fixed, re-armed for the next person. A
`copy_metadata_from(other)` on `AssistantMessage` used by both paths would close it; at
minimum, a note on the class saying new attributes must be added to that copy block but advised fix.

**2. It flips the copy semantics, opposite to its neighbour.** History now holds the same
object the `Response` holds. `post_llm` middleware receives that `Response`, so anything
mutating `response.message` now mutates history too; the old rebuild prevented (not sure if by
accident). So `run_tools` deliberately isolates history from the response while this path
deliberately shares it, I like the copy a bit better, but worth
picking one on purpose and a small text saying so in a comment.

Unifying is possible if you'd rather have one strategy: `response.message.content` is
already a `ToolCalls` carrying the same calls and text, so `run_tools` could append
`response.message` directly and pass the deep-copied `tool_calls` to `invoke_tools`
separately. That trades away the isolation in (2), which is the tradeoff to weigh.

CoronRing (commented):
All past comments resolved but a bit more complexity raise when I ran tests agains, more like discussion in design rather then hard fix. Let me know if you want to discuss any item further.

packages/railtracks/tests/unit_tests/llm/models/test_litellm_wrapper.py
CoronRing: Super! thanks for confimring

packages/railtracks/src/railtracks/llm/message.py
CoronRing: Awsome! one more triage to go!

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
CoronRing: See followup, if we wnat to log it somehow, or a new ticket as I know there is plan to update loggers, might be beyond this ticket.

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
Pooria90: Done. Added a `saw_thinking_blocks` flag set during the drain, and the reassembly only runs when a thinking delta was actually seen. Non-reasoning streams skip `stream_chunk_builder` entirely now. Added a test that makes reassembly raise and confirms a plain stream still completes.

packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py
Pooria90: Kept the broad catch (a reassembly failure should drop only the thinking blocks, never the answer or the stream) but it's no longer silent: it logs at debug with `repr(e)`.  I used `logging.getLogger("RT.litellm_wrapper")`, which still nests under the RT logger tree; same pattern as `context_injection_utils.py`. Happy to move to a fuller logging approach whenever the logger overhaul lands.

packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
Pooria90: Both addressed. Added `AssistantMessage.copy_metadata_from(other)` as the single place that carries the non-content metadata, with a class note that new attributes must go there; so the tool path can't silently fall behind again. And I unified on isolation (Option A): Content/Structured now rebuild a fresh message via the helper too, matching `run_tools`, with a comment stating that history is deliberately isolated from the `Response` that `post_llm` may mutate. New tests cover the helper and the `run_tools` preserve-and-isolate behavior.

CoronRing (dismissed):
Every item from the last round is addressed, and addressed really well!

One tiny things, will approve its anyways:

- Unbalanced paren in the new comment above the logger in `_litellm_wrapper.py`: `# ... upward (the llm package is deliberately isolated.`

- `raw_chunks` is still retained unconditionally, but the gate removed the CPU cost and the memory alone is the lesser half. Fine as is.

Great edits!

CoronRing (approved):

PR #1502, opened 2026-08-27
fix: preserve context injection templates across invocations

mikemikimike:
## Summary

Fixes #1378

Reusing an agent with context injection could reuse the first rendered system prompt on later invocations. The first call mutated the shared system message template in place, so a later context value was ignored.

## Changes

- Copy the system message before adding it to each model-call history.
- Add a regression test that invokes one agent twice with different context values.

## Tests

- `uv run --frozen --no-sync pytest -q --basetemp <unique temporary directory> packages/railtracks/tests/unit_tests/utils/test_context_injection.py packages/railtracks/tests/unit_tests/prompt/test_prompt_placeholder_filling.py packages/railtracks/tests/unit_tests/prompt/test_prompt.py packages/railtracks/tests/unit_tests/middlewares/test_prebuilt_context_injection.py packages/railtracks/tests/unit_tests/built_nodes/llm/test_model_invoker.py packages/railtracks/tests/integration_tests/middlewares/test_middleware_integration.py -k "not TestGuardrailsEndToEnd"` — 103 passed, 8 deselected.
- `uv run --frozen --no-sync pytest -q --basetemp <unique temporary directory> --maxfail=3 --tb=short --ignore=packages/railtracks/tests/unit_tests/retrieval packages/railtracks/tests/unit_tests/ packages/railtracks/tests/integration_tests/` — 1102 passed; three failures outside this change are listed below.
- `uv run --frozen --no-sync ruff check . --no-fix` — passed.
- `uv run --frozen --no-sync ruff format --check .` — passed; 613 files already formatted.
- `uv run --frozen --no-sync mypy --ignore-missing-imports packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py` — reports two existing protocol errors at lines 35 and 42.
- `uv build --package railtracks --wheel --out-dir <external build directory>` — passed.
- `git diff --check` — passed.

The three unrelated full-run failures are:

- `test_get_session_rejects_symlinks_outside_sessions_directory` and `test_ui_rejects_symlinks_outside_selected_bundle` cannot create symlinks on this Windows host because of `WinError 1314` (missing symlink privilege).
- `test_builtin_function_raises_runtime_error` does not raise on Python 3.13 because `inspect.signature(time.sleep)` returns `(object, /)`.

## Compatibility / Known limitations

The fix is backward-compatible, does not add dependencies, and does not require an LLM or API key. The mypy findings are in untouched protocols; this change does not alter them. AI assistance was used during preparation.

PR #1564, opened 2026-09-10
fix: keep caller system messages in the returned history

### LINKED ISSUE #1350: Define semantics for multiple system messages
## Problem

railtracks has no position on histories containing more than one system turn.

- `append_system_message` unconditionally does `message_history.insert(0, system_message)`
  ([llm_helpers.py:313](packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py#L313-L319)),
  so passing `system_message=` alongside a history that already has one yields
  two system turns the kwarg silently shadowing the existing one.
- `check_message_history` only warns about *zero* or *only* system messages; it
  has no branch for several
  ([validation.py:13](packages/railtracks/src/railtracks/validation/node_invocation/validation.py#L13-L37)).

OpenAI-format histories legally contain several system turns, so this is a real
input shape, however, Anthropic Claude API (Native SDK)+gemini sdk: Does not accept system messages inside the messages array. Anthropic uses a dedicated top-level system parameter, similarly with google.

## Proposal
Act Accordingly based on expecation, Supporters:
- **Ordered system turns**: preserve position and pass them through, for
  providers that accept mid-conversation system messages. Document well.
- Document what happens when system message is offered both as user_input and agent parameter (agent parameter inserted at position 0). 
Non supporters:
- Merge system message by appendinge each other.

CoronRing:
## Summary

Closes #1350.

`prepare_string_response` and `prepare_structured_response` returned `message_history.removed_system_messages()`, which drops **every** system message from `response.message_history`. A `SystemMessage` the caller placed in the history was destroyed on the way out, so feeding a response back for the next turn silently lost the prompt. It never showed up in tests because an agent-level `system_message=` is re-inserted on every call, which masks the loss.

That stripping is also what made the unconditional `insert(0, ...)` on the request path idempotent, so the two paths have to change together: remove the stripping alone and an agent's prompt accumulates one copy per turn.

`llm_invoke` now keeps one caller-owned `conversation` and builds a throwaway `wire = [agent system message, *conversation]` for each model call. The agent's message is configuration, so it lives only on the wire and never in the conversation: nothing to strip on the way out, nothing to deduplicate on the way in. Error paths report the wire, since they want to surface what was actually sent.

| You provide | Sent | Returned |
|---|---|---|
| `system_message=` only | `[agent prompt, ...history]` | history, no agent prompt |
| A `SystemMessage` in the history | unchanged | unchanged, prompt intact |
| Both | `[agent prompt, ...history]` | history, caller's prompt intact |
| Either, over N turns | one agent prompt, every turn | no accumulation |

`check_message_history` now runs after the agent's message is in place and scans the whole history, which fixes two smaller defects: it only inspected index 0, so a system message anywhere else drew a false "no system message" warning, and an all-system history of length >= 2 passed every branch and reached the provider with no user turn. It also gains one warning for a system message that follows a non-system message, since that is the shape whose meaning changes between providers: OpenAI keeps it in place, while Anthropic and Gemini hoist every system message into a top-level system field and drop it from the message list.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

## Notes

`response.message_history` now contains caller-supplied system messages where it previously contained none. Anything asserting on the exact contents of a returned history will see the difference. The agent's own prompt is still excluded, so handing a history to a different agent gives that agent's prompt rather than both.

Model middleware that *structurally* mutates the history (append/insert, as opposed to editing message content) now has that edit scoped to the single model call it ran for, rather than persisting into the returned history. Nothing shipped does this; the prebuilt PII guards return a replacement `MessageHistory` through `GuardrailDecision.transform_messages`, which was already per-call.

`check_message_history` loses its `system_message` parameter and runs once before the loop rather than per model call. `prepare_message_history` and `append_system_message` are gone; both were internal to `llm_helpers.py`.

New docs page at `documentation/agent_design/llms/system_messages.md` covering precedence, round-tripping, and the per-provider translation differences, with the examples in `docs/scripts/documentation/system_messages.py`.


docs/documentation/agent_design/llms/system_messages.md
```diff
@@ -0,0 +1,72 @@
+# System Messages
+
+A system message can reach a model two ways in Railtracks, and they behave differently on purpose. The `system_message=` argument on `rt.agent_node` is part of the **agent's configuration**: it describes who that agent is, and Railtracks sends it on every invocation. A `SystemMessage` you place in a `MessageHistory` is part of the **conversation**: it belongs to the caller, and Railtracks leaves it exactly where you put it.
+
+```python
```
Amir-R25, line 5: inline code in docs should go into the scripts folder

docs/documentation/agent_design/llms/system_messages.md
```diff
@@ -0,0 +1,72 @@
...
+| Both | `[agent prompt, ...history]`, the history's own system messages untouched |
+| Several `SystemMessage`s in the history | All of them, in the order you wrote them |
+
+Passing both is not a conflict and neither one is discarded. The agent's prompt goes first because it establishes the agent's identity, and the conversation's own system turns follow in place.
+
+## What comes back
+
+`response.message_history` is the conversation, so it contains every system message the caller put there, and **not** the agent's own configured prompt. That makes the returned history safe to feed straight back in for the next turn.
+
+```python
```
Amir-R25, line 41: same as above

packages/railtracks/src/railtracks/llm/message.py
```diff
@@ -358,6 +358,20 @@ def __init__(self, content: str):
         super().__init__(content=content, role=Role.system)
 
 
+class _AgentSystemMessage(SystemMessage):
```
Amir-R25, line 361: Adding a type that adds no functionality that inherits another class is a bit of a code smell

packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
```diff
@@ -360,11 +368,24 @@ def create_message_history(
...
+    history it returned never accumulates copies of its prompt. A fresh copy is inserted on every
+    call because the node-level message is shared across invocations while prompt injection fills
+    templates in place.
+
+    Args:
+        message_history: The history to modify in place.
+        system_message: The agent node's configured system message, if it has one.
+    """
+    message_history[:] = [
+        m for m in message_history if not isinstance(m, _AgentSystemMessage)
```
Amir-R25, line 384: can't you use the method you introduced `without_agent_system_messages` to do this?

packages/railtracks/src/railtracks/llm/message.py
Amir-R25: I added a review comment to discuss this

Amir-R25 (changes requested):
Thanks Guan.

This fixes the bug you documented, I commented some nits but my discussion below is what I think we should fix.

To address "the framework's prompt not appearing in returned history", this PR adds three pieces that need to coordinate: `_AgentSystemMessage`, `MessageHistory.without_agent_system_messages()`, and the extra strip step in `append_system_message`.

A couple of issues with that shape:

- `MessageHistory` (a collection type) now has a public method whose docstring has to explain what an agent system message causing a private/internal type from another module.
- Anyone who rebuilds a message downstream,`SystemMessage(msg.content)` in a middleware or guardrail, for instance drops the `_AgentSystemMessage` without noticing. The framework then treats what used to be its own prompt as a caller message, and the accumulation bug is back.

I think we can sidestep all of this with two changes in `llm_invoke`: replace `prepare_message_history` with a caller/agent split, and build a fresh wire history at the top of each loop iteration to hand to the model. In the error paths `message_history` becomes `wire`, since those errors want to surface what was actually sent.

That deletes `_AgentSystemMessage`, deletes `without_agent_system_messages`, and collapses `append_system_message` and `prepare_message_history` to almost nothing. The invariant becomes something the code shape makes true instead of something a filter has to enforce.

CoronRing:
@Amir-R25 I weighted alot between creating a temp object like wire or modifying message history on the spot. The decision that got me to manipulate directly is the feature that if we wire, we need to copy the object, which lead to modification of the prompt object directly to be swallowed. 

right now:
1. Pass system + user
2. Go through the guard, modify as we like. If temp, we can just clone the prompt; if permanent, we directly change the object
3. prompt made ...

But one thing you brought up is very real: we are touching the user prompt object, and it might break in complex concurrency, so your solution is more robust.

CoronRing:
`llm_invoke` keeps one caller-owned `conversation` and builds `wire = [agent system message, *conversation]` per model call. `_AgentSystemMessage`, `without_agent_system_messages`, `append_system_message` and `prepare_message_history` are all gone, and `llm/message.py` and `llm/history.py` are no longer touched by this PR at all. Error paths report the wire.

Two things I added that weren't in the sketch:

- **The prompt is copied on every wire build.** `ContextInjection` fills `{placeholder}` templates in place, and #1502 fixed exactly that by deepcopying; reusing the shared node-level object on the wire would bring it straight back. Building per call is better than what we had: the template is re-filled from current context on each pass of the tool loop instead of being consumed on the first.
- **Validation runs once, before the loop.** Called at the top of each iteration it would log the same warning on every pass of a tool-calling agent. There's a test pinning it to one call per invocation.

One correction to the reasoning, for the record: I went looking for code that rebuilds a message and found `PIIRedactInputGuard`, which does exactly that and explicitly scans `Role.system`. It clones with `deepcopy(msg)` though, which preserves the subclass, so the marker would actually have survived it. The concern was right in principle but wasn't a live bug — the shipped guard survives by accident of using `deepcopy` rather than by design, which I think still argues for the structural fix.

Docs examples moved to `docs/scripts/documentation/system_messages.py`.

packages/railtracks/src/railtracks/built_nodes/llm/llm_helpers.py
CoronRing: No longer relavent

docs/documentation/agent_design/llms/system_messages.md
CoronRing: Done!

packages/railtracks/src/railtracks/llm/message.py
CoronRing: Removed (I github cannot do live detect of outdated code, hard to find)

Amir-R25 (approved):
Thanks, looks great

CoronRing:
> Thanks, looks great

What happened to my memes! you seem not confident without it   O_o

Will run one more integration test before merge, should be in EoD

Amir-R25:
my bad, I forgot

image

PR #1565, opened 2026-09-11
fix(retrieval): scope the staleness check and delete-before-write to the ingesting tenant

### LINKED ISSUE #1183: [Bug] [Retrieval] Scope ignored by staleness check and document replacement — cross-tenant clobbering
### Description

`RetrievalRuntime` advertises multi-tenancy ("multi-tenant callers share one runtime and pass `scope` per ingest/retrieve call"), but neither the duplicate check nor the replace-on-reingest path is scope-aware:

1. `_is_complete_duplicate` filters only on `source_path` + `content_hash`:
https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/retrieval/runtime.py#L313-L316
If tenant A has ingested a document, tenant B ingesting the same source/content gets `DocumentSkipped` — B's scope never receives the entries, and B's scoped `retrieve()` silently misses the document.

2. `_embed_and_store` clears the prior version with `delete_where({"document_id": str(doc.id)})`:
https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/retrieval/runtime.py#L386
`Document.id` is uuid5-deterministic from `source`, so tenant B re-ingesting changed content **deletes tenant A's entries** and rewrites them under B's scope.

```python
runtime = RetrievalRuntime(chunker, embedder, store)
doc = Document(source="shared/doc.md", content="hello world")

await runtime.ingest_all(loader_for([doc]), scope=StoreScope(labels={"user_id": "alice"}))
stats = await runtime.ingest_all(loader_for([doc]), scope=StoreScope(labels={"user_id": "bob"}))

assert stats.documents_skipped == 1          # bob skipped — his scope has nothing
await runtime.retrieve("hello", scope=StoreScope(labels={"user_id": "bob"}))  # misses
```

### Additional Context

Likely fix: include `scope.to_payload_filters()` in `stale_filters` and in the pre-write `delete_where`. May also want document identity (uuid5) to incorporate scope so two tenants' copies of the same source are distinct documents.

AmirF194:
## Summary

Fixes #1183. Merges `scope.to_payload_filters()` into `_is_complete_duplicate`'s `stale_filters` and into `_embed_and_store`'s `delete_where`, the same way `VectorStore.read`/`nearest_neighbors` already scope reads in `stores/vector/base.py`.

## Type of change

- [x] Bug fix

## Checklist

## Notes

Two new tests in `test_runtime.py`, one per failure mode from the issue. Both fail on main (`test_duplicate_check_is_scoped_per_tenant` on the skip-as-duplicate case, `test_reingest_does_not_clobber_other_tenants_entries` on the clobber) and pass with the fix. Full `tests/unit_tests/retrieval/` suite (578 tests) green on the branch.

`delete_document(document_id)` has the same unscoped `delete_where` shape but its own signature takes no scope at all, so it's a separate public API surface, not the automatic ingest-path leak the issue reports. Left it alone; happy to fold it in if you'd rather it moved too.

### DIFF
--- a/packages/railtracks/src/railtracks/retrieval/runtime.py
+++ b/packages/railtracks/src/railtracks/retrieval/runtime.py
@@ -259,7 +259,7 @@ class RetrievalRuntime:
 
         await self._ensure_captured_model_seeded()
 
-        if await self._is_complete_duplicate(doc):
+        if await self._is_complete_duplicate(doc, scope):
             stats.documents_skipped += 1
             yield DocumentSkipped(document_id=doc.id, source=doc.source)
             return
@@ -297,7 +297,9 @@ class RetrievalRuntime:
         if doc_errors:
             yield self._record_document_failed(doc, doc_errors, stats)
 
-    async def _is_complete_duplicate(self, doc: Document) -> bool:
+    async def _is_complete_duplicate(
+        self, doc: Document, scope: StoreScope | None
+    ) -> bool:
         """Whether the store already holds a *complete* copy of ``doc``.
 
         Skip re-embedding only when as many chunks are present as the last
@@ -313,6 +315,7 @@ class RetrievalRuntime:
         stale_filters = {
             "source_path": doc.source,
             "content_hash": doc.content_hash,
+            **(scope.to_payload_filters() if scope is not None else {}),
         }
         existing = await self._store.find(stale_filters, limit=1)
         if not existing:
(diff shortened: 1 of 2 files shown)


Amir-R25 (approved):
Thanks @AmirF194, looks great

AmirF194:
Thanks, glad the scope fix landed clean.

PR #1559, opened 2026-09-10
docs: position railtracks as an agent harness framework

CoronRing:
## Summary

Positions railtracks as an agent harness framework. No ticket.

Adds an Agent Harness page mapping the five parts of a harness (loop, tool surface, context, controls, record) to the primitives that cover each, with a snippet source under `docs/scripts/` and two runnable examples: a read-only harness, and a coding harness whose file writes and shell calls are gated by `pre_verifier`.

Repositions the README, docs home, site description and package summary around "agent harness" while keeping "agent framework" for search, and adds PyPI keywords.

No API, module or doc-page renames.

Two drive-by fixes to existing docs found while writing the examples: `ToDoToolSet.pretty_dashboard()` and `get_incomplete_todos()` are async, and the todos page was showing them called without `await`.

The controls section documents `MaxCalls` in the `model_middleware=` slot and carries a warning pointing at #1560, where the same middleware in the `middleware=` slot never reaches its limit.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Breaking change
- [x] Docs
- [ ] Refactor / chore / build / tests

## Checklist

- [x] Lint & format pass (`ruff check . && ruff format .`)
- [ ] Tests added/updated and pass locally (`pytest tests`)
- [x] Docs updated if user-facing behavior changed
- [ ] Breaking changes include migration notes

## Notes

Known gap, left out of scope: `minimal_harness.py` joins model-supplied paths onto `REPO_ROOT` without confining them, so an absolute path or `..` escapes the repository. The tools are read-only. `coding_harness.py` has the confinement helper if we want to lift it across later.

PR #1420, opened 2026-08-18
docs: define and cross-link key keywords (Node, Agent Node, Sequential Flow, Direct Invocation)

rookepoole:
Resolves #1179.

This PR defines key Railtracks terms ('Node', 'Agent Node', 'Sequential Flow', 'Direct Invocation') in core documentation pages and adds cross-page hyperlinks so readers can easily navigate to full explanations of each concept.

---
Generated from a bounded immutable repository snapshot by PARS-Agent using Gemini 3.6 Flash. Tests listed above are recommendations unless GitHub checks report otherwise.

PR #1566, opened 2026-09-11
docs: define and cross-link Railtracks keywords (#1179)

### LINKED ISSUE #1179: [Docs] Define of keywords
### Issue Categorization

- [ ] Missing documentation
- [x] Incorrect or outdated information
- [ ] Overly verbose or complex explanation
- [ ] Broken links or formatting
- [ ] Missing code examples
- [ ] Typos or grammar issues

### Documentation Problem

We use some terms with a certain definition in mind, which others may not fully understand for example

"Node" "agent node", "sequential flow", "Direct invocation". Other than defineing them, it is a good practice to hyperlink the words to their definition page. 

### Additional Context

_No response_

CoronRing:
Resolves #1179. Written from scratch rather than continuing #1420, with that PR's review used as the specification.

## The problem

Framework vocabulary is used throughout the docs without being defined anywhere. `node` appears on 25 pages with no definition on any of them. "Direct Invocation" exists only as a nav label in `mkdocs.yml` and appears in no page. "entry point" and "run" carry real scoping rules (context lifetime, timeouts, per-run budgets) that a reader has no way to look up. Several concepts are also named two ways: a `Sequential Agent` card pointing at a `# Sequential Flows` page, an `## Agent Manifest` heading over `rt.ToolManifest` code, and a `Flows (Session Management)` title resting on a term the docs never define.

## Approach

**One canonical home per term.** The full description, with the reasons and an example, lives on the page that already owns the subject. The glossary is an index that points at those sections. No definition is written twice; a grep for each definition sentence returns exactly one file.

## What changed

**Definitions written where they were missing** (each with a stable anchor):

| Term | Home |
|---|---|
| Node, Agent Node | new `## Building Blocks` in `documentation/agent_design/overview.md` |
| Function Node | new lead section in `agent_design/tools/function_tools.md` (promoted out of a code annotation) |
| Flow, Entry Point, Run | `documentation/invocation/flows.md` |
| Direct Invocation | `documentation/invocation/call.md`, which previously had no H1 and never used its own nav name |
| Sequential Flow | `tutorials/concepts/architectures/sequential.md`, with a worked example |

**New glossary** at `docs/glossary.md`, a top-level nav entry before API Reference. 36 terms, grouped. Every entry is at most two sentences and ends in a link to the canonical section; no code, no examples. The page states that contract in a note so it stays an index rather than growing into a second set of docs.

**Definitions on hover.** Glossary entries become site-wide tooltips through `abbr`. The definition list is *generated* from `docs/glossary.md` by `scripts/mkdocs_hooks.py` (which already generates the API reference), so a tooltip and its glossary entry cannot drift apart. Two safeguards: the file is written only when its content actually changes, so `mkdocs serve`'s watcher cannot enter a rebuild loop; and a denylist keeps generic words (`Run`, `Tool`, `Context`, `Agent`, `Store`, `Document`, `LLM`, `Metric`, `Session`) out of the tooltips, since `abbr` rewrites every whole-word match on every page. Case sensitivity does useful work here too: `Flow` is tooltipped, the informal lowercase "flow" is left alone. A small CSS rule stops headings from being tooltipped, where the definition would only repeat the text directly beneath it.

**Keyword links at first mention** on the 18 pages that lean hardest on undefined vocabulary (quickstart, byfa, ryfa, the middleware pages, mcp, streaming, context, logging, error handling, the concept pages). First mention only, not every occurrence.

**Terminology reconciled:**
- `Sequential Agent` card renamed to `Sequential Flows` to match the page it links to.
- `## 2. Agent Manifest` renamed to `Tool Manifest` to match `rt.ToolManifest`, and given an actual definition.
- `(Session Management)` dropped from the Flows title.
- `tutorials/notebooks/architectures.md` carried prose byte-identical to two concept pages, so a glossary link had two equally valid targets. It now links to the canonical page instead of restating it.
- The informal-flow walkthrough gets one note distinguishing it from `rt.Flow`.
- Fixed two empty link targets (`[pass this function as a tool]()`, `[Tutorials]()`).
- Fixed a comment in `quickstart.py` calling an agent node a function node.

**From the #1420 review specifically:**
- Every definition link lands on an anchor, not a page top.
- `rt.call` is framed as composition *inside* a Flow, with a tip stating exactly what a top-level call gives up: Flow-scoped context, `timeout`, `end_on_error`, broadcast and payload callbacks, and the `flow.connect()` handle. Verified against `interaction/_call.py`, where a top-level call wraps itself in an implicit session with default config.
- The italicised *LLM*, *Tools* and *Middleware* on the agent-design page are now links.
- The Sequential Flow example lives in `docs/scripts/documentation/sequential.py` and is type-checked by `scripts/docs_validation.sh`, per the convention comments on that PR.
- `contributing_docs.md` gains a **Keywords and the glossary** section, so the convention is written down rather than relearned.

**Broken anchors now fail the build.** mkdocs reports unknown anchors at `info` level by default, which is how #1420 could ship links that landed on page tops. `validation.links.anchors: warn` plus `--strict` turns that into a build failure. Smoke-tested with a deliberately bad anchor: the build aborts. No pre-existing backlog surfaced.

**Terms from pages merged to main since this branch was cut.** Harness (#1559), System Message and Message History (#1542) each have a canonical page on main now, so each gets a glossary entry and a tooltip. System message is linked at its first mention on the agent design page and in the harness table, where it was the only unlinked primitive. The harness page now says a Flow starts a run rather than opens a session, matching the glossary's split between the execution (Run) and its record (Session).

## Verification

- `mkdocs build --strict` — clean, zero warnings, with anchor validation active.
- `scripts/docs_validation.sh` — passes, including the new `sequential.py`.
- `ruff check` / `ruff format --check` — clean on `scripts/mkdocs_hooks.py` (`docs/` is excluded from ruff in `pyproject.toml`).
- Confirmed `<abbr title=...>` renders, and that heading anchors survive the abbr pass.

## Deliberately out of scope

`Session` as an object and `Request` have no definition anywhere today, and the review of #1420 called `Session` a framework internal needing careful treatment of its own. So this PR defines the user-facing **Run** instead, and gives **Session** only its narrow verifiable meaning: the record written under `.railtracks/data/sessions/`.

Follow-ups worth their own tickets, all skipped here rather than given invented glosses:

- `Request` — appears in `request_id` and log output with nothing to point at.
- `LLMResponse` — `.text` / `.structured` are explained inside an admonition in `ryfa.md`, which produces no anchor to link.
- `AgentHub` — a nav group name that no page defines.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,9 @@ docs/api_reference/railtracks/
 docs/api_reference/index.html
 docs/api_reference/railtracks.html
 
+# Generated from docs/glossary.md by scripts/mkdocs_hooks.py
+docs/includes/glossary_tooltips.md
+
 # Documentation build output
 site/
 
--- a/docs/css/extra.css
+++ b/docs/css/extra.css
@@ -448,3 +448,14 @@
   color: var(--md-default-fg-color--light);
   margin-top: 1.5rem;
 }
+
+/* Glossary terms are headings as well as tooltipped keywords, and a tooltip on a
+   heading only repeats the text directly beneath it. Keep the definition available
+   without the dotted underline that invites a hover. */
+.md-typeset h1 abbr,
+.md-typeset h2 abbr,
+.md-typeset h3 abbr,
+.md-typeset h4 abbr {
+  text-decoration: none;
+  cursor: inherit;
+}
--- a/docs/documentation/advanced/context.md
+++ b/docs/documentation/advanced/context.md
@@ -1,10 +1,10 @@
 # Global Context
 
-Railtracks includes a concept of global context, letting you store and retrieve shared information across the lifecycle of a run. This makes it easy to coordinate data like config settings, environment flags, or shared resources.
+Railtracks includes a concept of global context, letting you store and retrieve shared information across the lifecycle of a [run](../invocation/flows.md#run). This makes it easy to coordinate data like config settings, environment flags, or shared resources.
 
 ## What is Global Context?
 
-The context system gives you a simple and clear API for interacting with shared values. It's scoped to the duration of a run, so everything is neatly contained within that execution lifecycle. One of the key features of the context system is that it can be accessed from within any node in your workflow, making it ideal for sharing data between different parts of your application.
+The context system gives you a simple and clear API for interacting with shared values. It's scoped to the duration of a run, so everything is neatly contained within that execution lifecycle. One of the key features of the context system is that it can be accessed from within any [node](../agent_design/overview.md#node) in your workflow, making it ideal for sharing data between different parts of your application.
 
 ## Core Functions
 
--- a/docs/documentation/agent_design/middleware/custom.md
+++ b/docs/documentation/agent_design/middleware/custom.md
@@ -1,6 +1,6 @@
 # Custom Middleware
 
-The middleware system is designed to support the custom creation of middleware to fit your needs. Each decorator below wraps a plain function into a `Middleware` object. Node-level decorators can be passed to `middleware=` on either `rt.agent_node` or `rt.function_node`. Model-level decorators can only be passed to `model_middleware=` on `rt.agent_node`. A `function_node` never calls a model, so it has no `model_middleware=` slot (the parameter doesn't exist on `function_node`, so passing one is a type/argument error, not a silent no-op).
+The middleware system is designed to support the custom creation of middleware to fit your needs. Each decorator below wraps a plain function into a `Middleware` object. Node-level decorators can be passed to `middleware=` on either an [agent node](../overview.md#agent-node) or a [function node](../tools/function_tools.md#what-a-function-node-is). Model-level decorators can only be passed to `model_middleware=` on `rt.agent_node`. A `function_node` never calls a model, so it has no `model_middleware=` slot (the parameter doesn't exist on `function_node`, so passing one is a type/argument error, not a silent no-op).
 
 | Decorator | Scope | Runs
 |---|---|---|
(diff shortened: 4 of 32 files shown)


Amir-R25 (dismissed):
Thanks Guan!

CoronRing:
### Polish & Tooltip Enhancements

Based on visual verification and second review, the following enhancements have been pushed:

1. **Anchor Jump Clearance**:
   - Added `scroll-margin-top: 9.5rem !important` in `docs/css/extra.css` across `:target`, `.md-typeset [id]`, and headings.
   - Fixes anchor fragment and Table of Contents navigation landing underneath the sticky header/tabs bar, ensuring titles remain fully visible.

2. **Hover Tooltip Styling & Behavior**:
   - **Eliminated Duplicate Tooltips**: Added an `on_page_content` hook in `scripts/mkdocs_hooks.py` that translates `<abbr title=...>` into `<abbr data-title=...>`. This prevents the browser from firing its native OS tooltip while allowing the custom styled card to render immediately.
   - **Wider Hover Card**: Increased popup card `max-width` by ~30% (`370px`) and improved line-height so definitions wrap naturally in 2–3 lines.
   - **Lowercase & In-Text Reference Matching**: Updated `_glossary_tooltips()` to generate lowercase alias entries for compound terms and key concepts (`function node`, `agent node`, `system message`, `middleware`, etc.), allowing natural in-text mentions and links to receive tooltips.
   - **Clean Heading and Link Visuals**: Suppressed dotted underlines on headings and on abbreviations inside `<a>` tags (`a abbr`), eliminating visual clashing with blue link text while keeping the definition popup active on hover.

3. **Config Cleanup**:
   - Removed `- content.tooltips` from `mkdocs.yml` (an unsupported Material Insiders flag).

PR #1568, opened 2026-09-13
Fix: null cost from failed pricing lookup renders as unknown, not $0.00

### LINKED ISSUE #1553: Failed cost lookups render as $0.00 instead of unknown
### Affected component

Observability / Visualizer

### Description

`extract_message_info` in `_litellm_wrapper.py` swallows any pricing-lookup
failure via `_return_none_on_error` and returns `total_cost=None`
(packages/railtracks/src/railtracks/llm/models/_litellm_wrapper.py:1054-1061).
Downstream, `row_mapping.py` coalesces that `None` to `0.0`:
```
    total_cost=float(row["total_cost"] or 0.0)
```
(packages/railtracks/src/railtracks/cli/viz_api/row_mapping.py:87,111)

So a cost we failed to compute (self-hosted model, brand-new model litellm
has no pricing entry for, unrecognized routing prefix, etc.) looks identical
to a call that actually cost $0.00.

### Impact

Anyone looking at session/viz cost totals can't tell "this was free" from
"we don't know" - cost visibility silently degrades to zero with no signal.

### Suggested fix

Carry the None/unknown state through to the UI/viz layer instead of
coalescing it to 0.0 - e.g. show "cost unknown" or omit the row from cost
totals rather than counting it as free.

### Reproduction

```python
from litellm.types.utils import ModelResponse, Usage
from railtracks.llm.models._litellm_wrapper import LiteLLMWrapper

response = ModelResponse(
    model="custom/some-unpriced-model",
    choices=[{"message": {"content": "hi"}, "finish_reason": "stop"}],
    usage=Usage(prompt_tokens=10, completion_tokens=5, total_tokens=15),
)

info = LiteLLMWrapper.extract_message_info(
    response, latency=0.0, requested_model="custom/some-unpriced-model"
)

print(info.total_cost)  # None: litellm has no pricing data for this model
# ...but row_mapping.py later does `float(row["total_cost"] or 0.0)`,
# turning this into an indistinguishable-from-free $0.00 in the UI.
```

### Error trace / logs

```shell

```

### System info

```shell

```

Christian-Sidak:
## Summary

- When a model has no litellm pricing entry (self-hosted, brand-new, or unrecognized routing prefix), `extract_message_info` returns `total_cost=None`. Three layers were silently converting that `None` to `0.0`, making an unresolvable cost look identical to a genuinely free call.
- Fix all three layers: the SQL queries (remove `COALESCE(total_cost, 0.0)` from per-row and per-node selects), the Pydantic models (widen `total_cost` from `float` to `float | None`), and the Python mapping/route code (replace `float(x or 0.0)` with `float(x) if x is not None else None`).
- Session and node rollups that have no LLM calls at all still report `0.0` (they are genuinely free). Aggregate stat tiles (SessionStats, LLMTraceStats) keep their `COALESCE` because they sum known costs across many sessions, which is still useful.

## Test plan

- [x] Added `test_null_cost_surfaces_as_null_in_session_list`: writes a session whose single LLM response has `total_cost=null` and asserts the sessions API returns `total_cost: null`.
- [x] Added `test_zero_cost_surfaces_as_zero_in_session_list`: writes a session with `total_cost=0.0` and asserts the API still returns `0.0` (genuinely free calls are unaffected).
- [x] Added `test_null_cost_surfaces_as_null_in_llm_traces`: same null-cost event, verified through the `/api/llm-traces` endpoint.

Fixes #1553

### DIFF
--- a/packages/railtracks/src/railtracks/cli/viz_api/queries/llm_traces.py
+++ b/packages/railtracks/src/railtracks/cli/viz_api/queries/llm_traces.py
@@ -272,7 +272,7 @@ def list_llm_trace_rows(
       r.exception_message                            AS error_message,
       COALESCE(r.input_tokens, 0)                    AS input_tokens,
       COALESCE(r.output_tokens, 0)                   AS output_tokens,
-      COALESCE(r.total_cost, 0.0)                    AS total_cost,
+      r.total_cost                                    AS total_cost,
       r.latency                                      AS latency_seconds,
       CAST(r.message_input AS VARCHAR)               AS message_input_json,
       CAST(r.output AS VARCHAR)                      AS output_json
--- a/packages/railtracks/tests/unit_tests/cli/viz_api/test_api_queries.py
+++ b/packages/railtracks/tests/unit_tests/cli/viz_api/test_api_queries.py
@@ -444,3 +444,131 @@ def test_query_failures_emit_structured_error_logs(
     assert payload["path"] == "/api/v2/sessions"
     assert payload["error_type"] == "RuntimeError"
     assert payload["error"] == "duckdb exploded"
+
+
+def _llm_session_events(
+    session_id: str,
+    node_id: str,
+    llm_id: str,
+    total_cost: float | None,
+) -> list[dict[str, object]]:
+    """Minimal event stream for a session with one LLM response.
+
+    ``total_cost`` is passed through unchanged so callers can write either a
+    known value or ``None`` (unpriced model) and verify how the API renders it.
+    """
+    stamp = "2026-01-01T00:00:00+00:00"
+    return [
+        _event(
+            f"started-{session_id}",
+            "session.started",
+            session_id,
+            {"session_id": session_id, "flow_name": f"flow-{session_id}"},
+            stamp=stamp,
+        ),
+        _event(
+            f"node-{session_id}",
+            "node.creation",
+            session_id,
+            {"node_id": node_id, "name": "agent", "node_type": "Agent"},
+            stamp=stamp,
+        ),
+        _event(
+            f"llm-creation-{session_id}",
+            "llm.creation",
+            session_id,
+            {"llm_id": llm_id, "model_provider": "openai", "model_name": "gpt-4o"},
+            stamp=stamp,
+        ),
+        _event(
+            f"llm-response-{session_id}",
+            "llm.response",
+            session_id,
+            {
+                "spatial_parent_node_id": node_id,
+                "parent_llm_type_id": llm_id,
+                "input_tokens": 10,
+                "output_tokens": 5,
+                "total_cost": total_cost,
+                "message_input": [{"role": "user", "content": "hi"}],
+                "output": {"role": "assistant", "content": "hello"},
+            },
+            stamp=stamp,
+        ),
+        _event(
+            f"completed-{session_id}",
+            "session.completed",
+            session_id,
+            {"status": "success", "duration_seconds": 1.0},
+            stamp=stamp,
+        ),
+    ]
+
+
+def test_null_cost_surfaces_as_null_in_session_list(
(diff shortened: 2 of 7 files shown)


packages/railtracks/tests/unit_tests/cli/viz_api/test_api_queries.py
```diff
@@ -444,3 +444,129 @@ def fail(*_args: object, **_kwargs: object) -> list[dict[str, object]]:
     assert payload["path"] == "/api/v2/sessions"
     assert payload["error_type"] == "RuntimeError"
     assert payload["error"] == "duckdb exploded"
+
+
```
Aryan-Railtown, line 448: also could you add a test for a failed (llm.failure) call in list_llm_trace_rows to lock this in, alongside the existing null/zero-cost cases?

packages/railtracks/src/railtracks/cli/viz_api/queries/llm_traces.py
```diff
@@ -272,7 +272,7 @@ def list_llm_trace_rows(
       r.exception_message                            AS error_message,
       COALESCE(r.input_tokens, 0)                    AS input_tokens,
       COALESCE(r.output_tokens, 0)                   AS output_tokens,
-      COALESCE(r.total_cost, 0.0)                    AS total_cost,
+      r.total_cost                                    AS total_cost,
```
Aryan-Railtown, line 279: Right now the blanket removal conflates the two, a failed call now shows null instead of 0.0, which re-creates the same null/zero confusion this PR is meant to fix, just in the other direction.

Aryan-Railtown (changes requested):
Nice find on the null/zero cost distinction...quick, targeted fix overall. One thing I think we need to tighten up before merging though:

Removing C`OALESCE(r.total_cost, 0.0`) fixes the case we care about, but it also changes behavior for a case we don't want to touch. There are two distinct sources of total_cost = NULL here:

1. Successful call, pricing lookup failed (what #1553 is actually about) -> cost is genuinely unknown, and should render as null/"unknown". This is the fix we want.
2. Call failed outright (llm.failure) -> never got a response, so nothing was billed. Cost is genuinely $0.00, not unknown. The _llm_calls_cte docstring already says this explicitly: "tokens and cost COALESCE to zero... nothing is billed for a call that never came back."


Suggested fix: branch on status (already computed on the CTE from event_type = 'llm.failure') instead of dropping COALESCE entirely:

CASE WHEN r.status = {_STATUS_ERROR_LITERAL} THEN 0.0 ELSE r.total_cost END AS total_cost,

That keeps null meaningful only for "pricing unresolvable on a call that did return," and restores 0.0 for calls that never came back.

Christian-Sidak:
Good catch. Addressed in 47d1d86.

Used `CASE WHEN event_type = 'llm.failure' THEN 0.0 ELSE total_cost END` inside `_llm_calls_cte`, which is the equivalent of your `r.status` branch but using the raw column since `status` is derived in the same SELECT. Added a test that feeds a `llm.failure` event through the traces endpoint and asserts `total_cost = 0.0`.

Aryan-Railtown (dismissed):
Looks good!! Thank you for the amazing work!!

Aryan-Railtown:
hi @Christian-Sidak could you look into why the unit tests for this are failing. That is the only blocker for now, the work looks good

Christian-Sidak:
Fixed in 79ad4b5: the two new tests were calling `/api/llm-traces` but the llm-traces router is mounted under the `/api/v2` prefix. The catch-all at the bottom of `viz_server.py` intercepts any unmatched `/api/*` path and returns 404, which is why both were failing at the status-code assertion before reaching the cost check. Changed both test calls to `/api/v2/llm-traces`.

Aryan-Railtown (approved):
LGTM

Issue #1425, opened 2026-08-18
Emit verifier decisions as observability events


Aryan-Railtown:
### Description

The general verifier middleware (#1266, shipped in #1401) currently only logs accepted-with-comment decisions via the Python logger. @soulFood5632 flagged on #1401 that verifier decisions (accept, decline, decline-with-comments, timeout) should be sent as specific observability events instead, so they show up in traces rather than just logs -- something like how guardrails emit structured trace records today.

Open how this is best modeled (a dedicated event type, extending existing middleware/guardrail event plumbing, or something else) -- left to whoever picks this up.

soulFood5632:
For implementation please use the the MiddlewareGuardrail events as a guideline. Make sure they are tagged in a similar way.

PR #1569, opened 2026-09-15
feat: verifier event emission + viz classification (#1425)

Aryan-Railtown:
## Summary

Closes the full scope of #1425 (Event Emission). Two parts, both included here:

1. **Event emission** (already committed, 4 commits): `pre_verifier`/`post_verifier` now emit
   `MiddlewareVerifierPre*`/`Post*` events carrying a structured `VerifierDecision`, instead of
   only logging.
2. **Viz classification follow-up**: the viz API now understands those events.
   - `verifier` middleware classifies as `MiddlewareKind.VERIFIER` instead of falling through to
     the generic `node_wrapper` bucket.
   - The verifier decision vocabulary (`accept` / `accept-with-override` / `decline`) normalizes
     onto the guard's `allow` / `transform` / `block`, so every existing aggregate (allows/
     transforms/blocks counts, block conditions, outcome rollups) is verifier-aware for free.
   - A verifier decline now produces session status `Blocked` instead of generic `Failed` — the
     same treatment guardrail blocks already get.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

- [ ] Docs updated if user-facing behavior changed — not needed: no docs/changelog reference
      `MiddlewareKind`'s enum values anywhere in the repo (confirmed by grep), and the public
      `approve_fn`/`Verdict` API is unchanged.
- [ ] Breaking changes include migration notes — n/a, not a breaking change.

## Explicitly out of scope (deferred on purpose)

No per-node "shield marker" for verifiers (the analogue of guardrails'
`list_guardrails_by_node` / `GuardrailsPopover`). No generic equivalent exists yet on either the
guard or verifier side, and the recommendation — recorded in the "Event Emission (#1425)" Notion
page — is to hold off until there's evidence node-level location is actually needed, since the
Middleware column's pill already surfaces outcome + reason at session grain.

## Files touched (viz follow-up)

- `cli/viz_api/models.py` — adds `MiddlewareKind.VERIFIER`.
- `cli/viz_api/queries/middleware.py` — adds the verifier rung to the kind ladder, and a `CASE`
  mapping the verifier action vocabulary onto guard vocabulary.
- `cli/viz_api/queries/sessions.py` — adds `VerifierRejectedError` to the `Blocked` status `CASE`,
  plus a `middleware.failure` branch in `last_failure`.
- `tests/unit_tests/cli/viz_api/test_api_queries.py` — new `_verifier_decision_events` helper +
  3 tests.

### How this was verified

Independent adversarial re-check, then a fix pass. Status: **no blocking issues outstanding.**

- Traced the full propagation path by hand: a `pre_verifier` decline raises
  `VerifierRejectedError` *before* the wrapped node ever runs, so `node.failure`/`llm.failure`
  never fire for it.
- Ran real scenarios end-to-end covering pre/post accept, accept-with-override, decline,
  decline-via-timeout, and an unexpected exception inside `approve_fn` (not a decline).
- Queried the real event store directly and confirmed: accept → `allow`, accept-with-override →
  `transform`, decline (incl. timeout) → `block` + session status `Blocked`; an `approve_fn`
  crash registers as blocked at the middleware grain but leaves session status `Failed`.
- `ruff check` / `ruff format --check` clean. Full suite: 3016 passed, 5 skipped, 4 pre-existing
  unrelated failures (Windows symlink + MCP integration), nothing new broke.

**Gap found and fixed:** zero test coverage for the kind ladder / vocabulary normalization /
Blocked status CASE existed anywhere (not even for guardrails). Added 3 new tests; suite re-run
clean after.

Rendering counterpart PR is in `RailtownAI/railtracks-visualizer` (`proto/verifier-event-rendering`
→ `v2`), opened separately.


packages/railtracks/src/railtracks/events/middleware.py
```diff
@@ -236,3 +237,54 @@ def event_type(self) -> str:
 class MiddlewareModelFailureEvent(MiddlewareModelEventBase, FailureMixin):
     def event_type(self) -> str:
         return "middleware.model.failure"
+
+
+# Verifiers (pre_verifier / post_verifier) — node-band, mirroring the guard
```
Amir-R25: let's remove these inline comments

Amir-R25 (dismissed):

Amir-R25 (approved):

PR #1571, opened 2026-09-15
feat(prebuilt): add ConversationMemory middleware and standardize prebuilt docs

### LINKED ISSUE #1570: Add prebuilt middleware to cache message history so it does not need to be passed back in a loop
### Problem Statement & Motivation

When developers build test harnesses, CLI tools, or interactive chat workflows using Railtracks (`rt.agent_node` and `rt.Flow`), each invocation is completely isolated and stateless by default.

For example, running:
```python
flow = rt.Flow("agent_flow", entry_point=MyAgent)
flow.invoke("What is my name?")
flow.invoke("What did I just ask?")
```
results in the second invocation having no awareness of the prior turn.

To maintain multi-turn history today, developers must manually:
1. Extract `response.message_history`
2. Construct and append `rt.llm.UserMessage`
3. Thread and pass the `MessageHistory` object back into subsequent invocations in a loop.

This adds unnecessary boilerplate for common multi-turn interaction patterns.

### Proposed Solution

Provide a drop-in prebuilt middleware (`ConversationMemory` in `railtracks.prebuilt.middleware`) that caches message history in the session context (`rt.context`) so it does not need to be manually passed back in a loop:
- Preserves conversation history across successive invocations.
- Automatically caches message history into session context variables (`rt.context`), namespaced by agent name (e.g. `conversation_history_<agent_name>`) so multiple agents in one flow do not collide.
- Automatically prepends existing history to incoming user inputs.
- Supports custom `context_key` for explicit session context naming or intentional memory sharing between agents.
- Supports initial history seeding via preloaded session/flow context.
- Supports sliding window truncation via `max_messages`.

CoronRing:
## Description

Closes #1570

This PR adds `ConversationMemory`, a prebuilt node-level middleware (`railtracks.prebuilt.middleware.ConversationMemory`) that automatically caches, accumulates, and prepends multi-turn conversation history across repeated node invocations into the active session context (`rt.context`) so that message history does not need to be manually passed back in a loop.

Additionally, this PR standardizes prebuilt middleware documentation pages (`docs/documentation/agent_design/middleware/prebuilt/list/`) so that document titles and prose match their exact class names (e.g. `MaxCalls`, `ContextInjection`, `ConversationMemory`).

### Motivation

Prior to this change, `agent_node` invocations were strictly stateless: each invocation was isolated. To maintain conversation history across multiple turns in harnesses, CLI tools, or multi-step flows, developers had to manually extract `response.message_history`, construct `UserMessage` objects, and pass `MessageHistory` back into subsequent invocations in a loop.

### Architecture & Design

- **Zero Framework Bloat (Fully Self-Contained)**:
  `ConversationMemory` requires **no changes to core framework files** (`node.py`, `nodes.py`, etc.). It behaves like any standard `Middleware`.
- **Automatic Per-Instance Isolation**:
  Each `ConversationMemory()` instance automatically generates a unique, isolated session context key (e.g. `conversation_history_<instance_id>`), ensuring that multiple agents running in the same flow each have completely independent conversation memory stores by default.
- **Intentional Sharing & Custom Keys (`context_key`)**:
  If developers want multiple agents to share a single conversational thread, they can pass the same explicit `context_key="team_chat"` to each memory instance or supply a custom key for easy session querying.
- **Context Inspection After Flow Completion (`conn.context`)**:
  `flow.invoke()` / `flow.ainvoke()` return only the flow result. To inspect context after an invocation finishes, users call `flow.connect()`, yielding a `FlowConnection` whose `.context.get(memory.context_key)` remains accessible after the run completes. Direct instance inspection is also supported via `memory.get_history()`.
- **Preloading & Seeding**:
  Supports seeding initial conversation context via Flow context (e.g. `rt.Flow(..., context={"team_chat": initial_history})` or `flow.ainvoke(..., context=...)`).
- **Node-Level Middleware (`middleware=`)**:
  Operates at the node boundary (`UserInput -> LLMResponse`) to accumulate turns across complete agent invocations.
- **Copy Protection (`__deepcopy__`)**:
  `Node.safe_copy()` copies node internals on each invocation. `ConversationMemory` implements `__deepcopy__` to preserve state references across node copies.
- **Sliding Window Pruning**:
  Optional `max_messages` parameter caps history to the most recent $N$ messages to avoid context window exhaustion.
- **Reset Support**:
  `memory.clear()` resets both the instance history and deletes the variable from the active session context.

---

### Core Package
- `packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py`: `ConversationMemory` middleware with per-instance unique context keys, optional explicit keys, and session context caching (`rt.context`).
- `packages/railtracks/src/railtracks/prebuilt/middleware/__init__.py`: Clean flat export of `ConversationMemory`.

### Documentation
- `docs/documentation/agent_design/middleware/prebuilt/list/conversation_memory.md`: Reference documentation for `ConversationMemory` detailing multi-agent isolation, per-instance and custom keys, and session context inspection.
- `docs/documentation/agent_design/middleware/prebuilt/list/max_calls.md`: Standardized title and text to `MaxCalls`.
- `docs/documentation/agent_design/middleware/prebuilt/list/context_injection.md`: Standardized title to `ContextInjection`.
- `docs/documentation/agent_design/middleware/prebuilt/overview.md`: Catalog table updated to include `ConversationMemory`.
- `docs/scripts/prebuilt_middleware.py`: Added runnable snippet for MkDocs documentation.

### Tests
- `packages/railtracks/tests/unit_tests/middlewares/test_prebuilt_conversation_memory.py`: 14 comprehensive unit tests covering:
  - Base middleware inheritance and flat export verification.
  - Multi-turn conversation accumulation.
  - Multi-agent isolation in the same flow (verifying separate memory stores per agent without framework coupling).
  - Explicit `context_key` memory sharing between agents.
  - Preloaded conversation history via session context.
  - Custom context key isolation.
  - History reset (`clear()`).
  - Sliding window capping (`max_messages`).
  - Keyword argument support (`user_input=...`).
  - Flow connection context inspection (`conn.context.get(...)`).
  - End-to-end sync (`flow.invoke`) and async (`flow.ainvoke`) agent workflows.

---

## Verification

- **Unit Tests**:
  - `test_prebuilt_conversation_memory.py`: 14 passed in 0.56s.
  - All middleware unit tests: 116 passed in 4.45s.
- **Linting & Formatting**:
  - `ruff check`: 0 errors.
  - `ruff format --check`: all files formatted cleanly.
- **Dependency Sorting**:
  - `scripts/check_dependencies_sorted.py`: Passed.
- **Documentation**:
  - `mkdocs build --strict`: Built successfully with zero warnings/errors in 30.95s.


docs/documentation/agent_design/middleware/prebuilt/list/conversation_memory.md
```diff
@@ -0,0 +1,29 @@
...
+
+## How It Works
+
+1. On the first turn, the user message is sent to the model normally, and the resulting message history is saved in the active session context (`rt.context`) under `context_key` (defaulting to `"conversation_history"`).
+2. On subsequent turns, the prior message history is retrieved from the session context, the incoming user input is appended, and the accumulated history is passed to the agent.
+3. The session context variable is automatically updated after every turn, accumulating multi-turn conversation context.
+
+## Session Context & Configuration
+
+- **Session Context Injection**: History is injected directly into `rt.context` under `context_key="conversation_history"` by default. You can inspect or access it via `rt.context.get("conversation_history")`, `session.context["conversation_history"]`, or `connection.context["conversation_history"]`.
```
CoronRing: Note I have no better solution, I thought of adding a system held "system" which is protected, but need a bit more discussion and beyond ticket.

docs/documentation/agent_design/middleware/prebuilt/list/conversation_memory.md
```diff
@@ -0,0 +1,43 @@
...
+- **Context Inspection After Flow Completion**: `flow.invoke()` and `flow.ainvoke()` return only the flow's final result. To inspect context after an invocation finishes, use `flow.connect()`, which returns a `FlowConnection`:
+  ```python
+  conn = flow.connect()
+  result = await conn.ainvoke("Follow up question")
+
+  # Inspect conversation history from the completed run's context:
+  history = conn.context.get(memory.context_key)
+  ```
+- **Avoid Passing History Manually**: Do not pass prior `MessageHistory` as user input when `ConversationMemory` is attached, as the middleware automatically accumulates and prepends history across turns.
+- **Max Messages**: Pass `max_messages=10` to prune history to the most recent $N$ messages and avoid exceeding model context windows.
```
Pooria90: small comment: we don't currently have LaTeX style formula across the docs. Let's switch to italic instead of $$.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
```diff
@@ -0,0 +1,154 @@
...
+        if is_context_present():
+            try:
+                context.delete(self._context_key)
+            except KeyError:
+                pass
+            if not self._explicit_key:
+                try:
+                    context.delete("conversation_history")
+                except KeyError:
+                    pass
```
Aryan-Railtown: clear() on one instance deletes the shared "conversation_history" seed key, any other default-keyed instance loses its preloaded history the moment any instance calls clear(). Repro attached:

```
async def bug_clear_wipes_other_instance_seed():
    section("BUG 1 (line 54): clear() deletes shared 'conversation_history' seed key")
    model = FakeEchoModel()
    preloaded = MessageHistory([])
    from railtracks.llm.message import UserMessage, AssistantMessage

    preloaded.extend([UserMessage("seeded question"), AssistantMessage("seeded answer")])

    memory_researcher = ConversationMemory()  # default-constructed, no explicit key
    memory_writer = ConversationMemory()  # default-constructed, no explicit key

    with rt.Session(context={"conversation_history": preloaded}) as s:
        print(f"Researcher sees seed BEFORE Writer.clear(): {memory_researcher.get_history() is not None}")
        memory_writer.clear()  # writer never used the seed itself
        after = memory_researcher.get_history()
        print(f"Researcher sees seed AFTER Writer.clear(): {after is not None}")
        print("EXPECTED: True (Researcher's own data untouched by Writer.clear())")
        print(f"ACTUAL:   {after is not None}  <-- silently wiped by an unrelated instance's clear()")
```

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
```diff
@@ -0,0 +1,154 @@
...
+        if isinstance(user_input, str):
+            history.append(UserMessage(user_input))
+        elif isinstance(user_input, UserMessage):
+            history.append(user_input)
+        elif isinstance(user_input, (list, MessageHistory)):
+            if len(user_input) >= len(history) and list(
+                user_input[: len(history)]
+            ) == list(history):
+                del history[:]
+                history.extend(user_input)
```
Aryan-Railtown: The dedup check for re-passed history uses == on Message objects, but Message has no `__eq__`, so it's really identity comparison against a deep-copied list and never matches, it ends up duplicating the whole conversation instead of deduping. Minor since it only bites if someone re-feeds get_history()'s own output back in, but since we already guard for exactly that case, it should actually work.

```
async def bug_dedup_identity_comparison_duplicates_history():
    section("BUG 2 (line 88-97): re-feeding get_history() output duplicates history instead of deduping")
    model = FakeEchoModel()
    memory = ConversationMemory()
    agent = rt.agent_node("DedupAgent", llm=model, middleware=[memory])
    flow = rt.Flow("dedup-flow", entry_point=agent)

    await flow.ainvoke("first turn")
    history_so_far = memory.get_history()
    print(f"history length after turn 1: {len(history_so_far)}")

    # A caller re-passes the full history it just got back (a documented inspection API)
    await flow.ainvoke(history_so_far)
    print(f"history length after re-feeding get_history() output: {len(memory.get_history())}")
    print("EXPECTED: dedup detects the prefix match and does not double the history")
    print("ACTUAL:   duplicated, because Message has no __eq__ so list equality is identity-based"
          " and never matches the deep-copied prefix")
```

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
```diff
@@ -0,0 +1,154 @@
...
+        super().__init__(self._middleware_fn)
+
+    @property
+    def context_key(self) -> str:
+        """The session context key holding the conversation history."""
+        return self._context_key
+
+    def get_history(self) -> MessageHistory | None:
+        """Return the current conversation history from session context or instance."""
+        return self._get_existing_history()
```
Aryan-Railtown, line 79: this function return the live MessageHistory object by reference, not a copy, so a caller mutating this object in place corrupts the middleware's internal state directly. Should be a one-line deepcopy on the way out.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
```diff
@@ -0,0 +1,154 @@
...
+        else:
+            history.append(UserMessage(str(user_input)))
+
+    def _combine_history(
+        self, existing: MessageHistory, user_input: object
+    ) -> MessageHistory:
+        combined = deepcopy(existing)
+        self._append_to_history(combined, user_input)
+        if self._max_messages is not None and len(combined) > self._max_messages:
+            return MessageHistory(combined[-self._max_messages :])
```
Aryan-Railtown, line 145: `max_messages=0` is silently treated as "no limit". Python's list[-0:] is the same as list[0:], so the slice is a no-op. Needs an explicit check (e.g. if self._max_messages is not None with a separate == 0 case) at both pruning sites

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
```diff
@@ -0,0 +1,154 @@
...
+
+        if has_input and existing:
+            combined = self._combine_history(existing, user_input)
+            if is_pos:
+                args = (combined, *args[1:])
+            else:
+                kwargs = {**kwargs, "user_input": combined}
+
+        result = await call(*args, **kwargs)
+        self._save_result(result)
```
Aryan-Railtown, line 180: No lock around the read → call model → save sequence in _middleware_fn, so concurrent invocations on the same instance asyncio.gather on the same agent) race and one turn's write can silently clobber the other's.
This might interfere with the subagent type usage we talked about. Repr:

```
async def bug_concurrent_calls_race():
    section("BUG 5 (line 141-153): no lock around read-modify-write; concurrent calls race")
    model = FakeEchoModel()
    memory = ConversationMemory()
    agent = rt.agent_node("RaceAgent", llm=model, middleware=[memory])
    flow = rt.Flow("race-flow", entry_point=agent)

    await asyncio.gather(flow.ainvoke("message A"), flow.ainvoke("message B"))
    final = memory.get_history()
    print(f"final history length after two concurrent turns: {len(final)}")
    print("EXPECTED: 4 messages (2 turns x user+assistant), both turns present")
    print(f"ACTUAL:   {len(final)} -- one turn's write can silently clobber the other's")
```

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def __init__(
+        self,
+        context_key: str | None = None,
+        *,
+        max_messages: int | None = None,
+    ):
+        self._instance_id = uuid.uuid4().hex[:8]
+        self._explicit_key = context_key is not None
+        self._context_key = context_key or f"conversation_history_{self._instance_id}"
```
Aryan-Railtown: context_key="" sets _explicit_key=True, but _context_key still falls back to the auto-generated key (line 39, or), an empty string silently ends up in an inconsistent explicit/auto state.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+class ConversationMemory(Middleware):
+    """Automatically cache and append conversation history across node invocations.
+
+    Node-level middleware (``middleware=``). Automatically preserves multi-turn
+    conversations in session context and prepends prior turns to incoming inputs.
+    Do not manually pass prior message history into invocations when this middleware
+    is attached, as history is accumulated automatically.
+
```
Aryan-Railtown, line 41: This is documented as node-level (middleware=) only, but nothing stops attaching it via` model_middleware=`. There it silently never persists anything instead of erroring, since the result at that boundary isn't an LLMResponse. We should raise a clear error on misuse rather than fail silently

Aryan-Railtown (changes requested):
Amazing work!! Love the idea. Claude came up with a few bugs, flagging the ones that i think are worthwhile :D

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
```diff
@@ -0,0 +1,154 @@
...
+        return self._context_key
+
+    def get_history(self) -> MessageHistory | None:
+        """Return the current conversation history from session context or instance."""
+        return self._get_existing_history()
+
+    def clear(self) -> None:
+        """Clear conversation history from both session context and instance."""
+        self._state["history"] = None
+        if is_context_present():
```
Pooria90, line 90: I can't think of a case where this block actually opens. Basically, this method only `None`s the history. In the case of `FlowConnection`, you can still access the history using `conn.context.get(memory.context_key)` even after you have called `memory.clear()`.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
```diff
@@ -0,0 +1,154 @@
...
+            try:
+                hist = context.get(self._context_key)
+                if hist is not None:
+                    return hist
+            except KeyError:
+                pass
+            if not self._explicit_key:
+                try:
+                    hist = context.get("conversation_history")
```
Pooria90: Which part of the code adds this key to the context?

Pooria90 (changes requested):
Great Work @CoronRing 
I have minor change requests and general questions.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
CoronRing: __init__ now raises ValueError on a blank or whitespace-only context_key, so the inconsistent explicit/auto state is unreachable rather than being papered over downstream. Test added.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
CoronRing: User input <1 message is very rare, but I made a fix anyway s. __init__ now normalizes with max_messages or None, so 0 is documented as "no limit" at one site instead of depending on the list[-0:] accident at two. Negative values now raise, which was the worse half of this: max_messages=-5 was slicing from the front and silently keeping the wrong end of the history. Tests added for both.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
CoronRing: Message has no __eq__ as you said, and the check was also comparing against combined, which is already a deepcopy of the stored history, so it could never have matched even if Message did define __eq__. Fix is a module-level _continues() helper comparing role and content, short-circuiting on is for the common case, and the check moved into _combine_history so it runs against the original history before the copy. I kept __eq__ off Message deliberately, that is a much wider blast radius than this PR should carry.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
CoronRing: The cross-talk only exists between default-keyed instances that have not yet taken a turn, because after the first save each instance reads and writes its own key. So the window is "two default-constructed instances, a shared seed in context, and a clear() before either has run". In practice one memory instance holds one conversation, clear() almost always means start from scratch, and callers reaching for a fresh conversation tend to construct a new instance rather than clear an old one.

The clean fix if we do want it later is adopt-once rather than an argument on clear(): the first time an instance falls back to the shared conversation_history key, copy it into that instance's own key and never read the shared one again. We can discuss this more if you feel strongly about .clear()

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
CoronRing: A lock inside the middleware serializes every invocation of the agent it is attached to, including cases where the caller deliberately fanned out. That is a silent throughput cost paid by everyone to protect against a case that is arguably already undefined: one ConversationMemory instance models one sequential conversation, and two concurrent turns reading the same prior history do not have a well-defined correct interleaving even with a lock.

The better composition already exists. Once https://github.com/RailtownAI/railtracks/pull/1561 lands and Lock works in the node slot, middleware=[Lock(), ConversationMemory()] gives exactly this, opt-in, with the ordering already correct since the first entry is outermost. For now I have documented the constraint explicitly on the docs page rather than leaving it implicit.

Happy to switch to a built-in lock if you feel strongly, but I would want it after https://github.com/RailtownAI/railtracks/pull/1561 so it composes with Lock rather than duplicating it.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
CoronRing: Fair point, but I do not think there is a clean guard to write. ModelMiddleware is a bare type alias of Middleware, so there is no runtime type to check at construction time and nothing to reject in __init__.

The only available signal is "the result was not an LLMResponse", which fires identically for legitimate-looking node attachment to a non-agent node, so a hard raise there would reject valid-ish usage on a path we cannot distinguish. I would rather leave the docstring's node-level statement as the contract than raise on a signal that does not actually mean what we want it to mean. If this bites someone in practice, a one-shot warning on the first non-LLMResponse result is the honest version and I would add it then.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
CoronRing: The reference is reachable several other ways regardless, conn.context.get(memory.context_key) being the documented one, so a deepcopy here closes one door in a room with several. It also is not free: the existing test at line 53 asserts memory.get_history() == session_hist, which passes only because it is identity comparison, so copying on the way out makes that assertion false and pulls Message.__eq__ back into scope. Not worth it for the protection actually gained.

docs/documentation/agent_design/middleware/prebuilt/list/conversation_memory.md
CoronRing: Good catch, fixed

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
CoronRing: The block does open. Inside a run is_context_present() is True and the delete lands, the key is gone from s.context.keys() afterwards. test_clear_method covers that path today.

Your conn.context point is correct though. After the run exits, is_context_present() is False, so clear() only drops the instance copy and conn.context.get(memory.context_key) still returns the pre-clear history. So clear() is silently scope-dependent, which is the actual defect here.

I do not think it is fixable from inside the middleware. conn.context returns the live session's context object, whereas the middleware only reaches context through the runner_context contextvar, which is unset once the run ends, so there is nothing to delete from at that point short of handing the middleware a session handle. I would rather not add that coupling for this. Instead the clear() docstring and the docs bullet now say plainly that the instance copy is always dropped but the session entry only goes when called during a run. Happy to reconsider if you think the coupling is worth it.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
CoronRing: Its not really anywhere. That bare conversation_history key is read-only: it is a seeding convention the caller is meant to set via rt.Flow(context={...}), so a default-constructed instance can be preloaded without knowing its generated key. The middleware always writes to self._context_key, never to this one.

It is also documented nowhere and only exercised by test_preloaded_session_context, and it is the direct cause of the clear() cross-instance bug Aryan flagged, since every default-keyed instance reads the same key.

I would like to just delete it. Seeding then goes through an explicit context_key, which is what the docs already recommend, and the fallback plus the _explicit_key flag it exists to gate come out together, about 13 lines lighter. That also resolves Aryan's finding by removing the shared state rather than by adding logic to manage it. Let me know if you see a use case for the implicit key that I am missing, otherwise I will take it out.

Pooria90: Deletion sounds good!

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
Pooria90: Sounds good.

packages/railtracks/src/railtracks/prebuilt/middleware/conversation_memory.py
Aryan-Railtown: fair enough

Aryan-Railtown (dismissed):
LGTM

CoronRing:
@Aryan-Railtown You find it good, you mind approving it? I will have both Aryan and Pooria signup since both of you helped review

Pooria90 (approved):
LGTM

Aryan-Railtown:
> @Aryan-Railtown You find it good, you mind approving it? I will have both Aryan and Pooria signup since both of you helped review

Already did yesterday

Issue #1574, opened 2026-09-18
Session flow_name warning fires on railtracks' own internal call paths


CoronRing:
### Affected component

Agents / Nodes

### Description

`Session.__init__` warns whenever it is constructed without a `flow_name`:

https://github.com/RailtownAI/railtracks/blob/main/packages/railtracks/src/railtracks/_session.py#L90-L94

The intent is to nudge users who write `with rt.Session():` toward `rt.Flow`. But railtracks constructs exactly that kind of flow-less session on its own internal paths, so the warning fires at users who never mentioned `Session`:

- `interaction/_call.py` opens `with Session():` when `rt.call` runs with no context present
- `interaction/_astream.py` does the same in its `_open` closure

The result is that a plain top-level `await rt.call(...)` warns. That call is a documented, recommended entry point ([Call](https://docs.railtracks.org/documentation/invocation/call/): "if you do not require any configurations, you can simply use the `railtracks.call` API"), so the framework is warning about a pattern its own docs endorse.

Three concrete problems:

1. The warning is attributed to `_session.py:91`, not to the user's line, so it is hard to locate and hard to suppress from user code.
2. It names an API (`Session`) the user never touched, so the suggested fix does not match anything in their code.
3. It says "This warning will become an error in future versions." If that happens as written, top-level `rt.call` breaks, because the internal path constructs the very session that would raise.

This also contradicts the project's own guidance in `utils/deprecation.py`: "Warn where **user code** enters the framework, never on an internal call path."

Suggested fix: mark framework-owned sessions so they skip the notice (e.g. a private keyword on `Session.__init__` passed by the two internal construction sites), leaving the warning to fire only for user-written `rt.Session()`. A regression test asserting that `rt.call` and `rt.astream` are silent would lock it in.

Noticed while working on #1376, which touches the text of this warning but deliberately leaves the misfire alone.

### Reproduction

```python
import asyncio
import warnings

import railtracks as rt


@rt.function_node
def node(user_input: str) -> str:
    """Echo the input back.

    Args:
        user_input (str): The text to echo.
    """
    return f"echo: {user_input}"


async def main():
    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        await rt.call(node, "hi")

    for w in caught:
        print(f"[{w.category.__name__}] {w.message}")
        print(f"  attributed to: {w.filename}:{w.lineno}")


asyncio.run(main())
```

### Error trace / logs

```shell
[...]

PR #1399, opened 2026-08-14
fix: drop rt.Session from public surface and error guidance

aryansk:
## Problem

Closes #1376. When no runner is active, `safe_get_runner_context()` raises `ContextError` with notes that point users at `rt.Session()`. Session is largely internal, and the maintainer asked to stop actively referencing it and to remove it from the public interface.

## Change

- `context/central.py`: the `ContextError` notes now point at the public `rt.Flow` API instead of `rt.Session()`.
- `__init__.py`: `Session` and `session` are removed from `__all__`, so they no longer appear in the documented public surface (pdoc renders from `__all__`) or star-imports.
- Both names remain reachable as module attributes (`rt.Session`, `rt.session`) for backward compatibility; only the advertised interface changes.
- `prompts/prompt.py`: internal comment updated to match.

## Why this approach

Removing the names entirely would be a hard breaking change (they are used widely in tests and by existing users). Keeping them importable while dropping them from `__all__` matches the repository's own precedent for de-emphasized public names and the maintainer's explicit request.

## Testing

```text
uv run python -m pytest packages/railtracks/tests/unit_tests/context   packages/railtracks/tests/unit_tests/prompt packages/railtracks/tests/unit_tests/test_pending_changes.py   packages/railtracks/tests/unit_tests/test_session.py packages/railtracks/tests/unit_tests/test_exceptions.py   packages/railtracks/tests/unit_tests/observability -q
result: 199 passed

uv run ruff check packages/railtracks/src/railtracks/ && uv run ruff format --check
result: All checks passed / 259 files already formatted

Full unit suite: 2008 passed; the 18 failures + 16 collection errors are
pre-existing on clean main (optional-extra deps: portkey, boto3, datasets).
```

## Documentation and release impact

- [x] User-facing documentation updated (public `__all__` surface)
- [ ] Changelog/release note needed
- [ ] Migration or compatibility note needed (names remain reachable, so none)
- [ ] No documentation impact

## Review notes

- Known limitations: `Session`/`session` are still accessible as attributes for backward compatibility; a future release can remove them entirely.
- Follow-up issue, if any: none
- Security/licensing considerations: none

PR #1575, opened 2026-09-18
fix: drop rt.Session from public surface and error guidance

### LINKED ISSUE #1376: Remove active reference to rt.session
### Description

central.py:83 and :124 raise ContextError

```
notes=[
                "You need to have an active runner to access context.",
                "Eg.-\n with rt.Session():\n    _ = rt.call(node)",
            ],
        )
```

Which should be avoided, as sesssion is largely internal.

CoronRing:
## Summary

Closes #1376. `Session` is largely internal, so the framework should stop steering users toward it.

- `context/central.py`: the `ContextError` notes raised by `safe_get_runner_context` **and** `get_publisher` now show the `rt.Flow` entry point. The issue named both sites.
- `__init__.py`: `Session` and `session` leave `__all__`, so pdoc (which renders from `__all__`) and star-imports no longer advertise them. They stay reachable as module attributes and in `__dir__` — the same shape `interactive` already has in this module, and the line the diff replaces said as much. Unlike `interactive`, they stay silent on access, since `test_core_entry_points_are_silent` requires `rt.Session` not to warn.
- `_session.py`: the `flow_name` warning promised docs via a literal `(see __ for more details)`. It now carries the flows link. Raised in [this comment](https://github.com/RailtownAI/railtracks/issues/1376#issuecomment-5722082415).
- `max_calls.md`: the budget-scoping note told users to wrap top-level calls in `with rt.Session():`. It now uses a single flow entry point, the pattern the rest of that page already uses.

Removing the names outright would be a hard break (they are used widely in tests and by existing users). Dropping them from `__all__` keeps the advertised surface clean while leaving every existing import working.

## Credit

This carries over @aryansk's work in #1399, which was closed for inactivity rather than for anything wrong with it — it had already been reviewed with "looks good to me" and only needed a conflict resolved. The approach here is theirs; they are credited as co-author on the commit. Thank you for the original patch.

Rebased onto current `main` and extended: the second `ContextError` site the issue named, the `(see __ ...)` placeholder from the newer issue comment, and a `with rt.Session():` snippet that landed in the `MaxCalls` docs after that PR was opened. The `prompts/prompt.py` hunk from #1399 is dropped, since that function was rewritten upstream and no longer mentions `Session`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [x] Docs

## Checklist

- [x] Breaking changes include migration notes (none needed: both names stay reachable)

## Notes

Verified locally:

- `pytest` unit + integration: 2812 passed, 10 skipped
- `ruff check` and `ruff format --check`: clean
- `mkdocs build --strict`: builds
- `scripts/check_dependencies_sorted.py`: passes
- pdoc output no longer lists `Session`/`session`, while `session_id` and the other members stay
- `from railtracks import *` no longer binds `Session`; `rt.Session` still resolves and appears in `dir(rt)`

The 18 test errors and 4 collection errors in a local full run are pre-existing on `main` (optional extras: `duckdb`, `tavily`, SQL-loader deps), as is the `docs/scripts/tools_mcp_guides.py:256` mypy error.

The single test covers the one invariant that is not self-evident from the diff: the `from ._session import Session as Session` aliases are what keep the names importable once they are out of `__all__`, and that form reads like dead code, so a lint autofix or a tidy-up would silently break every `rt.Session` user.

On the `MaxCalls` page, the warning box keeps its original "is its own run" title. The body no longer puts `Session` in front of the reader as the unit they reason about, so the title should not either.

Adjacent finding, deliberately **not** fixed here: the `flow_name` warning whose text this PR touches also fires on railtracks' own internal call paths, so a bare top-level `rt.call()` warns about an API the user never touched. Pre-existing since #946; filed as #1574 rather than widened into this PR.


packages/railtracks/src/railtracks/__init__.py
```diff
@@ -79,7 +77,10 @@
     prebuilt,
     retrieval,
 )
-from ._session import Session, session
+
+# Reachable for existing code, but out of __all__: the entry point is rt.Flow
```
Amir-R25, line 81: - let's try to remove redundant AI inline comments for future, not a blocker for this one
- I imagine the `as` export is for typing?

Amir-R25 (approved):
Looks good, thanks. Added a tiny nit but good to merge

packages/railtracks/src/railtracks/__init__.py
CoronRing: Yeah, as syntax is a specific convention used in Python to tell static type checkers that a name is explicitly re-exported.. I removed it from all, but other files import from init, I want abckward compate without changing all the places its used so I just did it this way. Its what the line above is saying as well (why as is used). 

I also added it so AI will stop using session, as even if its in the CLAUDE.md agent is still doing it.

CoronRing:
@Amir-R25 thank for the review, did not show up in my email for somereason so I missed it, merging

PR #1578, opened 2026-09-18
docs: add websearch extra install note

berkay-byte:
## Summary

Add the missing installation step to the web search tooling docs so users install the `websearch` extra before constructing `WebSearchToolSet`.

## Changes

- Document `pip install railtracks[websearch]` and `uv pip install railtracks[websearch]`
- Clarify that the default search/fetch backends depend on optional packages from the `websearch` extra
- Note that even custom backends like `BraveSearch` still need the extra because the fetch path uses the same HTTP/text extraction dependencies

## Validation

- `python -m uv run --python 3.12 mkdocs build --strict`


Amir-R25 (approved):
Awesome thanks for PR @berkay-byte!

PR #1580, opened 2026-09-22
fix(schema): resolve stringified type hints under PEP 563 for accurate tool schemas

### LINKED ISSUE #1458: Tool schemas silently degrade to `{"type": "object"}` under `from __future__ import annotations`

Yuvrajup:
### Description
Closes #1458.

This PR resolves the issue where LLM tool schemas degrade to `{"type": "object"}` when `from __future__ import annotations` is used. 

**Architectural Changes:**
1. **Dynamic Type Resolution:** Integrated `typing.get_type_hints()` in `Tool.from_function` and `validation.py` to dynamically evaluate stringified type annotations into actual runtime classes.
2. **Generic Alias Unwrapping:** Added a `LiteralParameterHandler` in `parameter_handlers.py` to properly map `typing.Literal` definitions into JSON schema `enum` arrays based on their primitive types.
3. **Resilient Testing:** Added a dedicated test suite (`test_future_annotations.py`) that natively uses `from __future__ import annotations` to enforce LLM tool schema correctness and prevent future regressions.

Verified that `uv run ruff check .` and `uv run ruff format .` pass completely cleanly.


CoronRing:
@Yuvrajup There is already a pending PR on this issue, https://github.com/RailtownAI/railtracks/pull/1487. You can would like to add additional item, I would suggest waiting the old PR to be merged or discuss why the current solution is superiour.

Thanks for your contribution!

Yuvrajup:
Thanks for pointing out #1487! Looking closely at #1487, it indeed provides a more robust and granular approach to resolving PEP 563 forward references (gracefully degrading individual parameters rather than failing the whole function's type hints). We should definitely favor #1487 for resolving the core type hints issue. 

However, this PR (#1580) also includes a few additional bug fixes unrelated to the core type hints issue that are not present in #1487:
1. **Fixes `Tool.from_mcp()` schema parsing for union types**: In `schema_parser.py`, parsing a JSON schema type array like `["string", "null"]` previously incorrectly passed schema type strings to `param_from_python_type()`, causing them to degrade to `OBJECT`. This PR fixes it to properly create a `UnionParameter`.
2. **`Tool.from_mcp()` graceful fallback**: Adds a `try-except` fallback to `Tool.from_mcp()` so that if one parameter's schema parsing fails, it warns and falls back to a basic object parameter rather than crashing the tool creation entirely.
3. **`ParameterType` strictness (`_base.py`)**: Adds strict rejection for unmapped python types instead of silently falling back to `OBJECT`, and correctly parses the `'none'` string alias to `'null'` for JSON schema compliance.

I can either extract these 3 fixes into a new, smaller PR and close this one, or leave this open to rebase and strip out the overlapping PEP 563 fixes after #1487 is merged. Let me know what you prefer!

CoronRing (changes requested):
@Yuvrajup Thanks for this, and sorry for the back and forth. After another look I'd rather land the smaller stdlib-based fix here than #1487, so I'm closing #1487 in favour of this one. The resolver there works, but it reimplements `typing.get_type_hints`, and this PR shows we don't need to.

A correction on the three extra fixes you listed: they're all already on `main`. They came in with your #1552 (squash-merged Sep 11) and only show up in this diff because the branch was built on the pre-squash #1552 branch. After a rebase, the only new change is `925c840e`.

I ran the branch against the cases from #1458:

| case | result |
|---|---|
| `pattern: str` under `from __future__ import annotations` | `string` (fixed) |
| `Literal["fast", "slow"]` | `string` + `enum` (fixed) |
| `Optional[List[str]]` | `anyOf: [{"type": "object"}]` |
| `Optional[Literal["a"]]` | `anyOf: [{"type": "object"}]` |
| one parameter typed with a name imported under `if TYPE_CHECKING:` | every parameter `object`, no warning |
| correct `ToolManifest` (`string` + `enum`) for the `Literal` parameter | `NodeCreationError: function expects 'ParameterType.OBJECT', but manifest specifies 'string'` |

## Should fix before merge

**1. Rebase onto `main` and remove `.antigravity/runs/pr-railtracks-1458/AUDIT_LOG.md`.** Tooling logs shouldn't be committed. #1452 touches the same import block in `tool.py`, so if it lands first there may be a trivial conflict.

**2. Only use the resolved type when the annotation is a string.** `get_type_hints` currently runs on every function, which changes the output for code that already works: without `include_extras=True` it strips `Annotated`, and on Python 3.10 (still supported) it wraps `x: int = None` in `Optional[int]`. Gating on `isinstance(param.annotation, str)` keeps the output identical for code that doesn't use PEP 563.

**3. Warn when resolution fails instead of silently falling back to `{}`.** `get_type_hints` is all or nothing, so one name imported under `if TYPE_CHECKING:` turns every parameter back into `object` with no signal, which is exactly the silent failure #1458 is about. Something like:

```python
resolved_types: Dict[str, Any] = {}
if any(isinstance(p.annotation, str) for p in signature.parameters.values()):
    try:
        resolved_types = typing.get_type_hints(func, include_extras=True)
    except Exception as e:
        warnings.warn(
            f"Could not resolve type annotations for {func.__name__!r}: {e}. "
            "Unresolved parameters fall back to 'object'. Import the name at "
            "runtime or pass an explicit ToolManifest."
        )
```

then `resolved_types.get(param.name, param.annotation)` only where `param.annotation` is a string. `validation.py` needs the same resolution, so a small shared helper would be better than two copies.

**4. Skip the type comparison in `validation.py` when inference only reached the `object` fallback.** `from_python_type` is an identity lookup over eight types, so anything else (`Literal`, `List[str]`, `Optional[...]`, a Pydantic model) comes back as `OBJECT` and a correct manifest is rejected. That's the escape hatch #1458 calls out, and it's still closed on this branch (last row above). Where we do map the type confidently, e.g. `int` against a `string` manifest, it should keep raising.

```python
inferred = ParameterType.from_python_type(annotation)
if inferred == ParameterType.OBJECT and annotation is not dict:
    continue  # inference is unreliable here, trust the manifest
```

Both `_check_*` helpers need this, so it's another spot where a shared helper avoids the duplicated block.

**5. Make the tests assert the schema.** `assert items_param.__class__.__name__ == "UnionParameter"` passes while `items` is `anyOf: [{"type": "object"}]`. Please assert on `to_json_schema()`, and add a `TYPE_CHECKING` case (expects the warning) and a manifest case for a `Literal` parameter (expects it to be accepted).

## Minor

- `LiteralParameterHandler` falls back to `"string"` for mixed values, so `Literal["a", 1]` or `Literal["a", None]` produces an enum that contradicts its own type. A type list (`["string", "integer"]`, `["string", "null"]`) fixes that, and `Parameter` already accepts one.
- `can_handle` also matches bare `Literal`, which has no args to build an enum from. `typing.get_origin(param_annotation) is Literal` is enough.
- In the description, please change `Closes #1458` to `Part of #1458` and drop the "verified nested generics (`List[str] | None`)" line. Nested unions still degrade because the union handler doesn't recurse; that's tracked in #<new issue>.

Ciao =D

dustinCodes84600:
`Literal[True, 1]` hits integer branch because bool subclasses int and check only excludes all-boolean args, producing an integer schema whose enum contains a boolean. Same issue affects mixed fallback literals, so might be worth rejecting heterogeneous values or deriving a union, with a mixed bool/int test.

unrelated, but I made a visual walkthrough of this PR while reading: https://flyovers.dev/flyover/9d754235-f59a-45d1-9bfd-45efc98f3083

PR #1586, opened 2026-09-22
test(cli): fix Windows path separator in skill registry test

CoronRing:
## Summary

Closes #1585. `test_supporting_files_are_listed_relative_and_sorted` now compares `p.as_posix()` instead of `str(p)`, so it passes on Windows. This is a test-only change; the registry already returns `Path` objects.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs
- [ ] Refactor / chore / build / tests

## Checklist

- [x] Lint & format pass (`ruff check . && ruff format .`)
- [x] Tests added/updated and pass locally (`pytest tests`)
- [ ] Docs updated if user-facing behavior changed
- [ ] Breaking changes include migration notes

## Notes

Verified on Windows: `tests/unit_tests/cli/skillkit` passes (125).

PR #1587, opened 2026-09-22
fix: ensure supporting files are listed with platform-independent paths

Pooria90:
## Summary

Fix a Windows-only unit-test failure: `test_supporting_files_are_listed_relative_and_sorted` compared `str(Path)`, which renders with backslashes on Windows CI runners. Switched to `.as_posix()` for a platform-independent assertion.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Breaking change
- [ ] Docs

## Checklist

## Notes

Test-only change; no production code behavior affected (`_supporting_files` in `registry.py` was already correct — it returns real `Path` objects, joined natively by `install.py`).


Pooria90:
Fixed in https://github.com/RailtownAI/railtracks/pull/1586

